PR #6, opened 2020-04-26
Dashicons - Adding the WP5.5 icons

ePascalC:
https://meta.trac.wordpress.org/ticket/5172
Fixed bbPress icon code
Added some missing icons from releases up to WP 5.4
Added the upcoming WP 5.5 dashicons


dd32:
Fixed via https://meta.trac.wordpress.org/changeset/10149

PR #7, opened 2020-04-27
Centralise the SEO fixes from all the plugins/themes into a single plugin

dd32:
Ticket: https://meta.trac.wordpress.org/ticket/5173


wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-seo/canonical.php
```diff
@@ -0,0 +1,98 @@
...
+		$url = home_url( '/' );
+	} elseif ( is_date() ) {
+		if ( is_day() ) {
+			$url = get_day_link( get_query_var('year'), get_query_var('monthnum'), get_query_var('day') );
+		} elseif ( is_month() ) {
+			$url = get_month_link( get_query_var('year'), get_query_var('monthnum') );
+		} elseif ( is_year() ) {
+			$url = get_year_link( get_query_var('year') );
+		}
+	}
```
braders, line 56: If I understand correctly, this will output a canonical meta tag for search results and archives. However, these content types are also no-indexed in robots.php.

Ideally the canonical url meta would no be output on no-index pages, and Google has previously indicated the presence of a canonical URL may take precedence over the no-index: https://www.searchenginejournal.com/google-dont-mix-noindex-relcanonical/262607/

dd32: If you feel strongly about that, you should report that upstream to the Meta Trac, I don't think that's something I'm going to fix here, considering I'm just trying to combine all the existing canonical and noindexing code we have.

Reading that article though, it suggests to me that if rel=canonical was directed to a noindex page, but the others weren't, then it may ignore the noindex.

If I had my way, no content on WordPress.org would be noindex.

PR #10, opened 2020-05-22
Add language requirement to user notes form

crstauf:
https://meta.trac.wordpress.org/ticket/5223

Adds a language requirement to the developer.wordpress.org user notes forum.

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-developer/inc/user-content.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-developer/inc/user-content.php
@@ -458,6 +458,8 @@ class DevHub_User_Submitted_Content {
 				. $feedback_rule
 				. '</li><li>'
 				. __( 'This form is not for support requests, spam, bug reports, complaints, or self-promotion. Entries of this nature will be deleted.', 'wporg' )
+				. '</li><li>'
+				. __( 'Notes and feedback must be written in English. Entries not written in English will be deleted.', 'wporg' )
 				. '</li><li class="user-notes-are-gpl">'
 				. $license_rule
 				. '</li></ul>';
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/themes/pub/wporg-developer/inc/user-content.php
```diff
@@ -458,6 +458,8 @@ public static function get_editor_rules( $context = '', $note_link = '' ) {
 				. $feedback_rule
 				. '</li><li>'
 				. __( 'This form is not for support requests, spam, bug reports, complaints, or self-promotion. Entries of this nature will be deleted.', 'wporg' )
+				. '</li><li>'
+				. __( 'Notes and feedback must be written in English. Entries not written in English will be deleted.', 'wporg' )
```
Kenshino, line 462: Hey @crstauf I think it'd be better if we updated the language at the end to read

`Entries not written in English will not be accepted`

wordpress.org/public_html/wp-content/themes/pub/wporg-developer/inc/user-content.php
```diff
@@ -466,6 +468,8 @@ public static function get_editor_rules( $context = '', $note_link = '' ) {
 		return '<ul><li>'
 			. __( 'This form is not for support requests, discussions, spam, bug reports, complaints, or self-promotion. Entries of this nature will be deleted.', 'wporg' )
 			. '</li><li>'
+			. __( 'Notes and feedback must be written in English. Entries not written in English will be deleted.', 'wporg' )
```
Kenshino, line 471: Same as above.

Take note that I'm aware the line that discourages spam / support requests says delete. I think those cases warrant a stronger line.

wordpress.org/public_html/wp-content/themes/pub/wporg-developer/inc/user-content.php
crstauf: @Kenshino thanks!

is there an alternative that doesn’t use a double negative, to make comprehension easier?

i’m not sure if it’s an actual grammar rule, but i try to avoid using them.

dd32: Perhaps another way to write this would be to remove the "will be deleted" from the existing rules and this new added one, and a new entry added at the end along the lines of "Any notes not meeting these requirements will be removed by the moderation team."

Kenshino: I like @dd32 's suggestion :)

crstauf: Adjusted in 694cbdb.

crstauf:
@dd32 Just checking to see if there's an additional step needed from me to merge this?

coffee2code:
Fixed in https://meta.trac.wordpress.org/changeset/9964.

Thanks for the PR!

crstauf:
Welcome! and thanks, @coffee2code!

PR #14, opened 2020-08-03
Fix PHP 7.1 compatibilty issue

xknown:
"$this" can no longer be used in a static function or method since PHP 7.1.


xknown:
@coffee2code can you please take a look at this diff when you have some free time?

I'm currently working on migrating the wp.org code to be compatible with PHP 7.4.

PR #15, opened 2020-08-09
Fix PHP 7 notices/warnings on codex.buddypress.org

xknown:
This removes the PHP4 style constructor as well as a couple of create_function calls.

This is part of an effort to migrate the existing code to PHP 7.4.


xknown:
@Otto42 @dd32 can you please take a look when you have some free time?

PR #17, opened 2020-08-10
Add Release Confirmation for plugins

dd32:
This is a Work in progress PR for adding release confirmations for plugin releases.

What this means, is that an email will be sent upon a new version of a plugin being released, and the import of that plugin will be deferred until appropriate responses are received from the required number of people.

This is a draft PR, for visibility, logic is still being discussed on the trac ticket.

Trac Ticket: https://meta.trac.wordpress.org/ticket/5352


dd32:
Merged

PR #19, opened 2020-09-10
Plugin Directory: Remove unused React code

ryelle:
Remove the react theme code, leaving only the screenshot component (the only part of the app currently in use). This leaves most of the CSS in place, but collapses it down into fewer folders. This also updates the build process to use `wp-scripts` for the JS, so we can get the dependency extraction code.

`npm run build` or `npm run watch` can be used to build the CSS and JS.

https://meta.trac.wordpress.org/ticket/5431


ryelle:
Committed in [r10605](https://meta.trac.wordpress.org/changeset/10605).

PR #24, opened 2021-02-08
Switch block editor import to use new trunk name for default branch

mkaz:
https://meta.trac.wordpress.org/ticket/5606


ryelle:
This was committed in [r10734-meta](https://meta.trac.wordpress.org/changeset/10734). Thanks!

PR #25, opened 2021-03-02
[WIP] Initial attempt at restful api for posting automating theme results.

StevenDufresne:
**Please do not merge**

## Description
This PR adds a Rest controller that will be used to post the results of automated theme testing to a theme Trac ticket. 

**Details**
- Url: `//wordpress.org/themes/wp-json/themes/v1/github/{theme_slug}/{trac_ticket}`.
- Auth:`Bearer {token}` must be included.
- Data format: `content={message}`

## Questions:
I am currently using the theme ticket's `keywords` to determine whether the ticket number and the theme line up. 
1. Do we even need to make that check?
2. If yes, is there a better way? (I don't see it anywhere else).
 

## Thoughts:
This probably needs more testing for layout purposes.

## Screenshot
The user that has posted the snippet below is not the same as the code. It will post the comment using the theme tracbot account.


dd32:
> I am currently using the theme ticket's keywords to determine whether the ticket number and the theme line up.
>
> Do we even need to make that check?

I don't think so, I think it can be trusted that the ticket is correct, as otherwise there's a data-error that's probably a bigger problem.

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-auto-review.php
```diff
@@ -0,0 +1,173 @@
...
+		$padding = PHP_EOL . PHP_EOL;
+
+		// Add some space around the titles
+		$content = str_replace( "]:", "]:" . $padding, $content );
+		$content = str_replace( "[", $padding . "[", $content );
+
+		// Remove empty spaces
+		$content = trim( $content );
+
+		$disclaimer = sprintf( __( "''The following has been generated by https://github.com/WordPress/theme-review-action, this is for informational purposes only.''" . PHP_EOL, 'wporg-themes' ) );
```
dd32, line 110: The Trac formatting `''` should be moved out of the translated string here.

I'm not sure translating it is of much benefit either though, as it'll probably always be in English. It could be marked for translation later if need be?

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-auto-review.php
```diff
@@ -0,0 +1,173 @@
+<?php
+
+namespace WPORG_Theme_Directory_Trac;
```
dd32, line 3: I think we should use a more consistent namespace, `WordPressdotorg\Theme_Directory` would match other properties within WordPress.org.

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-auto-review.php
```diff
@@ -0,0 +1,173 @@
...
+use WP_Error;
+use WP_REST_Controller, WP_REST_Server, WP_REST_Response;
+
+defined( 'WPINC' ) || die();
+
+/**
+ *
+ * @see WP_REST_Controller
+ */
+class WPorg_Themes_Rest_API_Auto_Review_Controller extends WP_REST_Controller {
```
dd32, line 14: Since this class is Namespaced, the class name can be shorter, consider that the classname is currently pretty long and duplicative `\ WPORG_Theme_Directory_Trac\WPorg_Themes_Rest_API_Auto_Review_Controller`

If it were changed as above, `WordPressdotorg\Theme_Directory\Rest_API\Auto_Review_Controller` is more "namespacey"

This is really a nitpick than anything else, and it doesn't affect any logic what so ever :D

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-auto-review.php
```diff
@@ -0,0 +1,173 @@
...
+
+		if( ! $updated_ticket ) {
+			return new \WP_Error(
+				'rest_error_updating_ticket',
+				__( 'We ran into an error updating the ticket.', 'wporg-themes' ),
+				array( 'status' => \WP_Http::INTERNAL_SERVER_ERROR )
+			);
+		}
+
+		return new WP_REST_Response( __( 'Success.', 'wporg-themes' ), \WP_Http::OK );
```
dd32, line 169: Simply `return true;` is probably enough here, there's probably no need to return a human-readable success message?

coreymckrill: I think it's good to include the `200` status code, since API clients will often explicitly look for that.

dd32:
This PR was auto-closed as the branch was deleted. The branch was deleted, as it was created on this repo rather than on a fork. The meta.git.wordpress.org => Github sync process will delete any branches that don't exist upstream.

StevenDufresne:
No problem. I'll use a fork instead.

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-auto-review.php
StevenDufresne: Yeah fair enough, I'll remove the translation for now since this is experimental.

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-auto-review.php
StevenDufresne: No, this makes sense, I copied another file's format. But it's much cleaner to use the namespaces logically.

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-auto-review.php
StevenDufresne: I was under the impression that passing `\WP_Http::OK` as the second param will set the status code to `200`?

StevenDufresne: Ah, in relation to @dd32's comment... I do want to return a proper `200` status, as @coreymckrill mentions. In case the client throws an exception for not receiving a 200.

coreymckrill: Oh yep, `\WP_Http::OK` is fine. I just meant I think it should include a status code.

PR #26, opened 2021-03-03
Rest Endpoint to post auto theme review results to theme trac ticket.

StevenDufresne:
This PR has been moved from https://github.com/WordPress/wordpress.org/pull/25.

## Description
This PR adds a Rest controller that will be used to post the results of automated theme testing to a theme Trac ticket. 

**Details**
- Url: `//wordpress.org/themes/wp-json/themes/v1/github/{theme_slug}/{trac_ticket}`.
- Auth:`Bearer {token}` must be included.
- Data format: `content={message}`

## Ideal Rollout
We are already running tests on theme uploads: https://github.com/WordPress/theme-review-e2e/actions

We can start by only postings **errors** from the UI check. Since the majority of the tests only issue warnings, it will only post a subset of errors. 

**More specifically, it will catch**:
- Errors in PHP
- Errors in JS
- Unapproved links
- Errors WCAG 2.0 Accessibility errors when the theme has accessible-ready.

## Screenshot
The user that has posted the snippet below is not the same as the code. It will post the comment using the theme tracbot account.


carolinan:
The difficult part is going to be how to make the messages more _understandable._

1) Where do theme authors go if they are not sure what the error means or how to fix it? 
Are there going to be more resources at the Theme Review Action link, or is it only a link to the repo?

2) How will the theme author know if they are required to fix it or if it is only recommended?

Right now, the theme check results are displayed in the trac ticket, and some authors has expressed that it is unclear:
Who is posting those results, and why, 
If the author are meant to fix the problems or wait for a reviewer

Is the result also sent by email to the theme author?

StevenDufresne:
@carolinan What do you think about something like:
https://github.com/WordPress/theme-review-action/pull/19

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-auto-review.php
```diff
@@ -0,0 +1,183 @@
...
+		$content = preg_replace(
+			'%See: \b(([\w-]+://?|www[.])[^\s()<>]+(?:\([\w\d]+\)|([^[:punct:]\s]|/)))%s',
+			'See: <a href="$1">$1</a>',
+			$content
+		);
+
+		// Remove empty spaces
+		$content = trim( $content );
+
+		$disclaimer = "<p><i>The following has been generated by <a href='https://github.com/WordPress/theme-review-action'>Theme Review Action</a>, this is for informational purposes only.</i></p>";
```
tellyworth, line 115: This could be a more explicit call-to-action (or call-to-inaction?). Perhaps add something like:

```This is intended to help find issues in the theme that fail to comply with the guidelines. It is not mandatory to pass all of these tests.
Theme developers: the documentation links may help you to discover and fix problems that might otherwise delay approval of your theme.
Theme reviewers: the test results may identify issues that need particular attention.
Errors and omissions in the test results can be reported to [GH issues link]
```

StevenDufresne:
I made some updates; here's an example of the output.

StevenDufresne:
Merged.

PR #31, opened 2021-05-20
Create a child theme for embedding Openverse

dhruvkb:
Fixes: WordPress/openverse-frontend#64
Closes: https://meta.trac.wordpress.org/ticket/5789

This PR adds a new child theme, to render the Openverse frontend under WordPress.org.


dhruvkb:
@zackkrida should I mark this PR 'Ready for review'? The only reason it is in WIP is because the URL for Openverse is not live yet.

zackkrida:
I think we'll wait until it's ready @dhruvkb

zackkrida:
@dhruvkb I want to add one more thing before we initializing merging this. We should add some _custom error messaging_ for when the iframe URL returns a 5xx error. Something like:

"Openverse is currently unavailable. Please get in touch with us in the [#Openverse] channel (https://wordpress.slack.com/messages/openverse/) on Slack or via a comment on https://make.wordpress.org/openverse/."

We probably need to do some polling or a check after the iframe is loaded to toggle a boolean, but I think this is worth doing in case of an outage.

zackkrida:
@dhruvkb Something critical I forgot—passing url paths from the parent to the iframe! So all URLs work—which will be especially important when we redirect all search.cc.org urls to wp.org/openverse. Something simple like this pseudo-js should be sufficient:

```js
document.addEventListener('DOMContentLoaded', () => {
 iframe.src = openverseUrl + window.location.pathname
})
```

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
```diff
@@ -0,0 +1,71 @@
...
+2.  You can choose to set up a new environment automatically or work in an
+    existing environment with manual setup.
+
+    **Automatic:**  
+    You can set up the development environment automatically using
+    `@wordpress/env`. _This requires Docker._
+
+    1.  Install `@wordpress/env` globally as an `npm` package.
+        ```bash
+        $ npm install -g @wordpres/env@latest
```
obulat, line 35: typo: wordpres -> wordpress

dhruvkb: Nice catch! In retrospect, should have copied from the terminal instead of typing 😆.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
```diff
@@ -0,0 +1,71 @@
...
+Follow these steps to set up a local playground for the theme:
+
+0.  Install all the prerequisites.
+
+    1.  **Required:** Node.js.
+    2.  **Recommended:** Docker (to use the automatic setup)
+
+1.  Build the parent theme WordPress.org theme.
+
+    1.  `cd` into the parent theme directory at `../wporg`.
```
obulat, line 16: I think it would be better to have the full path to the directory from content root.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
dhruvkb: Resolved in 772a211.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
dhruvkb: It's actually quite deep down in the folder tree so the full path is considerably long. Also since the `README.md` is itself not in the root of the repo, the person reading it will quite likely already be in the `wporg-openverse` folder. From there, a relative path will be more helpful imo.

dhruvkb: Why not both? Resolved in 2dc4ac2.

obulat (approved):
Works great, and so easy to customize and test! Happy to see Openverse on WordPress.org :)

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -0,0 +1,118 @@
+<?php
+
+namespace WordPressdotorg\Openverse\Theme;
+
+/**
+ * This is the URL on which the frontend site of Openverse is hosted. Unless
+ * overridden from the Customizer UI, this is the URL for the embedded `iframe`.
+ */
```
zackkrida, line 8:
```suggestion
/**
 * This is the URL on which the frontend site of Openverse is hosted. Unless
 * overridden from the Customizer UI, this is the URL for the embedded `iframe`.
 *
 * Note: Do not put a trailing slash '/' in this url, as it will cause problems.
 */
```

zackkrida (approved):
Works perfectly. I added a small comment based on an issue I encountered while testing with my local Openverse install, but this is excellent work.

zackkrida:
Improvements are welcomed, nice!

zackkrida:
@dhruvkb please edit the description when you can to include:

```
Closes: https://meta.trac.wordpress.org/ticket/5789#ticket
```

PR #33, opened 2021-07-02
Add a better search class

tellyworth:
This replaces the old, hacked-up jetpack-search.php class with a new Plugin_Search class that uses filters to interact with the Jetpack_Search class in the currently installed version of the Jetpack plugin.

The Plugin_Search class has a constant that can be used to switch between the old and new search. Theoretically they should produce the same search results.

There are unit tests that can be run like this:

```
composer install
wp-env start
composer test
```

The tests will respect the `USE_OLD_SEARCH` constant, so they can be used to test against both the old and new search classes. This is how I confirmed that the new search works more-or-less the same.

I've also tested the basics manually on my sandbox, but not extensively. It seems to work ok. If there's a serious problem it's most likely that I've missed something in copying stuff into this PR.

Some notes for the Jetpack search folks:

It seems that the `jetpack_search_es_wp_query_args` hook is the preferred way for a plugin to interact with and customize Jetpack_Search's behaviour. I've done that where possible, but since `convert_wp_es_to_es_args()` is very limited and particular about what parameters it accepts I had to resort to filtering and altering the query at a lower level, via `jetpack_search_es_query_args`. See the `jetpack_search_es_wp_query_args()` and `jetpack_search_es_query_args()` for examples.

The plugin directory is a very specific use case that's probably right on the edge. But there are likely some ideas to be found in there that might be worth moving upstream, to make it possible to use the higher-lever filter instead.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-search.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-search.php
@@ -1,0 +1,404 @@
+<?php
+namespace WordPressdotorg\Plugin_Directory;
+
+// Hmm
+add_filter( 'option_has_jetpack_search_product', '__return_true' );
+
+
+/**
+ ** Override Jetpack Search class with special features for the Plugin Directory
+ **
+ ** @package WordPressdotorg\Plugin_Directory
+ **/
+class Plugin_Search {
+
+	// Set this to true to disable the new class and use the old jetpack-search.php code.
+	const USE_OLD_SEARCH = true;
+
+	// Internal state
+	protected $locale;
+	protected $is_block_search;
+	protected $is_english;
+	protected $en_boost;
+	protected $desc_boost;
+	protected $desc_en_boost;
+
+	/**
+	 * Fetch the instance of the Plugin_Search class.
+	 *
+	 * @static
+	 */
+	public static function instance() {
+		static $instance = null;
+
+		if ( ! $instance ) {
+			$instance = new self();
+		}
+
+		return $instance;
+	}
+
+	/**
+	 * Plugin_Search constructor.
+	 *
+	 * @access private
+	 */
+	private function __construct() {
+		if ( isset( $_GET['s'] ) )
+			return false;
+
+		add_action( 'init', array( $this, 'init' ) );
+
+		return false;
+	}
+
+	public function init() {
+
+		if ( self::USE_OLD_SEARCH ) {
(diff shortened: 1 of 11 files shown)


wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-search.php
```diff
@@ -0,0 +1,404 @@
...
+		$args['query_fields'] = $matching_fields;
+
+
+
+
+		return $args;
+	}
+
+	public function jetpack_search_es_query_args( $es_query_args, $query ) {
+		// These are the things that jetpack_search_es_wp_query_args doesn't let us change, so we need to filter the es_query_args late in the code path to add more custom stuff.
```
gibrown, line 173: What about just bypassing everything that jetpack_search_es_wp_query_args does and building the whole query in here (completely ignore what jetpack_search_es_wp_query_args has given us).

I think the query could be much easier to review then and the current code feels very fragile in case anything above it changes.

I think you should use the query builder class to create the actual query. 

So effectively you fork the code from here https://github.com/Automattic/jetpack/blob/master/projects/plugins/jetpack/modules/search/class.jetpack-search.php#L912 and then use that to replace the ES query via this filter. The parser/builder takes care of the syntax for you and makes a lot of it more robust to changes.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-search.php
```diff
@@ -0,0 +1,404 @@
...
+					0 => 'author',
+					1 => 'contributors',
+				  ],
+				  'type' => 'best_fields',
+				  'boost' => 2,
+				],
+			];
+		}
+
+		if ( isset( $es_query_args[ 'query' ][ 'function_score' ][ 'functions' ] ) ) {
```
gibrown, line 301: The query builder supports all of this https://github.com/Automattic/jetpack/blob/2138f8787efc1e3552410666dfd14bae14b42c1a/projects/plugins/jetpack/_inc/lib/jetpack-wpes-query-builder/jetpack-wpes-query-builder.php#L87

gibrown (changes requested):
Using the query parser/builder should make this a lot easier to read and maintain.

I think the hooks and overrides into Jetpack should be reviewed also as I think there are better ways.

dd32:
@tellyworth I think this PR was merged in some form right? I don't need to worry about the accidental closure right?

PR #35, opened 2021-07-15
Enhance social images for plugins on Twitter and Facebook

pedro-mendonca:
The Twitter cart type should depend on the available images and image ratio:

- If banner (**2x** or **1x**) exist: Set Twitter card type to `"summary_large_image"`
- If banner don't exist: Fallback to Twitter card type `"summary"` to print a small square image (plugin icon or WordPress logo) 

For the image itself, probably it will be ok to include the **1x banner** for Twitter if there is no **2x banner**, and also probably it will be better to include the **2x banner** for Facebook in the first place.
So I suggest the same logic for both Twitter and Facebook:

- If banner exist: Use it for `twitter:image` and `og:image` (**2x** with fallback to **1x**)
- If only icon exist: Use it for `twitter:image` and `og:image` (**2x** with fallback to **1x**)
- If none exist: Fallback to WordPress square logo for `twitter:image` and `og:image `

https://meta.trac.wordpress.org/ticket/5826

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/functions.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/functions.php
@@ -360,10 +360,10 @@ function social_meta_data() {
 		// Set to banner-772x250.
 		$image = $banner['banner'];
 	} elseif ( ! $icon['generated'] && ( $icon['icon_2x'] ) ) {
-		// Set to icon256x256.
+		// Set to icon-256x256.
 		$image = $icon['icon_2x'];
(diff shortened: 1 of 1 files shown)


dd32:
@pedro-mendonca Apologies on ^. We migrated to a `trunk` main branch a while ago, but due to this repo's structure it wasn't a rename. Due to some unrelated things it's caused the `master` branch to be deleted and closed/invalidated your PR in the process. Please feel free to either rebase upon trunk, or leave it as-is as the diff/patch is still accessible.

pedro-mendonca:
Thanks @dd32 for the heads up.
Rebased :-)

PR #36, opened 2021-07-22
Theme Directory: Display theme upload errors as list.

StevenDufresne:
This PR makes it easier for users to identify the basic errors with their theme upload by showing them the detected errors as a list. This does not address messaging for results from the `theme-check` plugin. 

## Background
Users upload their theme on `wordpress.org/themes/upload`. Before triggering the theme check plugin, and creating a trac ticket, the back end code uses the `style.css` file in the theme and validates a number of things in order to process the upload properly. The problem is that each issue identified stops the upload. Users are forced to work through each issue one at a time because we currently only return one issue at a time.

You can read more about what happens on upload [here](https://github.com/WPTT/Theme-Requirements/issues/13).

## Solution
Let's collect all the errors related to an incorrect `style.css` file in the theme and display them all at one. This will save users revolutions in the `fix -> zip -> upload` loop.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
+++ b/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
@@ -393,7 +393,7 @@ class WPORG_Themes_Upload {
 		}
 
 		// If we had any issues with information in the style.css, exit early.
-		if( !empty( $style_errors ) ) {
+		if ( ! empty( $style_errors ) ) {
 			return $style_errors;
 		}
 
--- a/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
+++ b/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
@@ -187,7 +187,7 @@ class WPORG_Themes_Upload {
 			);
 		}
 
-		$style_errors = [];
+		$style_errors = array();
 
 		// Do we have a readme.txt? Fetch extra data from there too.
 		$this->readme = $this->get_readme_data( $theme_files );
@@ -198,7 +198,7 @@ class WPORG_Themes_Upload {
 		// We need a screen shot. People love screen shots.
 		if ( ! $this->has_screenshot( $theme_files ) ) {
 			/* translators: 1: screenshot.png, 2: screenshot.jpg */
-			array_push( $style_errors, sprintf( __( 'The zip file must include a file named %1$s or %2$s.', 'wporg-themes' ),
+			$style_errors[] = sprintf( __( 'The zip file must include a file named %1$s or %2$s.', 'wporg-themes' ),
 				'<code>screenshot.png</code>',
 				'<code>screenshot.jpg</code>'
 			);
@@ -225,16 +225,16 @@ class WPORG_Themes_Upload {
 				__( 'https://codex.wordpress.org/Theme_Development#Theme_Stylesheet', 'wporg-themes' )
 			);
 
-			array_push( $style_errors, $error );
+			$style_errors[] = $error;
 		}
 
 		// Do not allow themes with WordPress and Theme in the theme name.
 		if ( false !== strpos( $this->theme_slug, 'wordpress' ) || preg_match( '/\btheme\b/i', $this->theme_slug ) ) {
 			/* translators: 1: 'WordPress', 2: 'theme' */
-			array_push( $style_errors, sprintf( __( 'You cannot use %1$s or %2$s in your theme name.', 'wporg-themes' ),
+			$style_errors[] = sprintf( __( 'You cannot use %1$s or %2$s in your theme name.', 'wporg-themes' ),
 				'WordPress',
 				'theme'
-			) );
+			);
 		}
 
 		// Populate author.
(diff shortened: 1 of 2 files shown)


wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
```diff
@@ -196,10 +198,10 @@ public function process_upload( $file_upload ) {
 		// We need a screen shot. People love screen shots.
 		if ( ! $this->has_screenshot( $theme_files ) ) {
 			/* translators: 1: screenshot.png, 2: screenshot.jpg */
-			return sprintf( __( 'The zip file must include a file named %1$s or %2$s.', 'wporg-themes' ),
+			array_push( $style_errors, sprintf( __( 'The zip file must include a file named %1$s or %2$s.', 'wporg-themes' ),
```
dd32, line 201: I would personally avoid using `array_push()` and simply do `$style_errors[] = ...`
```suggestion
			$style_errors[] = sprintf( __( 'The zip file must include a file named %1$s or %2$s.', 'wporg-themes' ),
```

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
```diff
@@ -196,10 +198,10 @@ public function process_upload( $file_upload ) {
...
 				'<code>screenshot.png</code>',
 				'<code>screenshot.jpg</code>'
-			);
+			) );
```
dd32, line 204:
```suggestion
			);
```

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
```diff
@@ -364,31 +365,36 @@ public function process_upload( $file_upload ) {
...
+			array_push( $style_errors, sprintf( __( 'You need to upload a version of %1$s higher than %2$s. Increase the theme version number in %3$s, then upload your zip file again.', 'wporg-themes' ),
 				$this->theme->display( 'Name' ),
 				'<code>' . $this->theme_post->max_version . '</code>',
 				'<code>style.css</code>'
-			);
+			) );
+		}
+
+		// If we had any issues with information in the style.css, exit early.
+		if( !empty( $style_errors ) ) {
```
dd32, line 396:
```suggestion
		if ( ! empty( $style_errors ) ) {
```

wordpress.org/public_html/wp-content/plugins/theme-directory/upload.php
```diff
@@ -49,10 +49,20 @@ function wporg_themes_render_upload_shortcode() {
...
 	if ( ! empty( $_POST['_wpnonce'] ) && wp_verify_nonce( $_POST['_wpnonce'], 'wporg-themes-upload' ) && 'upload' === $_POST['action'] ) {
-		$message = wporg_themes_process_upload();
+		$messages = wporg_themes_process_upload();
 
-		if ( ! empty( $message ) ) {
-			$notice = "<div class='notice notice-warning'><p>{$message}</p></div>";
+		if ( ! empty( $messages ) ) {
+			$notice_content = "";
+
+			if( is_array( $messages ) ) {
```
dd32, line 57:
```suggestion
			if ( is_array( $messages ) ) {
```

wordpress.org/public_html/wp-content/plugins/theme-directory/upload.php
```diff
@@ -49,10 +49,20 @@ function wporg_themes_render_upload_shortcode() {
...
+				foreach( $messages as $message){
```
dd32, line 58:
```suggestion
				foreach ( $messages as $message){
```

wordpress.org/public_html/wp-content/plugins/theme-directory/upload.php
```diff
@@ -49,10 +49,20 @@ function wporg_themes_render_upload_shortcode() {
...
+					$notice_content .= "<li>{$message}</li>";
+				}
+			} else {
+				$notice_content = "<li>{$messages}</li>";
+			}
+
+			$notice = "<h2>" . esc_attr__( 'Upload Errors', 'wporg-themes' ) . "</h2><div class='notice notice-error notice-large'><ul>{$notice_content}</ul></div>";
```
dd32, line 65:
```suggestion
			$notice = "<h2>" . esc_html__( 'Upload Errors', 'wporg-themes' ) . "</h2><div class='notice notice-error notice-large'><ul>{$notice_content}</ul></div>";
```

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
```diff
@@ -341,10 +342,10 @@ public function process_upload( $file_upload ) {
 
 			if ( ! $is_allowed_to_upload_for_theme ) {
 				/* translators: 1: theme slug, 2: style.css */
-				return sprintf( __( 'There is already a theme called %1$s by a different author. Please change the name of your theme in %2$s and upload it again.', 'wporg-themes' ),
+				array_push( $style_errors, sprintf( __( 'There is already a theme called %1$s by a different author. Please change the name of your theme in %2$s and upload it again.', 'wporg-themes' ),
```
dd32, line 345: This error is pretty much a show-stopping "Any other error doesn't matter" case.. but I think the UX of getting given a full error list is worthwhile over burrying a critical block like this half way through the list.

dd32 (approved):
I like this change, although I've a few coding style suggestions inline :)

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
StevenDufresne: Wasn't aware. I like it better for certain!

wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
StevenDufresne: Are you suggesting I split it out? Or it's okay as is?

I considered moving this up to the top under the check for a slug but I just deferred to its current place out of not quite knowing why it's where it is :)

carolinan: I suggest splitting it out if it is not too much trouble.
But still show any other error that comes after it.

dd32: Let's see what feedback we get after this change to see what effect it has. It might make it more confusing for some authors when they get both "you can't do this" and "fix this before you upload" in the same error message, but for existing themes, I think that's likely to be less common.

StevenDufresne:
Applied in: https://meta.trac.wordpress.org/changeset/11168

PR #37, opened 2021-07-28
Update class-upload.php

kafleg:
Removed commented codes from class-upload.php of wordpress.org


dd32:
As all the other code related to this functionality still exists, I'd prefer to leave this commented-out code in place for now - even if there's only an incredibly slim chance of it being resurrected.

PR #39, opened 2021-09-07
Emit locale info from WordPress into Openverse iframe

dhruvkb:
Fixes WordPress/openverse-frontend#189

Closes https://meta.trac.wordpress.org/ticket/5892

---

This PR adds functionality for the Openverse frontend loaded in the `iframe` to ask the parent window information about the locale.

`iframe` &mdash;(`localeGet` event)&rarr; parent window
`iframe` &larr;(`localeSet` event)&mdash; parent window

## Description

The `localeSet` message contains the following data:
```json
{
  "type": "localeSet",
  "value": {
    "dir": "",
    "lang": "en-US",
    "locale": "en_US"
  }
}
```

| Key | Interpretation |
|-|-|
| `dir` | Value of parent HTML `dir` attribute |
| `lang` | Value of parent HTML `lang` attribute |
| `locale` | Value of WordPress `get_locale()` function |

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/header.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/header.php
@@ -22,8 +22,8 @@ get_template_part( 'header', 'wporg' );
 ?>
 
 <script>
-    // used in `message.js`
-    const currentLocale = '<?php echo get_locale() ?>';
+	// used in `message.js`
+	const currentLocale = '<?php echo get_locale() ?>';
 </script>
 
 <div id="page" class="site">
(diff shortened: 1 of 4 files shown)


wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/header.php
```diff
@@ -20,5 +20,11 @@
 
 get_template_part( 'header', 'wporg' );
 ?>
+
+<script>
+	// used in `message.js`
```
zackkrida, line 25: nothing like passing php vars to javascript. brings back lots of memories for me.

zackkrida (approved):
I appreciate you updating the test code!

PR #42, opened 2021-09-24
Warn people that the erasure will not erase their support forum details.

dd32:
See https://meta.trac.wordpress.org/ticket/5916


dd32:
Adding an example to test https://meta.trac.wordpress.org/changeset/11254

Screen Shot 2021-09-24 at 11 35 57 am

dd32:
Adding an example to test https://meta.trac.wordpress.org/changeset/11254 & https://meta.trac.wordpress.org/changeset/11255

Screen Shot 2021-09-24 at 11 35 57 am

PR #43, opened 2021-11-19
Update handbook style for readability, spacing

mkaz:
Increases the default font size from 13px up to a more accessible 16px.

Adds some spacing and slight borders around heading to delineate sections and make the text more readable.

Removes the back to top link that clutters documentation, particularly the API docs with numerous headings in a list. This makes it easier to read.

## Testing

I tested the style changes by adding the following CSS to the [Stylus browser plugin](https://chrome.google.com/webstore/detail/stylus/clngdbkpkpeebahjckkjfobafhncgmne) applied to developer.wordpress.org sites and then browsed the different handbooks.

```css
.single-handbook .devhub-wrap main h2 {
    margin-top: 4rem;
    padding-top: 4rem;
    border-top: 1px solid #AAA;
}

.single-handbook .devhub-wrap main h3 {
    margin-top: 4rem;
    padding-top: 4rem;
    border-top: 1px solid #CCC;
}

.single-handbook {
    font-size: 16px;
}

.single-handbook .site-main p,
.single-handbook .site-main ol,
.single-handbook .site-main ul {
    font-size: 16px;
}

.single-handbook .site-main li {
  margin-bottom: 2.0rem;
}

```

## Help

~~I might of messed up the main.scss file with an auto formatter.~~
~~How is that generated? Is there a build process?~~

I used the Grunt command. 👍

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-developer/inc/cli.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-developer/inc/cli.php
@@ -447,7 +447,6 @@ class DevHub_CLI {
 
 	protected static function add_ids_and_jumpto_links( $tag, $content ) {
 		$items = self::get_tags( $tag, $content );
-		$first = true;
 		$matches = array();
 		$replacements = array();
 
@@ -456,11 +455,6 @@ class DevHub_CLI {
 			$matches[] = $item[0];
 			$id = sanitize_title_with_dashes($item[2]);
 
-			if ( ! $first ) {
-				$replacement .= '<p class="toc-jump"><a href="#top">' . __( 'Top &uarr;', 'wporg' ) . '</a></p>';
-			} else {
-				$first = false;
-			}
 			$a11y_text      = sprintf( '<span class="screen-reader-text">%s</span>', $item[2] );
 			$anchor         = sprintf( '<a href="#%1$s" class="anchor"><span aria-hidden="true">#</span>%2$s</a>', $id, $a11y_text );
 			$replacement   .= sprintf( '<%1$s class="toc-heading" id="%2$s" tabindex="-1">%3$s %4$s</%1$s>', $tag, $id, $item[2], $anchor );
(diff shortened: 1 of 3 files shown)


joyously:
Unifying the font size across all the .org sites is desirable, because the default in the browser is to apply the user's size choice per domain. So if the handbook is too small and I zoom it, when I then go to the forums, it's way too big. This is better now than it used to be, but still could be better.

I think the best way is to not specify a font size at all, or use
 `html {font-size: 100%}` 
to always use the user's choice (which could be 30px due to macular degeneration -- I've known two of these people).
This is actually already in main.css at the top where there is a reset, but then there are other rules that override and handbooks are 13px. Instead of changing the override to 16px, I would remove the override altogether.
Screenshot_2021-12-11_11-12-54

I see that main.css has a rule for 
```
.single-handbook .site-main p, 
.single-handbook .site-main ol, 
.single-handbook .site-main ul {
	font-size: 1.05em;
}
```
Using `em` rather than `px` is much better, but even this bump in size isn't needed if the rule for `.single-handbook {font-size: 13px}` is removed. (and I can have 30px if I have already chosen that)


As for the other changes, I think it looks worse than before! 
_(I pasted your CSS above into Stylus, which I already had)_
I think those lines above the headings add clutter. 
The `4rem` margin and padding is too much, breaking up the flow of the document.
The `2rem` on the list items makes the table of contents too spread out and the lists in the text makes it look like paragraphs instead of sentences that go together.
### Before:
Global Variables Common APIs Handbook WordPress Developer Resources-44

### After:
Global Variables Common APIs Handbook WordPress Developer Resources-25

dd32:
@mkaz Apologies on ^. We migrated to a `trunk` main branch a while ago, but due to this repo's structure it wasn't a rename. Due to some unrelated things it's caused the `master` branch to be deleted and closed/invalidated your PR in the process. Please feel free to either rebase upon trunk, or leave it as-is as the diff/patch is still accessible.

PR #48, opened 2022-01-09
Fix old url so redirect not needed

mkaz:
The replacement in the code goes to the older designers-developers/developers link which then redirects to the reference-guide location.

This PR updates to the reference-guide location so an extra redirect is not needed and the link refers to the proper spot which is better for SEO.

https://meta.trac.wordpress.org/ticket/6018


SergeyBiryukov:
Thanks for the PR! Merged in https://meta.trac.wordpress.org/changeset/11422.

PR #50, opened 2022-01-14
Trac: Changes required for the new WordPress.org header

dd32:
This is a PR for migrating Trac to the new WordPress.org headers. See https://github.com/WordPress/wporg-mu-plugins/issues/53

 - New header/footer are pulled from w.org, saved to their respective files
 - DomDocument is used to convert it from HTML5 to XHTML compatible output
 - Some cleanups are done to make the XHTML script/style tags valid for browser & Trac HTML parser compatibility
 - ~Invalid HTML from the header is fixed. https://github.com/WordPress/wporg-mu-plugins/issues/87~
 - ~There are several load-order issues on the Trac side still, resulting in JS errors~ jQuery is not loaded for the Header, instead relying upon the jQuery 1.x loaded by Trac.
 - Search is redirected to Trac Search
 - Skip link is directed to the correct Trac element


dd32:
TESTTRAC

dd32:
Turned out that the jQuery issues were caused by the WordPress.org headers loading jQuery 3.x, but Trac using (and bundling) jQuery 1.x.

I've updated the script to exclude the jQuery `<script>` and everything seems to be working.

PR #53, opened 2022-01-31
A new 'clean' design for the login screen.

dd32:
See https://github.com/WordPress/wporg-news-2021/issues/61

This is put off for now, but I'm filing this draft PR as a future starting point.

Notes:
 - wp-login.php styling has been taken care of, by starting with a fresh slate for everything.
   - See the actions hooked at `-999` in `functions-wp-login-overrides.php` for the wacky things you have to do to remove the contents of the core `login_header()` function.
 - postcss for modularised CSS, and starting from scratch means simpler styles. These styles are not at all cleaned up at all.
 - Errors and informational callouts are not styled yet.
 - The CSS is not clean what so ever.
 - A bunch of this was me learning how to setup postcss.

Personal thoughts: The login screen is much cleaner. It would actually be nice to go this way, even if we remove the header from the page.

Screenshots:

Regular Login screen:
screencapture-login-wordpress-org-2022-01-31-16_40_56

wp-login.php (Note different label style compared to above, an "okay" edgecase):
screencapture-login-wordpress-org-wp-login-php-2022-01-31-16_41_27

Registration screen:
screencapture-login-wordpress-org-register-2022-01-31-16_41_55

Registration "That username is already in use, is it you?" (unstyled)
screencapture-login-wordpress-org-register-2022-01-31-16_41_38

post-register, pre-email-confirmed screen. Top text is a callout saying check your email, unstyled.
screencapture-login-wordpress-org-register-create-profile-2022-01-31-16_42_06


dd32:
Also, just for reference, here's it without menu items in the top-nav, and then without the header at all (which I prefer)
screencapture-login-wordpress-org-2022-01-31-16_54_17

No header:
screencapture-login-wordpress-org-2022-01-31-16_55_41

PR #54, opened 2022-02-01
Fix too small font on language selection under project

kharissulistiyo:
https://meta.trac.wordpress.org/ticket/6066

Note: 

15px may be ideal to retain consistency with other table that appears on sub projects table. 

Screen Shot 2022-02-01 at 18 58 46

Screenshot taken from [WordPress 5.9.x - Development](https://translate.wordpress.org/locale/id/default/wp/dev/).


SergeyBiryukov:
Thanks for the PR! Merged a slightly modified version in https://meta.trac.wordpress.org/changeset/11513.

PR #58, opened 2022-02-03
Git👑 instead of SVN💩

babakfp:
Hi
**PLEASE** allow us to use Git instead of dog 💩 SVN. I freaking waste hours of my time every time I want to upload my project on wordpress.org. **PLEASE**, stop using old dog 💩 tools.


dd32:
Hi @babakfp,

This repo is available for submitting Pull Requests against (and to provide GitHub access to) the WordPress.org Meta repo's, whose primary location is https://meta.svn.wordpress.org / https://meta.trac.wordpress.org/. That SVN repository is also available via Git at git://meta.git.wordpress.org.

Changing Version Control Systems is not a simple process for large projects, requiring significant changes to systems, authentication, management, and how other applications interact with them. It's a significant undertaking, especially when resources can be better utilised by other ongoing projects. As a result it's not straight forward to simply "stop using it" and "use something different". It's no easier for us to migrate to CVS or Mercurial as it would be to Git.

Using SVN as the deployment location for WordPress.org assets simplifies a lot of our systems, as we don't currently require more advanced feature-sets offered by Git. That has caused a lot of systems that has been made available via WordPress.org over the last two decades to inherit SVN, from a time before GitHub was "a common thing". WordPress even pre-dates GitHub!

However, WordPress.org has been making inroads into using Git (and specifically, GitHub) for more purposes.
On WordPress.org this means that you can contribute and submit PRs via GitHub to most WordPress.org projects that are primarily SVN, and that some projects are purely GitHub from a users perspective (For example; [Gutenberg](https://github.com/wordpress/gutenberg), [WordCamp.org](https://github.com/WordPress/wordcamp.org), [Learn.WordPress.org](https://github.com/wordpress/learn), [The Pattern Directory](https://github.com/wordpress/pattern-directory), [Openverse](https://github.com/wordpress/openverse/), [Five for the Future](https://github.com/WordPress/five-for-the-future), and others - Many of these are still ultimately deployed to SVN behind the scenes, but contributors and developers don't necessarily see that).

Due to your GitHub contribution history, I'm assuming none of those projects are relevant to your interests however, and that you're complaining directly about the Plugin Directory and https://plugins.svn.wordpress.org/. Going back to the second paragraph, changing VCS's like that is not a simple task, especially when multiple "untrusted users" are involved. By "untrusted users" here I'm referring to the fact that UserA can commit to PluginA, but should not be able to commit to PluginB.

WordPress.org does not currently have a native way to allow you to use Git (or GitHub) to submit Plugin updates, primarily due to shortcomings in Authentication, but that does not mean you're forced to use SVN directly.
SVN is used as the "public deployment location", but you can use another tool to do that on your behalf - such as https://github.com/10up/action-wordpress-plugin-deploy or https://github.com/10up/action-wordpress-plugin-asset-update (Please note: WordPress.org does not officially endorse these, or similar projects, as it requires storing your WordPress.org password with a 3rd party - GitHub secrets in this case).

Allowing Plugin Developers to upload a ZIP for updates (Similar to how initial submission for review is a ZIP upload) is a feature requested by many, as many developers prefer to use non-SVN locations for development (Such as GitHub, BitBucket, GitLab, local Git, etc), and providing Git (without GitHub) is not viable for another large selection of plugin developers.  Many would also prefer WordPress.org to provide an API for uploading releases, skipping Git/GitHub/SVN/etc entirely.

Integrations between the Plugin Directory and GitHub are likely to occur at some point in time, but it's not currently on a priority roadmap. This would likely be in addition to ZIP uploads and SVN access, as no single method suits all Developers.

The Theme directory also uses SVN as a deployment location, https://themes.svn.wordpress.org/, with the upload process being [via a form on WordPress.org](https://wordpress.org/themes/upload/) - historically no SVN access was granted. That changed in recent months and Theme Authors can submit theme updates directly via SVN now if they choose. I'm not aware of any GitHub actions that exist for Theme Authors however.

We acknowledge that SVN isn't what many people are familiar/comfortable with today however, and that sometimes Git or GitHub features are valuable to a project - that's why Gutenberg and others are developed on GitHub.
In some ways, the stability of WordPress is something many rely upon, and for us, the stability of our tried-and-true existing SVN systems is something that those who maintain WordPress.org rely upon - We don't have unlimited human-hours to maintain yet another process adding additional complexity. 

**There will come a day where you can use GitHub, ZIPs, SVN, or APIs to manage WordPress.org assets (Plugins, Themes, etc) that's just not today** (unfortunately), but I hope the above explanation provides you alternatives you can use, and gives some background as to why SVN is still used by us.

babakfp:
Hi
Thank you @dd32 for your very long explanation.
What is the very simple and fast solution for this feature that I'm asking? We don't wanna mass with the current projects, we don't wanna mass with anything. Provide a way to use Git and GitHub instead of SVN.
We don't need the `/branches` and `/tags` folders because Git/GitHub already has these features. So we are only going to have 2 folders in our project. `/assets` and `/trunk`.
So all wordpress.org is going to do is to get the current/newest package and serve it to the goal user. (I'm probably missing a few things here🥲)
Being able to use Git and GitHub helps a lot. It is worth the hours to put in to make it happen. Even if you create a poll, you will see that people want Git&GitHub, not SVN.
Thank you🌸 (waiting for your response...)

dd32:
tl;dr of it is, use https://github.com/10up/action-wordpress-plugin-deploy if you wish to use a GitHub action without any SVN interaction from yourself for plugin deploys - If you are comfortable saving your WordPress.org password in a GitHub secret.

Here's a WordPress.org plugin that's hosted on the WordPress GitHub organisation that uses it:

https://github.com/WordPress/theme-check/blob/master/.github/workflows/deploy.yml

There's no SVN interaction, [tag a release on Github](https://github.com/WordPress/theme-check/releases), and [it automatically gets pushed over](https://plugins.trac.wordpress.org/changeset/2638893/) to the [WordPress.org plugin directory](https://wordpress.org/plugins/theme-check).

dd32:
Just noting that I've filed https://meta.trac.wordpress.org/ticket/6087 (and https://meta.trac.wordpress.org/ticket/6086) for tracking of an official GitHub integration.

PR #59, opened 2022-02-08
Remove `[WordPress.org Forums]` prefix from bbPress subscription emails

dd32:
This removes `[WordPress.org Forums]` prefix from bbPress subscription notifications, and prefix the forum instead.

This results in two different style emails:
```
From: "WordPress.org Forums" <noreply@wordpress.org>
Subject: [Plugin Name] Help! My Topic Title!
```
or
```
From: "WordPress.org Forums" <noreply@wordpress.org>
Subject: [Fixing WordPress] How do I make WordPress write posts for me instead?
```

This is based on https://meta.trac.wordpress.org/ticket/6079

This might result in some email filters breaking, if they're expecting the `WordPress.org Forums` prefix, but the user-experience is sub-par at present as it results in longer email subjects, and duplicates the sender name on the emails.

This is potentially something that should be upstreamed to bbPress instead..


dd32:
I've upstreamed this to bbPress: https://bbpress.trac.wordpress.org/ticket/3446

Some changes would still be needed on WordPress.org, to replace the forum name with the Term subscribed to.

dd32:
Merged in https://meta.trac.wordpress.org/changeset/11653

PR #60, opened 2022-02-09
Openverse: Update the environment to use composer for managing required `mu-plugins`.

ryelle:
Update the Openverse theme setup process to use `composer` for installing the required `mu-plugins`. Openverse needs both the [wporg-mu-plugins](https://github.com/WordPress/wporg-mu-plugins/) (for the header/footer) & [mu-plugins/pub](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub) (for the locales library). Now these are both installed in a local `mu-plugins` folder, which is gitignored (along with `vendor`, a composer artifact). Neither of these folders should be synced to SVN.

This fixes an issue where syncing a single file causes issues with docker - either outright breaking the boot process, or just creating duplicate empty files - by mapping the whole `mu-plugins` folder and hand-creating the loader file.

This also adds steps for setting up the locale database so that the `WordPressdotorg\Locales\*` functions can be used.

See discussion in #neso on a8c slack for more background & troubleshooting.

cc @sarayourfriend

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
@@ -46,7 +46,7 @@ Follow these steps to set up a local playground for the theme:
     5.  `cd` back to the Openverse theme directory at `../..`
         (i.e. `wordpress.org/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse`).
 
-4.  Set up the locale database. The plugin was installed in the previous step, but it pulls from a separate database of locale data.
+3.  Set up the locale database. The plugin was installed in the previous step, but it pulls from a separate database of locale data.
 
     1.  Download the SQL file [wporg_locales.sql](https://raw.githubusercontent.com/WordPress/pattern-directory/trunk/.wp-env/data/wporg_locales.sql) to the theme directory.
     2.  Import the file.
@@ -54,7 +54,7 @@ Follow these steps to set up a local playground for the theme:
         $ wp-env run cli "wp db import wp-content/themes/wporg-openverse/wporg_locales.sql"
         ```
 
-5.  Tell WordPress to load the `mu-plugins`. Since these are in nested folders, they're not loaded automatically. You'll need to create a new file and `require` them.
+4.  Tell WordPress to load the `mu-plugins`. Since these are in nested folders, they're not loaded automatically. You'll need to create a new file and `require` them.
 
     1.  Create a new file `.mu-plugins/loader.php`
     2.  Add the following to this new file:
@@ -65,7 +65,7 @@ Follow these steps to set up a local playground for the theme:
         require_once WPMU_PLUGIN_DIR . '/wporg-mu-plugins/mu-plugins/skip-to/skip-to.php';
         ```
 
-6.  You can choose to set up a new environment automatically or work in an
+5.  You can choose to set up a new environment automatically or work in an
     existing environment with manual setup.
 
     **Automatic:**  
@@ -91,14 +91,14 @@ Follow these steps to set up a local playground for the theme:
     themes into the `wp-content/themes` directory. You must also load the MU
     plugins and activate them using the `mu-plugins.php` file.
 
-7.  Activate and customize the theme.
+6.  Activate and customize the theme.
 
     1.  Log into `/wp-admin`.
     2.  Under Appearance > Themes, activate the theme 'WordPress.org Openverse'.
     3.  To change the embed URL, open the customizer at Appearance > Customize
         and update the value in the 'Openverse embed' panel.
 
-8.  Test message passing.
+7.  Test message passing.
 
     1.  Change the Openverse embed to
         `/wp-content/themes/wporg-openverse/js/message_test.html`.
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
@@ -54,7 +54,18 @@ Follow these steps to set up a local playground for the theme:
         $ wp-env run cli "wp db import wp-content/themes/wporg-openverse/wporg_locales.sql"
         ```
 
-3.  You can choose to set up a new environment automatically or work in an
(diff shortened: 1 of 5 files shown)


ryelle:
I've just realized that the `mu-plugins/pub` directory is also in this repo, but I think it would be a faster setup process to have this installed in one step via composer, rather than adding an extra `mu-plugins` mapping.

coreymckrill (approved):
I didn't actually try to get the whole Openverse environment up and running, but the composer stuff looks good to me. Some of the added build steps in the readme could probably be put into a bash script like we did [for Learn](https://github.com/WordPress/learn/blob/trunk/bin/index.sh), but that's not strictly necessary.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
```diff
@@ -28,25 +29,43 @@ Follow these steps to set up a local playground for the theme:
...
         $ npm run build
         ```
-    5.  `cd` back to the Openverse theme directory at `..`
+    5.  `cd` back to the Openverse theme directory at `../..`
         (i.e. `wordpress.org/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse`).
 
-3.  You can choose to set up a new environment automatically or work in an
+3.  Set up the locale database. The plugin was installed in the previous step, but it pulls from a separate database of locale data.
+
+    1.  Download the SQL file [wporg_locales.sql](https://raw.githubusercontent.com/WordPress/pattern-directory/trunk/.wp-env/data/wporg_locales.sql) to the theme directory.
```
zackkrida:
```suggestion
    1.  Download the SQL file [wporg_locales.sql](https://raw.githubusercontent.com/WordPress/pattern-directory/trunk/.wp-env/data/wporg_locales.sql) to the theme directory.
    ```bash
    curl -O https://raw.githubusercontent.com/WordPress/pattern-directory/trunk/.wp-env/data/wporg_locales.sql
    ```
```

Could be nice to suggest a download command, unless we think it's too risky to suggest curl-ing random files.

coreymckrill: Could also just include the file in the repo somewhere. It doesn't change very often, and it probably doesn't need to be up to date for most testing purposes anyway.

zackkrida:
Worked for me! Only 'gotcha' was I needed to make sure to use node@14 for the npm steps.

CleanShot 2022-02-09 at 20 29 05@2x

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
```diff
@@ -28,25 +29,43 @@ Follow these steps to set up a local playground for the theme:
...
+    2.  Import the file.
+        ```bash
+        $ wp-env run cli "wp db import wp-content/themes/wporg-openverse/wporg_locales.sql"
+        ```
+
+4.  Tell WordPress to load the `mu-plugins`. Since these are in nested folders, they're not loaded automatically. You'll need to create a new file and `require` them.
+
+    1.  Create a new file `.mu-plugins/loader.php`
```
sarayourfriend:
```suggestion
    1.  Create a new file `./mu-plugins/loader.php`
```

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
```diff
@@ -28,25 +29,43 @@ Follow these steps to set up a local playground for the theme:
...
         (i.e. `wordpress.org/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse`).
 
-3.  You can choose to set up a new environment automatically or work in an
+3.  Set up the locale database. The plugin was installed in the previous step, but it pulls from a separate database of locale data.
+
+    1.  Download the SQL file [wporg_locales.sql](https://raw.githubusercontent.com/WordPress/pattern-directory/trunk/.wp-env/data/wporg_locales.sql) to the theme directory.
+    2.  Import the file.
+        ```bash
+        $ wp-env run cli "wp db import wp-content/themes/wporg-openverse/wporg_locales.sql"
+        ```
```
sarayourfriend: The order of this command is a little confusing. By this point in the instructions the reader won't have run `wp-env start` yet so this will fail. Should this be updated to something like:

```suggestion
    2.  After completing step 5 below to set up a local WordPress installation, import the `.sql` file into the local WordPress database. For example, using `wp-env`:
        ```bash
        $ wp-env run cli "wp db import wp-content/themes/wporg-openverse/wporg_locales.sql"
        ```
```

sarayourfriend (approved):
This is great! Thank you so much for improving this process. I think the Openverse contributors can iterate on this to add a bash script and stuff to further simplify the various files and changes that are being done manually right now. 🚀

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
ryelle: > Could be nice to suggest a download command, unless we think it's too risky to suggest curl-ing random files.

I like that. It's already recommending downloading a random file, so why not 😆 

> Could also just include the file in the repo somewhere.

Since this repo is synced to/from the SVN repo, I don't think it should be here, but maybe in [WordPress/wporg-repo-tools](https://github.com/WordPress/wporg-repo-tools)? I almost threw it into a gist, but figured linking off to the pattern-directory was a little more legitimate.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
ryelle: ah, good catch, I already had it up while I was testing. I'll just move this step entirely.

sarayourfriend:
@ryelle what's left for getting this merged? The additional locale plugin is a blocker for Openverse deploying changes as we'll need to update how the locale is forwarded to the iframe before we can deploy the latest version of the frontend.

zackkrida (approved):

ryelle:
@sarayourfriend @zackkrida Sorry about the delay! I've merged it in [r11559](https://meta.trac.wordpress.org/changeset/11559)/https://github.com/WordPress/wordpress.org/commit/80236410b264e5f7fc28a0552e4d4cb6efa0a630.

sarayourfriend:
Thanks!

PR #61, opened 2022-02-15
Update locale forwarding via path

sarayourfriend:
https://meta.trac.wordpress.org/ticket/6105

Forwards the locale via the iframe's path rather than relying on a cookie.

To test, follow setup instructions in the README of the wporg-openverse theme. Set the "Use path based locale forwarding" setting to on by checking the box in the customizer. And redirect the iframe to use staging or localhost as production does not yet support path based locales.

Next, to test that the locales are working, replace [the call to `get_locale()` with a `wp_locale`](https://github.com/WordPress/wordpress.org/pull/61/files#diff-0456ab0b8c830deeba20a153962f28a381c93ccde52e9dce715009a7cb3381c0R90) string (like `pt_BR` etc) to verify that it is working to forward the locale as expected (as far as I know there's no way to test this "for real" locally as subdomain locale choosing isn't supported in the local wp-env when I tried it).

Be sure to navigate around the various pages of the application. Pay special attention to the URL being shown in the host frame. It should never include the locale slug.

Next, test it with the `en_US` locale and make sure everything works, again paying attention to the locale slug.

Finally, turn off the path based locale forwarding setting and switch the frame back to production. Ensure that it continues to work as expected, especially the host frame URL updates. I'm not sure how to test other locales other than again, manually replacing the call to `get_locale()` with a locale string of choice. I noticed that it takes at least one page navigation for the locale to actually propagate via the cookie. This is the same as the behavior on `trunk`.

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/js/iframe_nav.js
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/js/iframe_nav.js
@@ -5,7 +5,7 @@ document.addEventListener('DOMContentLoaded', () => {
     .toLocaleLowerCase()
     .replace(openverseSubpath, '')
     .replace(/^\/$/, ''); // Remove Openverse site subpath
-  iframePath = `${openverseUrl}${iframePath}${query}`; // Add domain and query
+  iframePath = `${openverseUrl}${localeSlug ? `/${localeSlug}` : ''}${iframePath}${query}`; // Add domain and query
 
   console.log(`Navigating iframe to ${iframePath}`);
   const iframe = document.getElementById('openverse_embed');
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/js/iframe_nav.js
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/js/iframe_nav.js
@@ -5,7 +5,12 @@ document.addEventListener('DOMContentLoaded', () => {
     .toLocaleLowerCase()
     .replace(openverseSubpath, '')
     .replace(/^\/$/, ''); // Remove Openverse site subpath
-  iframePath = `${openverseUrl}${localeSlug ? `/${localeSlug}` : ''}${iframePath}${query}`; // Add domain and query
+  const localePart = localeSlug
+    ? openverseUrl.endsWith('/')
+      ? localeSlug
+      : `/${localeSlug}`
+    : '';
+  iframePath = `${openverseUrl}${localePart}${iframePath}${query}`; // Add domain and query
 
   console.log(`Navigating iframe to ${iframePath}`);
   const iframe = document.getElementById('openverse_embed');
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/.gitignore
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/.gitignore
@@ -3,3 +3,4 @@
 
 mu-plugins/
 vendor/
+wporg_locales.sql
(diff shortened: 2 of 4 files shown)


sarayourfriend:
cc @WordPress/openverse-developers

sarayourfriend:
Before merging this it needs the new functionality to be put behind a theme setting.

sarayourfriend:
Okay this is ready for review now. @WordPress/openverse-developers please take a look at this, thanks!

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/js/iframe_nav.js
```diff
@@ -5,7 +5,7 @@ document.addEventListener('DOMContentLoaded', () => {
     .toLocaleLowerCase()
     .replace(openverseSubpath, '')
     .replace(/^\/$/, ''); // Remove Openverse site subpath
-  iframePath = `${openverseUrl}${iframePath}${query}`; // Add domain and query
+  iframePath = `${openverseUrl}${localeSlug ? `/${localeSlug}` : ''}${iframePath}${query}`; // Add domain and query
```
obulat:
```suggestion
  iframePath = `${openverseUrl}${localeSlug ? `${localeSlug}` : ''}${iframePath}${query}`; // Add domain and query
```

obulat: This slash causes the `openverse_embed` url have 2 slashes: `<iframe id="openverse_embed" src="http://localhost:8443//ru">`.
Without it, everything works, and if I use the branch from the[ translation banner fix PR](https://github.com/WordPress/openverse-frontend/pull/904), I can see the banner!

obulat (approved):

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/js/iframe_nav.js
sarayourfriend: Huh, I don't see that locally. For example, if I change the `get_locale` call to just be `pt_BR` I see this logged:

```
Navigating iframe to https://search-staging.openverse.engineering/pt-br
```

sarayourfriend: Ah, I think this depends on what you set the `ov_embed_url` value to in customizer. If you add a trailing slash to that then it doubles the slash.

The default value does not include a trailing slash and I've just been testing by changing the subdomain to `search-staging` so that's what I haven't seen the double slash locally.

I guess we could make the code detect whether a the trailing slash is there and add it conditionally.

obulat: The way I was testing it is by setting the language of the site in `wp-admin`. So, the top bar would display in the language I selected.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/js/iframe_nav.js
```diff
@@ -5,7 +5,12 @@ document.addEventListener('DOMContentLoaded', () => {
     .toLocaleLowerCase()
     .replace(openverseSubpath, '')
     .replace(/^\/$/, ''); // Remove Openverse site subpath
-  iframePath = `${openverseUrl}${localeSlug ? `/${localeSlug}` : ''}${iframePath}${query}`; // Add domain and query
+  const localePart = localeSlug
+    ? openverseUrl.endsWith('/')
```
obulat, line 9: Nice!

obulat:
We can also remove step 3 from the README instructions, as the mu-plugin is already set up, at least it is for me?

sarayourfriend:
> We can also remove step 3 from the README instructions, as the mu-plugin is already set up, at least it is for me?

That's just you :slightly_smiling_face: On a fresh clone they're set up by `composer install`.

PR #69, opened 2022-03-25
Add optional attributes to the wpcredits shortcode

coreymckrill:
Adds `class` and `separator` attributes to the shortcode callback function so that the string of props names can better be styled to match the new design for the WordPress.org News site.

The defaults for these attributes are set so that all existing uses of the shortcode on the News site will be able to match the design without having to go back and edit a lot of posts.

See:
* https://github.com/WordPress/wporg-news-2021/issues/355 
* https://github.com/WordPress/wporg-news-2021/pull/363

### To test

1. Upload this to your sandbox.
2. While sandboxing wp.org, try viewing [a major release post](https://wordpress.org/news/2022/01/josephine/) and finding the credits.
3. You can try out different values for the attributes without editing the actual post by changing the defaults and re-uploading.

### DIFF
--- a/api.wordpress.org/public_html/core/credits/shortcode.php
+++ b/api.wordpress.org/public_html/core/credits/shortcode.php
@@ -4,7 +4,29 @@ defined( 'WPINC' ) or exit;
 
 add_shortcode( 'wpcredits', 'wporg_wordpress_credits_shortcode' );
 
-function wporg_wordpress_credits_shortcode( $attrs, $content = null ) {
+/**
+ * Shortcode callback to render a list of props names, given a WP version.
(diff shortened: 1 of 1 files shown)


iandunn:
I think this was auto-closed by the sync script, since the branch was created in this repo instead of a fork.

I can still review it here, though, no need to recreate.

iandunn (commented):
This looks good 👍🏻

coreymckrill:
Oops, didn't realize it needed to be a fork.

coreymckrill:
Committed in https://meta.trac.wordpress.org/changeset/11711/

PR #73, opened 2022-06-02
Add margin before Download button in plugin header

aurooba:
Hallo, this is a tiny fix for something that has been bothering me. When the Plugin title is a bit long on desktop, it bumps right into the `Download` button like so:
Screen Shot 2022-06-02 at 4 43 15 PM

I've added a little margin (to match the margin between the thumbnail and title) so this doesn't happen:

Screen Shot 2022-06-02 at 4 43 46 PM

I haven't rebuilt the style files yet, but if this change is okay, do let me know and I'll look up the docs to build the files and stuff. :)


StevenDufresne:
Thanks for this PR! I've made the change and deployed. 
PRs here are not merged here since this is a code mirror.

PR #75, opened 2022-06-17
WIP: Update the profile handler to log last contribution

### LINKED ISSUE #189: Update Usermeta with latest 5ftf contribution.
We want to query for inactive accounts in #27 and #210. In order to do so we should:

- [ ] Wait until #119 is decided, since that determines the criteria to use
- [ ]  Create safe list of activities that qualify as 5ftF contributions (activity in github.com/wordpress repos, posts/comments on Make blogs, WordCamp organizing, etc)
- [ ]  Add a "latest 5ftf contribution" meta field to usermeta. Update that with `time()` whenever a 5ftf contribution is added to their profile.
    - [ ] This should be centralized in wporg-profiles-activity-handler, rather than each individual notifier having to do it. 
    - [ ] Don't update it for non-contribution activity, like opening posts on support forums (and commenting on posts they opened), committing to personal plugins/themes, etc.
- [ ] Some things like GitHub activity wouldn't be covered by the Profiles handler, though, and would need do be updated separately.
    - [ ] document that in the centralized code, so future devs will know to update new sources that are added.

This would be more performant than having to query all their activity and filter it. It can also be used in https://github.com/WordPress/five-for-the-future/issues/129.

StevenDufresne:
Closes: https://github.com/WordPress/five-for-the-future/issues/189

This is a very early PR, mostly designed to discuss the strategy behind the code since I'm not incredibly familiar with all the inputs/outputs.

Also, the structure and format of these changes are not expected to be final, just designed to initiate the conversation.

## Notes:
- Can we update the meta value in `handle_activity`?
  - Are there any other paths that don't come through `handle_activity`? 
- I think we could use an object, called `SavedActivity` here, that returns whether the activity is _contribution_ worthy.
  - That would allow us to keep the logic specific to the input
    - IE: forum activity handler determines whether it's worthy
  - Does this make sense? Are there better approaches?

### DIFF
--- a/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
+++ b/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
@@ -29,7 +29,26 @@ defined( 'ABSPATH' ) or die();
 
 if ( ! class_exists( 'WPOrg_Profiles_Activity_Handler' ) ) {
 
+	class SavedActivity {
+		public $id;
+		public $type;
+		public $is_contribution;
+
+		function __construct( $id, $type, $is_contribution ) {
+			$this->id = $id;
+			$this->type = $type;
+			$this->is_contribution = $is_contribution;
+		}
+	}
+
 	class WPOrg_Profiles_Activity_Handler {
+
+		/**
+		 *  Name of meta key that tracks last contribution as a unix timestamp.
+		 *
+		 * @var string
+		 */
+		const last_contribution_meta_key = 'wporg_5ftf_last_contribution';
 
 		/**
 		 * Constructor.
@@ -204,7 +223,7 @@ if ( ! class_exists( 'WPOrg_Profiles_Activity_Handler' ) ) {
 						$activity_id = bp_activity_add( $this->sanitize_activity( $_POST ) );
 						break;
 					case 'plugin':
-						$activity_id = $this->handle_plugin_activity();
+						$activity = $this->handle_plugin_activity();
 						break;
 					case 'theme':
 						$activity_id = $this->handle_theme_activity();
@@ -232,6 +251,10 @@ if ( ! class_exists( 'WPOrg_Profiles_Activity_Handler' ) ) {
 					throw new Exception( $activity_id );
 				} elseif ( false === $activity_id || intval( $activity_id ) <= 0 ) {
 					throw new Exception( '-1 Unable to save activity' );
+				}
+
+				if( $activity instanceof SavedActivity && $activity->is_contribution ) {
+					self::update_last_contribution_meta( $_POST['user_id'] );
 				}
 
 				$response = '1';
@@ -399,7 +422,7 @@ if ( ! class_exists( 'WPOrg_Profiles_Activity_Handler' ) ) {
 					'hide_sitewide'     => false,
 				);
 
-				return bp_activity_add( $args );
+				return new SavedActivity( bp_activity_add( $args ), 'forum_activity', true );
 
 			} elseif ( in_array( $type, array( 'forum_topic_remove', 'forum_reply_remove' ) ) ) {
 				// Remove activity related to a topic or reply.
@@ -443,7 +466,7 @@ if ( ! class_exists( 'WPOrg_Profiles_Activity_Handler' ) ) {
 				'hide_sitewide'     => false,
 			);
 
-			return bp_activity_add( $args );
+			return new SavedActivity( bp_activity_add( $args ), 'plugin_activity', true );
 		}
 
 		/**
@@ -846,6 +869,16 @@ if ( ! class_exists( 'WPOrg_Profiles_Activity_Handler' ) ) {
 			}
 
 			return $user_case_args;
+		}
+
+		/**
+		 * Updates meta value to current timestamp indicating the user's last contribution.
+		 * 
+		 * @return int|bool result of update_user_meta();
+		 */
+		protected function update_last_contribution_meta() {
+			$user = self::get_user( $_POST['user'] );
+			return update_user_meta( $user->ID, self::last_contribution_meta_key, time() );
 		}
 
 	} /* /class WPOrg_Profiles_Activity_Handler */
--- a/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
+++ b/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
@@ -422,7 +422,7 @@ if ( ! class_exists( 'WPOrg_Profiles_Activity_Handler' ) ) {
 					'hide_sitewide'     => false,
 				);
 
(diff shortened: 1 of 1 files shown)


profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
```diff
@@ -29,8 +29,27 @@
 
 if ( ! class_exists( 'WPOrg_Profiles_Activity_Handler' ) ) {
 
+	class SavedActivity {
```
StevenDufresne, line 32: This wouldn't live here. We could move it to its own file.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
```diff
@@ -234,6 +253,10 @@ public function handle_activity() {
 					throw new Exception( '-1 Unable to save activity' );
 				}
 
+				if( $activity instanceof SavedActivity && $activity->is_contribution ) {
```
StevenDufresne, line 256: The `SavedActivity` class could also be used for all the code paths in the `handle_` functions by creating other properties for the other returns types. Ie ($activity->error_msg, etc...)

profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
iandunn, line 32: If we go the OOP route, another option might be to make a base `ActionHandler` abstract class, that has an abstract `is_5ftf_contribution()` method. That base method could just use [the simple array safelist](https://github.com/WordPress/wordpress.org/pull/75#issuecomment-1159238916), but for more complex ones like Forums, the child class could override the method and run some logic to check _which_ forum, etc.

Then `handle_activity()` would instantiate the base class for the easy actions, and subclasses like `ForumActivityHandler` for the complex ones.

e.g.,

```php
switch ( $source ) {
	case 'forum':
		$activity_id = new ForumActivityHandler( $args );
		break;
	case 'learn':
	case 'wordcamp':
		$activity_id = new ActivityHandler( $args );
		break;
}
```

I say that because this kind of feels like a glorified array, since the `$is_contribution` logic is done elsewhere.

At the moment it still feels cleaner to me to [do it procedurally](https://github.com/WordPress/wordpress.org/pull/75#issuecomment-1159239221), but I'm open to other ideas.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
```diff
@@ -848,6 +871,16 @@ protected function handle_props_given( $post_unsafe ) {
 			return $user_case_args;
 		}
 
+		/**
+		 * Updates meta value to current timestamp indicating the user's last contribution.
+		 * 
+		 * @return int|bool result of update_user_meta();
+		 */
+		protected function update_last_contribution_meta() {
+			$user = self::get_user( $_POST['user'] );
```
iandunn, line 880: This works for most, but not all. The Slack notifier sends multiple usernames in different keys.

iandunn:
> Can we update the meta value in handle_activity?

That's what I was thinking, yeah; or somewhere "nearby".

> Are there any other paths that don't come through handle_activity?

GitHub and Trac are stored in their own db tables, but I think everything else is in BuddyPress. See `WPORG_Profiles::get_timeline()`, and `::get_code()`.

iandunn:
> use an object... keep the logic specific to the input

My initial instinct is that that feels like overkill. I was thinking of something like this:

```php
function is_5ftf_contributon( $action ) {
    $5ftf_actions = array( ... );

   return in_array( $action, $5ftf_actions, true );
}
```

...but there'll probably need to be more logic for some of them. For instance, posting replies to _someone else's post_ in the Fixing WP forums is a 5ftf contribution, but posting in a plugin support forum isn't. If there are a lot of other examples like that, then it may be cleaner to have something more abstract. 

It might help to start by gathering a list of all the distinct actions that are handled, and any relevant qualifiers (like _which_ forum was posted in).

iandunn:
>  Are there better approaches?

Not sure if it's better, but my first instinct would be to hook into the `bp_activity_add` action. That only fires when an activity is successfully added to the db. It also gives you access to the user ID (via fetching the activity ID), so you don't have to worry about the complexity of the Slack handler having multiple users.

For GitHub/Trac, it might be fine to make the `update_usermeta()` call directly in the code that inserts those into the DB, rather than trying to centralize. I haven't looked at how it's structured, though. I think the only conditions there would be to make sure that GitHub actions are under the github.com/wordpress org, and the tracs are Core/Meta (not Plugins/Themes).

StevenDufresne:
> ...but there'll probably need to be more logic for some of them. For instance, posting replies to _someone else's post_ in the Fixing WP forums is a 5ftf contribution, but posting in a plugin support forum isn't. If there are a lot of other examples like that, then it may be cleaner to have something more abstract.

Does it make sense to create these as new types in the [wporg-profiles-activity-hander](https://github.com/WordPress/wordpress.org/blob/trunk/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php)? Something like `forum_fixing_reply`, or whatever convention makes sense?

If that were the case, we could hook into `bp_activity_add` and check if the type is in the array. We would just have to ensure that the format is visible and predictable since the type strings will be duplicated in the handler and whichever plugin is hooking into `bp_activity_add`.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
StevenDufresne: Something like #78?

iandunn: Yeah, that's that what I was thinking 👍🏻 

What are your thoughts on how that turned out?

iandunn:
> create these as new types in the `wporg-profiles-activity-hander`

Do you mean the `type` that gets passed to `bp_activity_add` (like `forum_reply_create`, `plugin_create`, `trac_props_mention` , etc); or were you thinking about [the main `switch` statement](https://github.com/WordPress/wordpress.org/blob/2d3cae5f929c3c20acbeb2e6c6dd74ea4f09dc02/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php#L199) ?

profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php
StevenDufresne: I think that works fine, I like that it's decoupled.

StevenDufresne:
Will use #78 instead.

PR #78, opened 2022-06-27
WIP: Add plugin that updates meta value for 5ftf

StevenDufresne:
Addresses: https://github.com/WordPress/five-for-the-future/issues/189

# Summary
We need to store a timestamp of the last five for the future contribution to be used later to make more efficient SQL queries in [soon-to-be-written](https://github.com/WordPress/five-for-the-future/milestone/9) code. [Read More](https://github.com/WordPress/five-for-the-future/issues/189).

# Inputs
### Activities in WordPress & WordPress communities
Activities that happen on WordPress.org and other adjacent WordPress communities get added to the database via `bp_activity_add`.

Most of this logic is handled in [profiles.wordpress.org/wporg-profiles-activity-handler.php](https://github.com/WordPress/wordpress.org/blob/trunk/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php)

### Activity on GitHub
Activities that happen on GitHub get stored in `wporg_github_activity` via an [API endpoint](https://github.com/WordPress/wordpress.org/blob/trunk/api.wordpress.org/public_html/dotorg/github/activity.php).
### Trac commits
// TODO Not sure what happens here. 

# Context
In its simplest form, we need to run `update_user_meta( $user_id, 'key_name', time() )` when a valid contribution is _detected_. 

We have the option to detect contributions in 2 ways:
- Active -> When the activity is being logged to the database (used in this PR) 
  -  Maybe it's not logged but we detect it in an event
- Passive -> Using a task that runs periodically and queries for all the relevant data sources.

This PR uses an active approach and listens for filters that indicate new activities were saved to the database. 

Making sure the timestamp is up-to-date is important, seeing that it will be used to initiate downstream notifications. A cron job would theoretically introduce an edge case where a contribution could be missed, albeit temporarily.

The potential downside of using this approach is that our input list could grow horizontally and we would need to make 5ftf-related code updates to code that shouldn't know about it. In other words, we should be able to remove all 5ftf code without side effects or remaining technical debt. 

The last consideration is that this code will be the only place where 5ftf contributions are clearly defined. So while we could call `update_user_meta` wherever an activity occurs, keeping track of the contribution list (or logic) in one spot will give us the ability to audit/add/remove contributions more easily, and since missing contributions could result in _extra_ pledge correspondence, we'll want to be able to see the landscape in one frame.

# Approach

Most of the wp actions are currently stored by BuddyPress via `bp_activity_add`. After discussion with @iandunn, we've decided its probably best to hook into the BuddyPress filter `bp_activity_add`. I think we can also take that approach for all inputs by creating our own filters.

- `apply_filter( 'make_relevant_filter_name', $data )` for all inputs when activities are saved or detected
- Activate this plugin on all sites where inputs exist
- Add a filter for the relevant input in this plugin
- Handle the `update_user_meta` in this file only

# Tasks

- [ ] Identify the relevant `type`s that are considered contributions
  - Some of these will be conditional and vary per site/team
- [ ] Identify the relevant GitHub `category`
  - Some of these will be conditional and vary per site/team
- [ ] Update the contribution based on trac commits?
- [ ] other stuff, not sure yet

### DIFF
--- a/api.wordpress.org/public_html/dotorg/github/activity.php
+++ b/api.wordpress.org/public_html/dotorg/github/activity.php
@@ -71,6 +71,22 @@ function github_user_to_user_id( $user ) {
 	return $user_id ? intval( $user_id ) : false;
 }
 
+/**
+ * Save github activity to database.
+ *
+ * @param array $data
+ *
+ * @uses apply_filters() Calls 'wporg_github_added_activity'
+ */
+function save_github_activity( $data ) {
+	$wpdb->insert(
+		'wporg_github_activity',
+		$data
+	);
+
+	apply_filters( 'wporg_github_added_activity', $data );
+}
+
 $event   = $_SERVER['HTTP_X_GITHUB_EVENT'];
 $payload = get_signed_payload_or_die();
 
@@ -144,8 +160,7 @@ switch ( $event ) {
 		}
 
 		// Pushed commits.
-		$wpdb->insert(
-			'wporg_github_activity',
+		save_github_activity(
 			( $user_id     ? compact( 'user_id' )     : [] ) + // Leave user_id as null if not known.
 			( $github_user ? compact( 'github_user' ) : [] ) + // Leave github_user as null if not known.
 			[
@@ -195,8 +210,7 @@ switch ( $event ) {
 		// Can use Sender here, as they'll be the one to create/close the Issue.
 		$user_id = github_user_to_user_id( $payload->sender->login );
 
-		$wpdb->insert(
-			'wporg_github_activity',
+		save_github_activity(
 			( $user_id ? compact( 'user_id' ) : [] ) + // Leave user_id as null if not known.
 			[
 				'ts'          => gmdate( 'Y-m-d H:i:s' ),
@@ -243,8 +257,7 @@ switch ( $event ) {
 			$event = 'pr_merge';
 		}
 
-		$wpdb->insert(
-			'wporg_github_activity',
+		save_github_activity(
 			( $user_id ? compact( 'user_id' ) : [] ) + // Leave user_id as null if not known.
 			[
 				'ts'          => gmdate( 'Y-m-d H:i:s' ),
--- a/api.wordpress.org/public_html/dotorg/github/activity.php
+++ b/api.wordpress.org/public_html/dotorg/github/activity.php
@@ -84,7 +84,7 @@ function save_github_activity( $data ) {
 		$data
 	);
 
-	apply_filters( 'wporg_github_added_activity', $data );
+	do_action( 'wporg_github_added_activity', $data );
 }
 
 $event   = $_SERVER['HTTP_X_GITHUB_EVENT'];
--- a/profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
+++ b/profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
@@ -18,7 +18,6 @@ if ( 'staging' !== wp_get_environment_type() || 'cli' !== php_sapi_name() ) {
 }
 
 const TEST_USERNAME = 'dufresnesteven';
-const TEST_USERID   = '17657928';
 
 /** @var array $args */
 main( $args[0] );
@@ -27,9 +26,9 @@ function main( string $case ) : void {
 
 	require_once dirname( __DIR__ ) . '/wporg-5ftf-activity-handler.php';
 
-	call_user_func( __NAMESPACE__ . "\\test_$case", WPOrg_5ftf_Activity_Handler::get_instance() );
+	$user = get_user_by( 'slug', TEST_USERNAME );
 
-	// echo "\nThere should be new activity on https://profiles.wordpress.org/$user->user_nicename/ \n";
+	call_user_func( __NAMESPACE__ . "\\test_$case", WPOrg_5ftf_Activity_Handler::get_instance(), $user );
 }
 
 function expect( $fail, $result ) {
@@ -50,19 +49,19 @@ function expectFalse( $result ) {
  * Tests the handling of GitHub activities
  */
 
-function test_github_filter( WPOrg_5ftf_Activity_Handler $handler ) : void {
-	test_github_filter_success( $handler );
+function test_github_filter( WPOrg_5ftf_Activity_Handler $handler, $user ) : void {
+	test_github_filter_success( $handler, $user );
 	test_github_filter_no_user_id( $handler );
-	test_github_filter_unsupported_category( $handler );
+	test_github_filter_unsupported_category( $handler, $user );
 }
 
-function test_github_filter_success( WPOrg_5ftf_Activity_Handler $handler ) : void {
+function test_github_filter_success( WPOrg_5ftf_Activity_Handler $handler, $user ) : void {
 	expectTrue(
 		$handler->handle_github_activity(
 			array(
 				'category' => 'pr_merged',
 				'repo'     => 'wordpress.org',
-				'user_id'  => TEST_USERID,
+				'user_id'  => $user->ID,
 			)
 		)
 	);
@@ -79,13 +78,13 @@ function test_github_filter_no_user_id( WPOrg_5ftf_Activity_Handler $handler ) :
 	);
 }
 
-function test_github_filter_unsupported_category( WPOrg_5ftf_Activity_Handler $handler ) : void {
+function test_github_filter_unsupported_category( WPOrg_5ftf_Activity_Handler $handler, $user ) : void {
 	expectFalse(
 		$handler->handle_github_activity(
 			array(
 				'category' => 'not_supported',
 				'repo'     => 'wordpress.org',
-				'user_id'  => TEST_USERID,
+				'user_id'  => $user->ID,
 			)
 		)
 	);
--- a/profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
+++ b/profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
@@ -17,7 +17,7 @@ if ( 'staging' !== wp_get_environment_type() || 'cli' !== php_sapi_name() ) {
 	die( 'Error: Wrong environment.' );
 }
 
-const TEST_USERNAME = 'dufresnesteven';
+const TEST_USERNAME = 'metatestaccount';
 
 /** @var array $args */
 main( $args[0] );
@@ -46,7 +46,6 @@ function expectTrue( $result ) {
 function expectFalse( $result ) {
 	expect( substr( $result, 0, 2 ) !== '-1', $result );
 }
-
 
 /**
  * Tests the handling of `bp_activity_add` activities
--- a/profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
+++ b/profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
@@ -8,6 +8,7 @@
  */
 
 namespace WordPressdotorg\WPORG_5ftf_Activity_Notifier\Tests;
+
 use WPOrg_5ftf_Activity_Handler;
 
 ini_set( 'display_errors', 'On' ); // won't do anything if fatal errors
@@ -17,7 +18,7 @@ if ( 'staging' !== wp_get_environment_type() || 'cli' !== php_sapi_name() ) {
 }
 
 const TEST_USERNAME = 'dufresnesteven';
-const TEST_USERID = '17657928';
+const TEST_USERID   = '17657928';
 
 /** @var array $args */
 main( $args[0] );
@@ -28,22 +29,21 @@ function main( string $case ) : void {
 
 	call_user_func( __NAMESPACE__ . "\\test_$case", WPOrg_5ftf_Activity_Handler::get_instance() );
 
-
-	//echo "\nThere should be new activity on https://profiles.wordpress.org/$user->user_nicename/ \n";
+	// echo "\nThere should be new activity on https://profiles.wordpress.org/$user->user_nicename/ \n";
 }
 
 function expect( $fail, $result ) {
-	if( $fail) {
-		echo 'Test Error: ' . PHP_EOL . ' Received:' . PHP_EOL . $result  . PHP_EOL;
+	if ( $fail ) {
+		echo 'Test Error: ' . PHP_EOL . ' Received:' . PHP_EOL . $result . PHP_EOL;
 	}
 }
 
 function expectTrue( $result ) {
-	expect( ( $result !== true || substr($result, 0, 1) !== '1' ), $result);
+	expect( ( $result !== true || substr( $result, 0, 1 ) !== '1' ), $result );
 }
 
 function expectFalse( $result ) {
-	expect(substr($result, 0, 2) !== '-1', $result);
+	expect( substr( $result, 0, 2 ) !== '-1', $result );
 }
 
 /**
@@ -57,24 +57,36 @@ function test_github_filter( WPOrg_5ftf_Activity_Handler $handler ) : void {
 }
 
 function test_github_filter_success( WPOrg_5ftf_Activity_Handler $handler ) : void {
-	expectTrue( $handler->handle_github_activity( [
-		"category" => "pr_merged",
-		"repo" => "wordpress.org",
-		"user_id" => TEST_USERID
-	] ) );
+	expectTrue(
+		$handler->handle_github_activity(
+			array(
+				'category' => 'pr_merged',
+				'repo'     => 'wordpress.org',
+				'user_id'  => TEST_USERID,
+			)
+		)
+	);
 }
 
 function test_github_filter_no_user_id( WPOrg_5ftf_Activity_Handler $handler ) : void {
-	expectFalse( $handler->handle_github_activity( [
-		"category" => "pr_merged",
-		"repo" => "wordpress.org",
-	] ) );
+	expectFalse(
+		$handler->handle_github_activity(
+			array(
+				'category' => 'pr_merged',
+				'repo'     => 'wordpress.org',
+			)
+		)
+	);
 }
 
 function test_github_filter_unsupported_category( WPOrg_5ftf_Activity_Handler $handler ) : void {
-	expectFalse( $handler->handle_github_activity( [
-		"category" => "not_supported",
-		"repo" => "wordpress.org",
-		"user_id" => TEST_USERID
-	] ) );
+	expectFalse(
+		$handler->handle_github_activity(
+			array(
+				'category' => 'not_supported',
+				'repo'     => 'wordpress.org',
+				'user_id'  => TEST_USERID,
+			)
+		)
+	);
 }
--- a/profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
+++ b/profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
@@ -22,6 +22,8 @@ const TEST_USERNAME = 'dufresnesteven';
 /** @var array $args */
 main( $args[0] );
 
+
+// Test Helper functions
 function main( string $case ) : void {
 
 	require_once dirname( __DIR__ ) . '/wporg-5ftf-activity-handler.php';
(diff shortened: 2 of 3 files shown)


profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,65 @@
...
+				self::update_last_contribution_meta( $args['user_id'] );
+			}
+		}
+
+		/**
+		 * Returns whether action is considered a contribution.
+		 */
+		public function is_5ftf_contribution( $action ) {
+			$wordpress_actions = array( 'blog_post_create' );
+			$wordcamp_actions  = array( 'wordcamp_speaker_add', 'wordcamp_organizer_add' );
```
StevenDufresne: Maybe it makes sense to keep different arrays for readability, maybe the prefix is enough,

iandunn: That's a good thought. Either way works for me 👍🏻

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,65 @@
...
+
+			$valid_actions = array_merge( $wordpress_actions, $wordcamp_actions );
+
```
iandunn, line 70: There could be more stuff here for the complex cases, like 

```php
if ( 'forum_topic_reply' === $action ) {
    return is_5ftf_forum_contribution( $action, $forum_id, $etc );
}
```

... which calls something like

```php
function is_5ftf_forum_contribution( $action, $forum, $etc ) {
   // if creating a topic, return false
   // if replying to their own topic, return false
   // if plugin/theme forum, return false
   // return true
}
```

there might be a better way, that's just what feels the most straight-forward at the moment. what do you think?

StevenDufresne: I'm a bit concerned that it may become difficult to read if we are doing a lot of source information validation (like checking for a specific forum `id`, running extra queries and running additional logical statements). 

My proposed alternative, which I was trying to portray in [our conversation on the other pull request](https://github.com/WordPress/wordpress.org/pull/75#issuecomment-1168090751), is that within the main profile activity handler, within the specific handle functions (ie: `handle_forum_activity`), we could create new `type`s that we can filter for in this plugin. This doesn't remove the need to check for source information, that would still have to happen in the main profile activity handler, but it at least keeps source information and validation where we already do some of that and have access to _all_ the activity information, not only the information that is passed to `bp_activity_add`. 

Additionally, this plugin will probably become the go-to reference to determine what is considered a contribution and what is not as it will be the only place where the safelist will exist, so it's important we can read this file quickly & easily. 

So, as an example:

**Case**:
"For instance, posting replies to someone else's post in the Fixing WP forums is a 5ftf contribution, but posting in a plugin support forum isn't."

**Example**:
[Source](https://github.com/WordPress/wordpress.org/blob/2d3cae5f929c3c20acbeb2e6c6dd74ea4f09dc02/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php#L306)
```
private function handle_forum_activity() {
  $user = self::get_user( $_POST['user'] );
  $type = '';

  // replying to someone else's post in fixing wp forum
  if() {
   $type = 'forum_fixing_WP_reply`
  }

  ...
  $args = array(
    'component'         => 'forums',
    'type'              => $type,
  );

  ...
}
```

Then in this plugin, look for that type.
```
$forum_actions  = array( 'forum_fixing_WP_reply' );
```

I'm sure there's a better convention for the `type` string, but it's just an example. 

Now, I'm not sure this is the best, and I don't know what implications it would have to change the type... Maybe there is downstream logic that would break if we did it. I would have to investigate more.

StevenDufresne: In looking at the GitHub notification code, I think we can apply a filter and handle it here. We'll need logical statements for that in this plugin anyways, so yeah. Your suggestion seems to be the best approach. 👍

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,65 @@
...
+		 *
+		 * @var string
+		 */
+		const last_contribution_meta_key = 'wporg_5ftf_last_contribution';
+
+		/**
+		 * Constructor.
+		 */
+		public function __construct() {
+			add_filter( 'bp_activity_add', array( $this, 'handle_contribution' ) );
```
iandunn: We'll need to remember to only activate this on profiles.w.org, rather than network-activating it on the entire buddypress.org network. If it were network-activated there may be some unintended side-effects for buddypress.org, bbpress.org, etc. Probably not w/ the current code, but maybe in the future. It's only needed on profiles.w.org, so best to keep it there.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
iandunn: > GitHub notification code... apply a filter and handle it here

🤔 The GitHub code runs on w.org, but the activity handler runs on buddypress/profiles; are you thinking this plugin would run on both of those networks?

I like that idea, since the same function could be used for DRYness.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,86 @@
...
+		 * @var string
+		 */
+		const last_contribution_meta_key = 'wporg_5ftf_last_contribution';
+
+		/**
+		 * Constructor.
+		 */
+		public function __construct() {
+			add_filter( 'bp_activity_add', array( $this, 'handle_activity' ) );
+			add_filter( 'wporg_github_added_activity', array( $this, 'handle_github_activity' ) );
```
iandunn:
```suggestion
			if ( 'profiles.wordpress.org' === site_url() ) {
				add_filter( 'bp_activity_add', array( $this, 'handle_activity' ) );
			} elseif ( str_ends_with( 'wordpress.org', site_url() ) {
				add_filter( 'wporg_github_added_activity', array( $this, 'handle_github_activity' ) );
			}
```

That would help self-document the underlying assumptions.

That `elseif` would probably cover all the sites in the w.org multi-network in practice, but it's not really future proof, since we could add something like `doaction.org` in the future. It also matches sites in other networks that end in the domain, which we might want to activate the plugin on in the future.

Since we can't use any of the constants in the shared directory, so we might want to add something like an `IS_WPORG_MULTINETWORK` constant to `wp-config.php`.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
iandunn: > it may become difficult to read if we are doing a lot of source information validation ...  

Yeah, that's a good concern. I'm not sure about changing the `type` that gets passed to `bp_activity_add()`, though, since we'd lose the ability to perform actions on them all as a group (eg, `wp bp activity list --type=topic_forum_reply`). We'd also have to retroactively update all the existing items with their new types. It seems like it might be using `type` for something it wasn't intended for.

We could add that extra data as a meta item on the activity though, and then `is_5ftf_contribution()` could lookup that value for its logic?

iandunn: Er, extra meta fields might not be great either, since it'd add a lot of data to the db, and it feels a bit unnecessary.

Functions like `handle_forum_activity()` are already pretty long and do too many things, so adding more to them might not help avoid the "difficult to read" problem.

🤔 

I like your point about keeping the 5ftf check in the same place as the other validation, where we have access to all of the data. At the same time it feels nice to have that stuff in a separate plugin, so that each is only doing "one thing".

Is there anything we'd need for the `is_5ftf_contribution()` calculation that doesn't exist in the database? It doesn't seem like a big deal for query for a few things if they're already saved as activity meta data. There might be cases where we'll have to do that anyway, like getting the author of the forum post that is being replied to (to check if it's the same as the person making the reply).

I'm still open to exploring stuff like #75 or adding the logic to the `handle_...` functions too, so feel free to push back if you disagree.

iandunn: 😆 I might have confused myself a bit, I'm not sure if you're still wanting to explore the `type` approach or not.

StevenDufresne: I will **not** be exploring the `type` approach. I agree with your assessment.

StevenDufresne: > 🤔 The GitHub code runs on w.org, but the activity handler runs on buddypress/profiles; are you thinking this plugin >would run on both of those networks?
>
> I like that idea, since the same function could be used for DRYness.

Yeah, I'm thinking we could also handle trac in the same way. Only this file will, provided I don't run into any blockers, save to the meta field.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
StevenDufresne: Great, thanks for that context.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
StevenDufresne: Very helpful. I'm not sure I'll get this right the first time since I lack the context.

I think we agree on the approach, so I'll get this PR in a better state and ping you then. Thanks for the help so far 👍

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,113 @@
...
+Plugin Name: WordPress.org 5ftf Activity Handler
+Plugin URI: http://wordpress.org
+License: GPL2
+Version: 1.1
+Description: Handles saving of last 5ftf contribution.
+*/
+
+defined( 'ABSPATH' ) or die();
+
+if ( ! class_exists( 'WPOrg_5ftf_Activity_Handler' ) ) {
```
iandunn:
```suggestion
if ( class_exists( 'WPOrg_5ftf_Activity_Handler' ) ) {
	return;
}
```

This way the rest of the file doesn't have to be indented.

api.wordpress.org/public_html/dotorg/github/activity.php
```diff
@@ -71,6 +71,22 @@ function github_user_to_user_id( $user ) {
...
+ *
+ * @uses apply_filters() Calls 'wporg_github_added_activity'
+ */
+function save_github_activity( $data ) {
+	$wpdb->insert(
+		'wporg_github_activity',
+		$data
+	);
+
+	apply_filters( 'wporg_github_added_activity', $data );
```
iandunn:
```suggestion
	do_action( 'wporg_github_added_activity', $data );
```

I think an action fits better here, since the value isn't being changed.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/tests/test.php
```diff
@@ -0,0 +1,91 @@
...
+
+use WPOrg_5ftf_Activity_Handler;
+
+ini_set( 'display_errors', 'On' ); // won't do anything if fatal errors
+
+if ( 'staging' !== wp_get_environment_type() || 'cli' !== php_sapi_name() ) {
+	die( 'Error: Wrong environment.' );
+}
+
+const TEST_USERNAME = 'dufresnesteven';
```
iandunn: It'd be best to setup a test account for this purpose, so that your real account doesn't get test data mixed in with the real data.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,113 @@
...
+
+	class WPOrg_5ftf_Activity_Handler {
```
iandunn: What's the benefit of using a class in this case? It doesn't seem like it's using any of the distinctive features of OOP like inheritance, polymorphism, encapsulation, etc.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,113 @@
...
+
+			return self::update_last_contribution_meta( $args['user_id'] );
+		}
+
+		/**
+		 * Returns whether action is considered a contribution.
+		 *
+		 * @return boolean True if action == contribution
+		 */
+		public function is_5ftf_contribution( $action ) {
```
iandunn: It might be good to clarify the name of this function and/or `handle_github_activity()`, since the name here implies that it'd also be used for GitHub/Trac/etc.

Or maybe `handle_github_activity()` should call this function, to be consistent w/ `handle_wordpress_activity()`? Or the contents of this function moved into `handle_wordpress_activity()`?

I don't feel strongly which one, as long as the end result is consistent/clear.

iandunn (commented):
This is shaping up well 👍🏻 

Left a few suggestions and questions, but nothing blocking or anything serious, feel free to use your judgement/preferences.

api.wordpress.org/public_html/dotorg/github/activity.php
StevenDufresne: I understand the difference now :) Thanks!

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
StevenDufresne: Yep.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
StevenDufresne: Fair question. :)

I originally wanted to pass in wp dependencies so we could write tests that don't touch the database but stopped short. I'm not convinced I can do that in a way that doesn't equal a huge constructor with too many optional parameters filled with wp objects and wp functions. Additionally, reconstructing mock objects or finding drop-in replacements for this specific plugin is probably overkill.

I don't think an OOP approach introduces any future problems and I think it still leaves the door open to more abstracted code in the future.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
StevenDufresne: Yeah I agree, I feel like until I run into handling specific cases, which will definitely happen soon, I'm not sure what convention will work best.

 I'm in a bit of a spot with this PR where I'm thinking:

1. Get this code in production early and have it working (even if the value is not used), then follow up with the more specific cases and refactoring
2. Wait for more events or clearer definitions for teams, refactor, then deploy.

I'm leaning towards #1. Get something up and iterate as more data sources are introduced or team activities are better solidified. 

When I look at the team-specific tickets in the [non-code milestone](https://github.com/WordPress/five-for-the-future/milestone/8) none of them are complete yet or have all their discussions resolved. I don't think I want this PR hanging around that long unless we think we'll never get to use this meta value.

Thoughts?

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,125 @@
...
+		return in_array( $action, $valid_actions, true );
+	}
+
+	/**
+	 * Updates meta value to current timestamp indicating the user's last contribution.
+	 *
+	 * @return int|bool result of update_user_meta();
+	 */
+	protected function update_last_contribution_meta( $user_id ) {
+		return update_user_meta( $user_id, self::last_contribution_meta_key, time() );
```
StevenDufresne, line 130: @iandunn Should we save a separate meta field with information about activity that led to the update for debugging purposes and maybe dispute resolution?

`wporg_5ftf_last_contribution_meta` = `{ "activity": "blog_post_create", "source": "bp", ... }`

iandunn: That's a good idea 👍🏻

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
iandunn: Yeah, testing in WP is not ideal :/

I lean more towards YAGNI and lean/straight-forward code where OOP is only used when it solves a tangible problem, but this is a valid approach too 👍🏻

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
iandunn: Yeah, `#1` sounds like the better option to me too 👍🏻

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
```diff
@@ -0,0 +1,136 @@
...
+		return self::$instance;
+	}
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		if ( 'profiles.wordpress.org' === site_url() ) {
+			add_filter( 'bp_activity_add', array( $this, 'handle_activity' ) );
+		} elseif ( str_ends_with( 'wordpress.org', site_url() ) ) {
```
iandunn: Oh, before merging it'd be best to [add the `IS_WPORG_MULTINETWORK` bit](https://github.com/WordPress/wordpress.org/pull/78#discussion_r908806352), though, or at least a comment explaining that this matches all of the sites in the w.org multinetwork (*.w.org). It's kind of confusing because `profiles.wordpress.org` is actually on the _buddypress_ network, even though it has a w.org hostname.

WIthout some kind of explanation the "if .org else .org" is confusing at a quick glance.

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
StevenDufresne: Could we not look for the presence of the filter?

`has_filter( 'wporg_github_added_activity', 'example_alter_the_content' )`

StevenDufresne: Nevermind, I misunderstood how that works. WP Newb :) 🤕

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
StevenDufresne: Oh right, sorry I missed that. Added a comment back in the original thread.

https://github.com/WordPress/wordpress.org/pull/78#discussion_r908806352

profiles.wordpress.org/public_html/wp-content/plugins/wporg-5ftf-activity-handler/wporg-5ftf-activity-handler.php
StevenDufresne: https://github.com/WordPress/wordpress.org/pull/78/commits/ec154defc4fa59471832966ecf60b028e498b61c

Is that what you were thinking?

iandunn: Pretty close; I don't think w/ need the hostname check now that we have the constant, and it's a little safer to check the value of the constant:

```php
} elseif ( defined( 'IS_WPORG_MULTINETWORK' ) && IS_WPORG_MULTINETWORK ) {
```

StevenDufresne: Ha so close 👌.

iandunn:
Did this get merged, or did you close it for another reason?

StevenDufresne:
I see, I deleted my fork, I should still have this locally. I'll restore it.

PR #81, opened 2022-07-01
Turn on notifications for developer network.

StevenDufresne:
Related to https://github.com/WordPress/five-for-the-future/issues/178.

We want to turn on `Tracking code examples or user contributed notes, both for the submitter and the reviewer` on developers.wordpress.org. Currently, developer.wordpress.org posts are not considered [notifiable](https://github.com/WordPress/wordpress.org/blob/58668f17c7870eaa251e3c2ee1f2a7b0405dc55b/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php#L85).

This PR updates the developer theme to return `true` when the `wporg_profiles_wp_activity-is_post_notifiable` filter is applied.

### Considerations
Although our specific goal is to track comments on posts, specifically `wp-parser-hook`, `wp-parser-class`, `wp-parser-method`, `wp-parse-function` posts, this PR will turn on notifications for updates to all posts on developer.wordpress.org. As an alternative, we could add logic to make sure it's specific to the post types above and only for comments but I can't see a good reason why we shouldn't notify for all updates. The code already protects again [missing authors](https://github.com/WordPress/wordpress.org/blob/58668f17c7870eaa251e3c2ee1f2a7b0405dc55b/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php#L114) and [imports](https://github.com/WordPress/wordpress.org/blob/58668f17c7870eaa251e3c2ee1f2a7b0405dc55b/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php#L158), which I think should remove any potential issue since those posts types are unique (although I haven't tested this assumption). So other updates should also be considered activity & [5ftf](https://wordpress.org/five-for-the-future/) worthy.

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-developer/functions.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-developer/functions.php
@@ -168,6 +168,7 @@ function init() {
 	add_filter( 'breadcrumb_trail_items',  __NAMESPACE__ . '\\breadcrumb_trail_items_for_handbook_root', 10, 2 );
 
 	add_filter( 'syntaxhighlighter_htmlresult', __NAMESPACE__ . '\\syntaxhighlighter_htmlresult' );
+	add_filter( 'wporg_profiles_wp_activity-is_post_notifiable','__return_true' );
 }
 
 /**


wordpress.org/public_html/wp-content/themes/pub/wporg-developer/functions.php
```diff
@@ -168,6 +168,7 @@ function init() {
 	add_filter( 'breadcrumb_trail_items',  __NAMESPACE__ . '\\breadcrumb_trail_items_for_handbook_root', 10, 2 );
 
 	add_filter( 'syntaxhighlighter_htmlresult', __NAMESPACE__ . '\\syntaxhighlighter_htmlresult' );
+	add_filter( 'wporg_profiles_wp_activity-is_post_notifiable','__return_true' );
```
iandunn, line 171: 🤔 On the one hand, using this filter is good, because it decouples this "theme" from the notifier plugin. On the other hand, it's not consistent with how other post types are enabled. I personally lean towards keeping all the  logic in the notifier plugin, but I can see the case for this approach too. Up to you 👍🏻 

If we do stick with this approach, it looks like it'll need to be done more like the following, to avoid overriding a `false` value from the caller:

```php
function enable_notifications( $enabled ) {
    if ( ! $enabled ) {
         return false; // The notifier plugin has already determined it's not notifiable.
    }

    return true;
}
```

I would think this would also need to be affect `wporg_profiles_wp_activity-is_comment_notifiable`, but it worked in your tests?

iandunn (approved):
LGTM 👍🏻 

After merging, it might be worth backporting this to the experimental repo, or opening an issue there to backport, or add a checklist item to #178. That'd ensure it won't get overwritten if that repo replaces the current production codebase.

wordpress.org/public_html/wp-content/themes/pub/wporg-developer/functions.php
StevenDufresne: Yeah, I actually flipped back and forth on that. The existence of this filter makes it tricky to understand what the original author of the notifier had intended. My instinct suggests that the _growth_ of theme-specific logic in `is_post_notifiable` was not the desired architecture otherwise I don't understand what that filter would be used for. 

To me, since the filter is called last, it ultimately should determine whether it's `true` or `false` and the value that it's calculated to that point should be irrelevant. But that would also not be ideal because maybe there are some generic cases that it should not run for all cases. A theme may not have that context. 🤔 

Ideally, I think `is_post_notifiable` would make more sense like:
```
$notifiable = apply_filters( 'wporg_profiles_wp_activity-is_post_notifiable', $post )

...
generic cases that turn it off...
...

return $notifiable;
```
But anyway, no time to go back in history.

 My first attempt was actually to add all the `wp-parser-` types to the notifier, but that `is_post_notifiable` function was becoming pretty unwiedly. 

🤔 🤔 🤔 🤔 🤔 🤔 ⌛ 

You know what, I'll just assume that when we [added the learn post types](https://github.com/WordPress/wordpress.org/commit/3c872232db10fd5b1921ad33fb3a3ab7fcd4f574), we chose our path and go with that.

StevenDufresne:
>I would think this would also need to be affect wporg_profiles_wp_activity-is_comment_notifiable, but it worked in your tests?

It doesn't need any updates, it's only getting filtered on the post type right now. There doesn't appear to be any active filter calls that return false.

StevenDufresne:
Decided to just add the post types:

See https://meta.trac.wordpress.org/changeset/11934.

PR #83, opened 2022-07-07
Show patterns in theme directory.

StevenDufresne:
I've opened up this PR as a place for discussion on https://meta.trac.wordpress.org/ticket/5995. While this PR comes with code updates, it's purely experimental and not necessarily the best implementation or experience. I would have preferred to use an issue but they are closed for this repository and having the discussion in [trac](https://meta.trac.wordpress.org) feels a bit limiting.

# Assumption
It would be helpful for users to be able to preview patterns that are bundled with themes in the theme directory.

# Details
### Static code analysis
- Search for:
  - `register_block_pattern` ([JS](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-patterns/#register_block_pattern), [PHP](https://developer.wordpress.org/reference/functions/register_block_pattern/))
  - `theme.json` ([Link](https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-json/#patterns))
  - `/patterns` ([Link](https://github.com/WordPress/gutenberg/pull/36751))

### Dynamic analysis (with running environment)
 -  JS: `wp.data.select('core').getBlockPatterns()` ([Link](https://developer.wordpress.org/block-editor/reference-guides/data/data-core/#getblockpatterns))
 - PHP: `WP_Block_Patterns_Registry::get_all_registered()` ([Link](https://developer.wordpress.org/reference/classes/wp_block_patterns_registry/get_all_registered/)) 

 
## When can we detect patterns?
- Theme Upload
  - `class-wporg-themes-upload.php` (Static) ([Link](https://github.com/WordPress/wordpress.org/blob/trunk/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php))
   - Get Results from Theme Review Action (Dynamic) ([Link](https://github.com/WordPress/theme-review-action))
- Cron Job
  - I don't think any exist yet  
  
... store results in meta for `repopackage`...

- Dynamically on page load
  - This would only work for the `Theme Preview` since it has a running environment when the user interacts with it.

### Theme Directory
 - Directory View ([Link](https://wordpress.org/themes/))
 - Single Theme View ([Link](https://wordpress.org/themes/twentytwentytwo/))
 - Theme Preview ([Link](https://wordpress.org/themes/twentytwentytwo/) -> Click "Preview Button")

## Ideas
The theme directory is likely to undergo a redesign in the next year or so. If we want to get something up before that completes, which probably makes sense, we should aim to make the most minimal changes.

### Directory View
We could add Icons to the theme tile to denote that a theme bundles patterns.
| A  | B |
| --- | --- |
| Screen Shot 2022-07-08 at 8 24 15 AM | Screen Shot 2022-07-08 at 8 36 13 AM |

### Single Theme View 

We can add a section below the tags. It could be a list of the pattern names or could also use [pattern directory previewer](https://github.com/WordPress/pattern-directory/tree/trunk/public_html/wp-content/themes/pattern-directory/src/components/pattern-preview).  If so, It would have to have some sort pagination control experience since some themes bundle 10+ patterns.

Screen Shot 2022-07-08 at 8 46 25 AM 

### Theme Preview
What if we give a navigation item that users can click that loads a dynamically generated page with all the patterns? I've added blocks to the menu just as an example.

Screen Shot 2022-07-08 at 9 10 36 AM

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/theme-directory/class-themes-api.php
+++ b/wordpress.org/public_html/wp-content/plugins/theme-directory/class-themes-api.php
@@ -64,6 +64,7 @@ class Themes_API {
 		'requires'           => false,
 		'requires_php'       => false,
 		'trac_tickets'       => false,
+		'patterns'           => false,
 	);
 
 	/**
@@ -892,6 +893,33 @@ class Themes_API {
 				$phil->sections['description'] = GlotPress_Translate_Bridge::translate( $phil->sections['description'], $glotpress_project );
 			}
 
+		}
+
+		if ( $this->fields['patterns'] ) {
+			$pattern_meta = get_post_meta( $theme->ID, '_patterns', true );
+			$patterns_data = array();
+
+			if ( ! empty( $pattern_meta ) ) {
+				switch_to_blog( 699 ); // Pattern Directory
+
+				$args = array(
+					'post_name__in' => $pattern_meta,
+					'post_type'     => 'wporg-pattern',
+					'post_status'   => 'publish',
+				);
+
+				foreach ( get_posts( $args ) as $pattern_post ) {
+					$p = array();
+					$p['name'] = $pattern_post->post_title;
+					$p['link'] = get_permalink( $pattern_post );
+
+					$patterns_data[] = $p;
+				}
+
+				restore_current_blog();
+			}
+
+			$phil->patterns = $patterns_data;
 		}
 
 		wp_cache_set( $cache_key, $phil, $this->cache_group, $this->cache_life );
--- a/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
+++ b/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
@@ -76,6 +76,13 @@ class WPORG_Themes_Upload {
 	public $theme;
 
 	/**
+	 * The uploaded theme.json.
+	 *
+	 * @var WP_Theme_JSON
+	 */
+	public $theme_json = null;
+
+	/**
 	 * The theme slug being uploaded.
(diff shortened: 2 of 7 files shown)


adamwoodnz:
For the Single Theme view, even simpler than using a pattern preview could be having a carousel of patterns, with each pattern using something like a pattern grid item:
Screen Shot 2022-07-08 at 1 24 56 PM

I'm not sure if having all the functionality of the preview is necessary here (drag to resize, etc.), although I guess that could be disabled.

adamwoodnz:
I think all these ideas have merit. If we were to have a Patterns view from the Theme Preview page that could probably use a pattern grid too.

adamwoodnz:
On theme upload seems the most efficient, as long as that lifecycle runs every time a theme is changed.

mtias:
Note that themes can auto-register patterns by placing them in a `/patterns` folder with the appropriate headers: https://github.com/WordPress/gutenberg/pull/36751

StevenDufresne:
> For the Single Theme view, even simpler than using a pattern preview could be having a carousel of patterns, with each pattern using something like a pattern grid item: [...]
>
> I'm not sure if having all the functionality of the preview is necessary here (drag to resize, etc.), although I guess that could be disabled.

Just noting, we don't have screenshots of patterns yet, and the image you shared uses the [underlying control ](https://github.com/WordPress/pattern-directory/blob/f5ca69c1bebe286203f008b62f03d8233ab387e0/public_html/wp-content/themes/pattern-directory/src/components/pattern-thumbnail/index.js#L38)of the pattern preview control, without the controls.

>On theme upload seems the most efficient, as long as that lifecycle runs every time a theme is changed.

Yep, new uploads and updates to themes are centralized.

PR #84, opened 2022-07-10
Profiles: Add activity when added as a presenter for Learn workshop

outdoor2kode:
Issue: https://github.com/WordPress/five-for-the-future/issues/179 - Add activity when added as a presenter for Learn workshop

## Question
1. If an author published a post, then switched back to draft, and published again, there'll be two identical activities on assigned presenters' profiles, is it expected behavior?

2. The icon before the assigned presenter activity is also changed as the screencast below shows, but I can't locate the `style.css` of profiles in the repo. How do we usually modify this file, do we directly modify `themes/profiles.wordpress.org/style.css` on the sandbox?

## Screencast of results
https://user-images.githubusercontent.com/18050944/178162081-1670a4e6-6fa1-4b99-9276-173840ce3b74.mov

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
@@ -58,6 +58,7 @@ class WPOrg_WP_Activity_Notifier {
 	public function init() {
 		add_action( 'transition_post_status',    array( $this, 'maybe_notify_new_published_post'   ), 10, 3 );
 		add_action( 'post_updated',              array( $this, 'maybe_notify_updated_post'         ), 10, 3 );
+		add_action( 'wp_insert_post',         	 array( $this, 'insert_post' ),                       10, 2 );
 		add_action( 'transition_comment_status', array( $this, 'maybe_notify_new_approved_comment' ), 10, 3 );
 		add_action( 'wp_insert_comment',         array( $this, 'insert_comment' ),                    10, 2 );
 
@@ -164,6 +165,67 @@ class WPOrg_WP_Activity_Notifier {
 		$this->notify_blog_post( $post, 'new' );
 	}
 
+	public function insert_post ($postId, $post) {
+		if ( defined( 'REST_REQUEST' ) && REST_REQUEST ) {
+			return;
+		}
+		
+		if ( ! $this->is_post_notifiable( $post ) ) {
+			return;
+		}
+
+		if ( $post->post_type === 'wporg_workshop' ) {
+			$this->notify_workshop_presenter( $post );
+		}
+	}
+
+	/**
+	 * Sends activity notification for workshop presenter.
+	 *
+	 * @param WP_Post $post The published post
+	 */
+	public function notify_workshop_presenter( $post ) {
+		if ( defined( 'WP_IMPORTING' ) && WP_IMPORTING ) {
+			return;
+		}
+
+		$presenter_wporg_username = get_post_meta( $post->ID, 'presenter_wporg_username');
+
+		if ( empty( $presenter_wporg_username ) ) {
+			return;
+		}
+		
+		$permalink = get_permalink( $post );
+
+		$title = wp_kses_data( $post->post_title );
+		
+		$content = wp_trim_words(
+			strip_shortcodes( has_excerpt( $post ) ? $post->post_excerpt : $post->post_content ),
+			55
+		);
+
+		foreach ($presenter_wporg_username as $username) {
+			$userId = get_user_by( 'slug', strtolower( $username ) )->ID;
+
+			$args = array(
+				'action'       => 'wporg_handle_activity',
+				'component'    => 'learn',
+				'type'         => 'workshop_presenter_assign',
+				'user_id'      => $userId,
+				'primary_link' => $permalink,
+				'item_id'      => $post->ID,
+				'content'      => $content,
+				'message' => sprintf(
+					'Assigned presenter to the workshop, <i><a href="%s">%s</a></i>, on the site %s',
+					$permalink,
+					$title,
+					get_bloginfo( 'name' )
+				),
+			);
+	
+			Profiles\api( $args );
+		}
+	}
 
 	/**
 	 * Sends activity notification for new blog post.
--- a/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
@@ -223,10 +223,9 @@ class WPOrg_WP_Activity_Notifier {
 				'item_id'      => $post->ID,
 				'content'      => $content,
 				'message'      => sprintf(
-					'Assigned presenter to the tutorial, <i><a href="%s">%s</a></i>, on the site %s',
+					'Assigned as a presenter on the Learn WordPress tutorial, <i><a href="%s">%s</a></i>',
 					$permalink,
 					$title,
-					get_bloginfo( 'name' )
 				),
 			);
 
--- a/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
@@ -292,7 +292,6 @@ class WPOrg_WP_Activity_Notifier {
 	 * @param WP_Comment $comment Comment.
 	 */
 	public function insert_comment( $id, $comment ) {
-		if ( 1 == $comment->comment_approved ) {
 		if ( 1 === $comment->comment_approved ) {
 			$this->maybe_notify_new_approved_comment( 'approved', '', $comment );
 		}
--- a/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
@@ -223,7 +223,7 @@ class WPOrg_WP_Activity_Notifier {
 				'item_id'      => $post->ID,
 				'content'      => $content,
 				'message'      => sprintf(
-					'Assigned presenter to the workshop, <i><a href="%s">%s</a></i>, on the site %s',
+					'Assigned presenter to the tutorial, <i><a href="%s">%s</a></i>, on the site %s',
 					$permalink,
 					$title,
 					get_bloginfo( 'name' )
--- a/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
@@ -60,7 +60,6 @@ class WPOrg_WP_Activity_Notifier {
 	public function init() {
 		add_action( 'transition_post_status',    array( $this, 'maybe_notify_new_published_post'   ), 10, 3 );
 		add_action( 'post_updated',              array( $this, 'maybe_notify_updated_post'         ), 10, 3 );
-		add_action( 'wp_insert_post',         	 array( $this, 'insert_post' ),                       10, 2 );
 		add_action( 'transition_comment_status', array( $this, 'maybe_notify_new_approved_comment' ), 10, 3 );
 		add_action( 'wp_insert_comment',         array( $this, 'insert_comment' ),                    10, 2 );
 
@@ -100,8 +99,6 @@ class WPOrg_WP_Activity_Notifier {
 
 		// All actions can notify about handbooks.
 		$notifiable_post_types = array( 'handbook' );
-		// wp-parser-* post types belong to developer.wordpress.org.
-		$notifiable_post_types = array( 'post', 'handbook', 'wporg_workshop', 'lesson-plan', 'course', 'wp-parser-hook', 'wp-parser-function', 'wp-parser-method', 'wp-parser-class' );
 
 		// There's a large number of custom handbooks, and more will be created in the future.
 		if ( str_contains( $post->post_type, '-handbook' ) ) {
@@ -139,19 +136,6 @@ class WPOrg_WP_Activity_Notifier {
 			$notifiable = false;
 		} elseif ( $post->_xpost_original_permalink || str_starts_with( $post->post_name, 'xpost-' ) ) {
 			$notifiable = false;
-		if ( is_plugin_active( 'subscribers-only.php' ) ) {
-			$notifiable = false;
-		} elseif ( ! in_array( $post->post_type, $notifiable_post_types, true ) ) {
-			$notifiable = false;
-		} elseif ( 'publish' !== $post->post_status ) {
-			$notifiable = false;
-		} elseif ( ! empty( $post->post_password ) ) {
-			$notifiable = false;
-		} elseif ( $post->wporg_markdown_source || ! $post->post_author ) {
-			// Some Handbook posts are automatically created and don't have an author.
-			$notifiable = false;
-		} elseif ( $post->_xpost_original_permalink || str_starts_with( $post->post_name, 'xpost-' ) ) {
-			$notifiable = false;
 		} else {
 			$notifiable = true;
 		}
@@ -167,16 +151,36 @@ class WPOrg_WP_Activity_Notifier {
 	 * @param WP_Post $post The post.
 	 */
 	public function maybe_notify_new_published_post( $new_status, $old_status, $post ) {
-		if ( ! $this->is_post_notifiable( $post ) ) {
+		if ( 'publish' != $new_status ) {
 			return;
 		}
 
+		/**
+		 * Gutenberg sends two requests when we hit the Publish/Update button.
+		 * https://github.com/WordPress/wordpress.org/pull/84#discussion_r919290748
+		 *
+		 * For the first request, $old_status would be different from $new_status,
+		 * if the post, for example, is changed from draft to published.
+		 * For the second request (from the same Publish/Update button hit),
+		 * $old_status would be the same as $new_status in the same example,
+		 * both their values would be 'publish'.
+		 *
+		 * This brings the result that only the first request from Gutenberg could
+		 * pass the condition if ($old_status == $new_status) { return; }.
+		 *
+		 * However, only the second request would carry the data from meta boxes,
+		 * which is what we need here, so we need to put this logic above that
+		 * condition.
+		 */
 		if ( 'wporg_workshop' === $post->post_type ) {
 			$this->notify_workshop_presenter( $post );
 		}
 
+		if ( $old_status == $new_status ) {
+			return;
+		}
+
 		if ( ! $this->is_post_notifiable( $post, 'publish' ) ) {
-		if ( $old_status === $new_status ) {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
```diff
@@ -150,6 +151,71 @@ public function maybe_notify_new_published_post( $new_status, $old_status, $post
 		$this->notify_new_blog_post( $post );
 	}
 
+	public function insert_post ($postId, $post) {
+		if ( defined( 'REST_REQUEST' ) && REST_REQUEST ) {
```
outdoor2kode: Avoid the second firing action triggered by Gutenberg.
(For the case: Publish->Switch to Draft->Publish Again)

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
```diff
@@ -150,6 +151,71 @@ public function maybe_notify_new_published_post( $new_status, $old_status, $post
...
+		$title = wp_kses_data( $post->post_title );
+		
+		$content = wp_trim_words(
+			strip_shortcodes( has_excerpt( $post ) ? $post->post_excerpt : $post->post_content ),
+			55
+		);
+
+		foreach ($presenter_wporg_username as $username) {
+			
+			sleep(1);
```
outdoor2kode: It seems we got the rate limit, so sleep a second before the next call out.
(For the case that author and presenter are the same person)

tellyworth: I don't think `sleep()` is a good solution in production, though it's useful for diagnosing the problem. It could theoretically lead to unexpected UI delays when publishing a post. Better to find a more robust solution: either find out why the timeout is occurring and solve that, or schedule the function to run in a cron job.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
tellyworth: Is `REST_REQUEST` a reliable way to detect that sequence of events? And is that a common occurrence?

At a quick glance it seems some of the other notification types don't try to prevent duplicate activities when something is published twice. It might be overkill, but if it's common then as a general solution I would consider using postmeta to store a list of notified actions to prevent dupes. Perhaps that's better done on the profile API side - @dd32 do you have an opinion on this?

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
```diff
@@ -51,6 +51,7 @@ private function __construct() {
 
 	public function init() {
 		add_action( 'transition_post_status',    array( $this, 'maybe_notify_new_published_post'   ), 10, 3 );
+		add_action( 'wp_insert_post',         	 array( $this, 'insert_post' ),                       10, 2 );
```
tellyworth: I'm curious why you used the `wp_insert_post` action rather than `transition_post_status` (as used by `maybe_notify_new_published_post()`.

outdoor2kode: I used `transition_post_status` in the beginning, but since the action is triggered before saving data to DB, I can't get the post meta with the function [get_post_meta](https://developer.wordpress.org/reference/functions/get_post_meta/).
[The problem](https://github.com/WordPress/wordpress.org/pull/84#discussion_r918328703) would be gone if `transition_post_status` can be used here, are there any simple ways to get the post_meta before saving it to DB here?

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
outdoor2kode: Since `wp_insert_post` would be fired twice because of Gutenberg (according to these two posts: [post1](https://wordpress.stackexchange.com/questions/395666/why-is-save-post-hook-being-called-twice-despite-all-my-efforts), [post2](https://wordpress.org/support/topic/publish_post-hook-trigger-twice-when-i-publish-post/)) and causes the issue showed in the screencast below. (You can see that there is another updating... after published, making the account rentest have duplicate activities for the assigned presenter)

https://user-images.githubusercontent.com/18050944/178351308-647ccfc7-e6eb-460e-89c1-e7e7b6f9efa9.mp4

I first tried the suggested solution brought up in post1, which is to use [wp_after_insert_post](https://developer.wordpress.org/reference/functions/wp_after_insert_post/) instead, but the result didn't change.

So I then use the method mentioned in post2, which is using `REST_REQUEST` to detect the REST API fired by the Gutenberg editor.

It does smell though... not sure if there's a better way to handle this.

> using postmeta to store a list of notified actions to prevent dupes

I might need to ask you more about this.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
outdoor2kode: I should remove this. It was added because at first I was putting the logic in `transition_post_status`.
Now that the logic is under `wp_insert_post`, the problem of writing data to DB twice in a short time no longer exists.
(It would write the `Wrote a new workshop` activity first during `transition_post_status` and then after the post is published, the `Assigned presenter` is sent to DB. Previously, these two would be fired almost at the same time, causing weird behavior of notifying activity.)

I'm still curious what's the practice for us to handle such case (write DB in a row in a short time) since the logic might be changed back to use `transition_post_status`.
cron job sounds like a good idea.

> either find out why the timeout is occurring and solve that

What does timeout here refer to, timeout from sleep? or timeout of writing DB? Thanks!

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
dd32: > Since wp_insert_post would be fired twice because of Gutenberg

You should be able to look at the metadata of the insert I believe, and see the state change somehow (I'm guessing post_status: auto-draft -> draft -> published maybe?)

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
dd32: >> either find out why the timeout is occurring and solve that
> What does timeout here refer to

I believe Alex meant rate-limit here, not timeout.

There shouldn't be any rate-limits in production that would catch this - if there is, we should probably get those lifted for self-calls.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
dd32: > It might be overkill, but if it's common then as a general solution I would consider using postmeta to store a list of notified actions to prevent dupes.

Dupes should be able to be handled on the profiles side, every activity entry has a contextual link/id/etc.. but avoiding sending duplicate data in the same pageload is obviously beneficial from a performance standpoint..

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
tellyworth: Yep sorry Dion's right, I meant the rate limit.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
tellyworth: Hm, that's interesting about the extra call to wp_insert_post. I wonder if this is the cause:

https://wordpress.stackexchange.com/questions/395666/why-is-save-post-hook-being-called-twice-despite-all-my-efforts#comment576840_395672

> Some plugins incorrectly make a second REST API request to save meta after a post has been saved because the author doesn't understand the core data API and doesn't realise Gutenberg will send the meta for them if they register it and update the posts meta directly. This is where that comes from, and it's bad practice

I don't know how the presenter metadata is handled, but it's possible that (or other postmeta) is causing the duplicate call. That might explain why you were having trouble fetching the postmeta from the `transition_post_status` hook - maybe the meta is only stored in the second call, after the post status has changed. I'll do some testing and see what I can turn up.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
tellyworth: After thinking about it some more: you're on the right track, `wp_insert_post` is (almost) the right action to use here. If it was only on `transition_post_status` then we'd miss seeing presenters that were added to the post after it was published.

I think what's needed is actually the `wp_insert_post_data` filter. That's triggered during `wp_insert_post()`, but before the post is inserted or updated:

https://github.com/WordPress/wordpress-develop/blob/9b105d92a4b769f396ba798db1f106abab75001f/src/wp-includes/post.php#L4361-L4374

If you hook that one, the `$postarr` array should contain all the data being inserted, including the `meta_input` data that gets [inserted as postmeta](https://github.com/WordPress/wordpress-develop/blob/9b105d92a4b769f396ba798db1f106abab75001f/src/wp-includes/post.php#L4498). And the `$update` param should tell you if it's an update or an insert. So in the case of an insert, you'd add activity for all of those users. If it's an update, you could use `get_post_meta()` to figure out if any new presenters are being added, and only add activity for those.

Does that make sense? I think that would mostly deal with the issue of duplicates too, since you'd only notify when a new username is found in `$postarr`.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
outdoor2kode: I found some discussion here, I guess it explains why this happened.
https://github.com/WordPress/gutenberg/issues/12903#issue-391380038
> As we know that Gutenberg sends two requests when we hit the Publish/Update button:
a) Request to WP REST API to create/update the post
b) Request to post.php to save/submit the metaboxes

> a) Gutenberg saves data via one or two requests; the second save request only fires if there are meta boxes, so a plugin cannot depend on this request happening.
b) If there are going to be two requests, the plugin can't necessarily depend on the first request, because the first doesn't contain all the data for the post.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
outdoor2kode: `wp_insert_post_data`  -> This is nice!
It could deal with the case: 
[run1] presenter: ren -> send activity to ren
[run2] presenter: ren, ren2 -> send activity to ren2

But if it's the following case, it still duplicates, but I believe it's acceptable.
[run1] presenter: ren, ren2 -> send activity to ren, ren2
[run2] presenter: ren -> send activity to no one
[run3] presenter: ren, ren2 -> send activity to ren2 again

tellyworth: Yeah I agree - handling of that edge case isn't quite right, but it's good enough for now. 

If we also add some basic dupe handling on the API side then that should be pretty solid. Let's make that a separate issue. 

There's also the issue of removal, but since I don't think there's a delete API there's nothing we can do about it for now. We could consider handling that in a separate issue later if it becomes an issue.

outdoor2kode:
Changed back to use `transition_post_status`.
@dd32 @tellyworth Could you help review it again? thanks!

1. This [issue](https://github.com/WordPress/wordpress.org/pull/84#discussion_r918328703) and [The weird behavior caused by sending two activities](https://github.com/WordPress/wordpress.org/pull/84#discussion_r918361536) (`Wrote a new workshop` & `Assigned presenter`) together to the same user in a short time **WON'T** happen anymore because ```$this->notify_new_blog_post( $post );``` would be executed during the first request, and `$presenter_wporg_username` would have value only when Gutenberg makes the second request, so its following logic will be skipped during the first request due to ```if ( empty( $presenter_wporg_username ) ) {
			return;
		}```. 

2. For now, both a) Updating a post and b) Switching back to draft and publishing a post again will trigger notify_workshop_presenter and cause duplicate activities. After several tries, it seems to make more sense and efficient to implement the logic on the Profile API side. [Issue](https://github.com/WordPress/five-for-the-future/issues/204) opened.

3. What's the flow of modifying files that are not on Github repos?
> The icon before the assigned presenter activity is also changed as the screencast below shows, but I can't locate the style.css of profiles in this repo. How do we usually modify this kind of file, do we directly modify themes/profiles.wordpress.org/style.css on the sandbox and commit it?

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
```diff
@@ -186,24 +230,24 @@ public function notify_new_blog_post( $post ) {
...
 	 *
-	 * @param int        $id      Comment ID
-	 * @param WP_Comment $comment Comment
+	 * @param int        $id      Comment ID.
+	 * @param WP_Comment $comment Comment.
 	 */
-	function insert_comment( $id, $comment ) {
-		if ( 1 == $comment->comment_approved ) {
+	public function insert_comment( $id, $comment ) {
+		if ( 1 === $comment->comment_approved ) {
```
iandunn: Make sure you test changes like this if you haven't already, because WP APIs sometimes return ints as strings, etc.

I'll often only run `phpcbf` on old code, because it will make the formatting consistent without making any changes that are potentially breaking (and therefore require testing). 

Similarly, renaming functions needs testing too, in case some of the callers were missed, or maybe there are child classes that need updating too, etc.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
```diff
@@ -130,30 +120,84 @@ public function is_post_notifiable( $post ) {
...
+				'content'      => $content,
+				'message'      => sprintf(
+					'Assigned presenter to the tutorial, <i><a href="%s">%s</a></i>, on the site %s',
+					$permalink,
+					$title,
+					get_bloginfo( 'name' )
+				),
+			);
+
+			Profiles\api( $args );
```
iandunn, line 236: Rather than making multiple HTTP requests, you can send a single batch request, with an array of activities.  Here's an example:

* Notifier:
https://github.com/WordPress/wordpress.org/blob/0362b60552015b5433c4abffabd61a2dd760f789/wordpress.org/public_html/wp-content/plugins/wporg-gp-profiles/profiles.php#L65-L100
* Handler: https://github.com/WordPress/wordpress.org/blob/0362b60552015b5433c4abffabd61a2dd760f789/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php#L886-L904.
 
You'll also need to update [the `user_id` check](https://github.com/WordPress/wordpress.org/blob/0362b60552015b5433c4abffabd61a2dd760f789/profiles.wordpress.org/public_html/wp-content/plugins/wporg-profiles-activity-handler/wporg-profiles-activity-handler.php#L185-L192) if you use that method. It may make sense to just turn that into a generic check, like:

```php
if ( empty( $_POST['user'] ) && empty( $_POST['user_id'] ) && empty( $_POST['activities'][0]['user_id'] ) ) {
    throw new Exception( '-1 No user specified.' );
}
```

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
```diff
@@ -130,30 +120,84 @@ public function is_post_notifiable( $post ) {
...
+				'type'         => 'workshop_presenter_assign',
+				'user_id'      => $user_id,
+				'primary_link' => $permalink,
+				'item_id'      => $post->ID,
+				'content'      => $content,
+				'message'      => sprintf(
+					'Assigned presenter to the tutorial, <i><a href="%s">%s</a></i>, on the site %s',
+					$permalink,
+					$title,
+					get_bloginfo( 'name' )
```
iandunn:
```suggestion
					'Assigned as a presenter on the Learn WordPress tutorial, <i><a href="%s">%s</a></i>',
					$permalink,
					$title
```

This feels like it flows a bit smoother to me, but I don't have a strong opinion.

iandunn (commented):
I haven't tested, but the code LGTM 👍🏻 

I left a few suggestions, but they're not blockers.

It might also help to add a case to `tests/e2e.php`. I've found that easier for testing during development than doing it manually each time. It also avoids disrupting production data with testing.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
outdoor2kode: > Make sure you test changes like this if you haven't already, because WP APIs sometimes return ints as strings, etc.

Wasn't aware of this fact, thanks for the reminder, I've reverted it.

> I'll often only run phpcbf on old code, because it will make the formatting consistent without making any changes that are potentially breaking (and therefore require testing).

Good point, thanks! 

> Similarly, renaming functions needs testing too, in case some of the callers were missed, or maybe there are child classes that need updating too, etc.

👍 

The functions I've renamed in this PR are `_notify_forum_topic_payload` and `_notify_forum_reply_payload`. I've grep them in sandbox and they are only used in the file I renamed them. It seems safe to me so I keep the change.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
outdoor2kode: 👍 Yours is better. Changed.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
outdoor2kode: Thanks for sharing this!
I've taken a look at the examples, and I thought I might be able to also use similar logic to remove duplicates, so I'll tweak the logic when working on the [issue](https://github.com/WordPress/five-for-the-future/issues/204) together with adding a case to `tests/e2e.php`.

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
iandunn: 👍🏻

wordpress.org/public_html/wp-content/plugins/wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
iandunn: 👍🏻

iandunn:
merged in https://meta.trac.wordpress.org/changeset/12062/

PR #87, opened 2022-07-14
Bust cache for new full site editing filter

apeatling:
Bust cache for new full site editing filter

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-themes/functions.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-themes/functions.php
@@ -139,12 +139,12 @@ function wporg_themes_scripts() {
 		$GLOBALS['concatenate_scripts'] = true;
 	}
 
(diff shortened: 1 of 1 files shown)


adamwoodnz:
Sorry but I'm confused by this. The [changes to add the new filter](https://github.com/WordPress/wordpress.org/pull/85) appear to be all php, how does busting the JS and CSS cache help with this?

apeatling:
The final changeset is here, you can see new JS and CSS build files: https://meta.trac.wordpress.org/changeset/11963

apeatling:
I'm not super familiar with how this all works, but looking at past commits, this is what's needed.

adamwoodnz:
> The final changeset is here, you can see new JS and CSS build files: https://meta.trac.wordpress.org/changeset/11963

Ah makes more sense now, thanks

adamwoodnz:
Fixed in https://meta.trac.wordpress.org/changeset/11965

Thanks!

PR #89, opened 2022-07-22
Update trac header and footer for updated navigation link colors

adamwoodnz:
See https://github.com/WordPress/wporg-mu-plugins/issues/218

The link colors for the global navigation on wordpress.org sites were updated in https://github.com/WordPress/wporg-mu-plugins/pull/214

This PR updates the trac headers to take these changes.

I ran https://github.com/WordPress/wordpress.org/blob/trunk/trac.wordpress.org/templates/update-headers.php [to make the updates as suggested](https://github.com/WordPress/wporg-mu-plugins/issues/218#issuecomment-1188080194).


dd32:
@adamwoodnz For sync things like this, you can just run the script and commit the diffs without needing to get a review, as long as you're not applying any customisations to it :)

adamwoodnz:
> @adamwoodnz For sync things like this, you can just run the script and commit the diffs without needing to get a review, as long as you're not applying any customisations to it :)

Thanks, I struggled to find the update-headers.php script in my sandbox but I'll take another look

dd32:
>  I struggled to find the update-headers.php script in my sandbox

Ah, it doesn't exist, as not all files are checked out to sandboxes - you have to do a `svn co https://meta.svn.wordpress.org/sites/trunk/trac.wordpress.org/templates/` and then run/commit it.

Because of that, I've just added `bin/sync/trac-templates.sh` just so we have all scripts in one place :)

PR #93, opened 2022-08-19
Fix Openverse footer from being cut off

ramadanomar:
This should prevent the website from cutting off . See issue #1583 https://github.com/WordPress/openverse-frontend/issues/1583

[Ticket](https://meta.trac.wordpress.org/ticket/6450#ticket)

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/css/openverse.css
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/css/openverse.css
@@ -18,6 +18,7 @@ body {
 .site,
 .site-content {
     width: 100% !important;
+    height: calc(100vh - var(--wp-global-header-height)) !important;
     padding: 0 !important;
     max-width: 100% !important;
 }
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/css/openverse.css
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/css/openverse.css
@@ -2,7 +2,7 @@
     display: block;
 
     width: 100%;
-    height: calc(100vh - var(--wp-global-header-height) - var(--wp-admin--admin-bar--height))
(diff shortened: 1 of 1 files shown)


zackkrida (changes requested):
```
#openverse_embed {
    height: calc(100vh - var(--wp-global-header-height) - var(--wp-admin--admin-bar--height))
}

```

@ramadanomar I believe this is the correct fix. You can observe it by viewing https://wordpress.org/openverse, inspecting the page in devtools, and replacing the iframe height with this value.

ramadanomar:
Thanks for the pointers! Didn't know i could use variables directly in browser dev tools.

StevenDufresne:
Does this PR need to be deployed?

ramadanomar:
Yes. The footer appears to be cut off

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/css/openverse.css
```diff
@@ -2,7 +2,8 @@
     display: block;
 
     width: 100%;
-    height: calc(100vh - var(--wp-global-header-height));
+    height: calc(100vh - var(--wp-global-header-height) - var(--wp-admin--admin-bar--height))
```
ocean90:
```suggestion
    height: calc(100vh - var(--wp-global-header-height) - var(--wp-admin--admin-bar--height, 0px))
```

Might be good to add a fallback there in case the admin bar gets disabled again.

ramadanomar:
That's a great suggestion, thanks @ocean90 !

StevenDufresne:
I can deploy this change, let me know!

zackkrida (approved):

zackkrida:
Lets get this deployed; thanks @StevenDufresne

StevenDufresne:
Deployed in https://meta.trac.wordpress.org/changeset/12040.

PR #94, opened 2022-08-30
Remove the Logo and Search nodes from the admin bar

dd32:
See https://github.com/WordPress/wporg-mu-plugins/issues/252

This PR removes the W logo and Search items from the admin bar.

Logged out view:
| Before | After |
| --- | --- |
| Screen Shot 2022-08-30 at 5 05 44 pm | Screen Shot 2022-08-30 at 5 05 56 pm |

Logged in view:
| Before | After |
| --- | --- |
| Screen Shot 2022-08-30 at 5 07 35 pm | Screen Shot 2022-08-30 at 5 07 20 pm |

This doesn't go as far as suggested in https://github.com/WordPress/wporg-mu-plugins/issues/252 but is a stepping stone towards it. There's more implementation required in order to remove the admin bar entirely, and ensure that the toolbar items that are required by those with elevated WordPress.org access can still access the tools/features that we have present in the admin bar.

### DIFF
--- a/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-admin-bar.php
+++ b/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-admin-bar.php
@@ -1,29 +1,20 @@
 <?php
+namespace WordPressdotorg\MU_Plugins\Admin_Bar;
 /**
  * Customizations to the Admin Bar across all networks.
  */
 
 /**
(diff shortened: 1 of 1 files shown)


dd32:
Also, we can probably shift this MU plugin over to https://github.com/WordPress/wporg-mu-plugins but I've created the PR here since we'd need to remove this file / alter this file anyway.

dd32:
Closing in favour of https://github.com/WordPress/wporg-mu-plugins/pull/256

PR #95, opened 2022-10-05
Update "Update PHP" support page

ironprogrammer:
Brings parts of the "Update PHP" support page up to date. See Trac ticket for additional details.

Meta Trac ticket: https://meta.trac.wordpress.org/ticket/6516

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-support/helphub-update-php-strings.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-support/helphub-update-php-strings.php
@@ -86,6 +86,18 @@ function wporg_support_filter_update_php_content( $content ) {
 	$content .= '</ul>';
 	$content .= '<p>' . __( 'These benefits are good for you, and good for your website&#8217;s visitors. These are the reasons you should update PHP today. The next section will show you how to do this.', 'wporg-forums' ) . '</p>';
 
+	// Section "How to check your PHP version".
+	$content .= '<h3>' . __( 'How to check your PHP version', 'wporg-forums' ) . '</h3>';
+	$content .= '<p>' . __( 'To check what version of PHP your WordPress site is using, from the WordPress Dashboard, select <em>Tools &gt; Site Health</em> from the sidebar menu, and then select the Info tab. Expand the Server section and scroll down until you see <strong>PHP version</strong>.', 'wporg-forums' ) . '</p>';
+	$content .= '<p>';
+	$content .= sprintf(
+		/* translators: %s: recommended PHP version */
+		__( 'If this number is at or higher than PHP %s, then you don&#8217;t need to update PHP at this time!' ),
+		RECOMMENDED_PHP
+	);
+	$content .= '</p>';
+	$content .= '<p>' . __( 'However, if your site is lower than the recommended version, running on an outdated and insecure version of PHP, then it&#8217;s time to update. You also may have noticed a warning such as "PHP Update Required" or "PHP Update Recommended" on the Dashboard or in Site Health – these serve as important reminders to update PHP and make your site more secure.', 'wporg-forums' ) . '</p>';
+
 	// Section "Before you update your PHP version".
 	$content .= '<h3>' . __( 'Before you update your PHP version', 'wporg-forums' ) . '</h3>';
 	$content .= '<p>' . __( 'This section starts off with some warnings, but don&#8217;t be afraid! As with most things technical, we just need to cover some background before we can get to the part where you update your PHP version.', 'wporg-forums' ) . '</p>';
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-support/helphub-update-php-strings.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-support/helphub-update-php-strings.php
@@ -56,7 +56,7 @@ function wporg_support_filter_update_php_content( $content ) {
 	$content .= '<p>';
 	$content .= sprintf(
 		/* translators: %s: link URL about keeping WordPress up to date */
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/themes/pub/wporg-support/helphub-update-php-strings.php
```diff
@@ -63,13 +63,20 @@ function wporg_support_filter_update_php_content( $content ) {
 	$content .= '<p>' . __( 'There are two main benefits to keeping PHP up-to-date:', 'wporg-forums' ) . '</p>';
 	$content .= '<ul>';
 	$content .= '<li>';
+	$content .= sprintf(
+		/* translators: 1: link URL to WordPress requirements page, 2: recommended PHP version */
+		__( '<strong>Your website will be faster</strong> because the latest version of PHP is more efficient. Updating to <a href="%1$s">WordPress&#8217;s recommended PHP version</a> (currently %2$s or higher) can deliver a huge performance increase; up to 3 or 4x faster than older versions.', 'wporg-forums' ),
```
azaozz: "Your website may be faster". "will be" is not guaranteed :) (may be slower too because a plugin doesn't work well with the never PHP version, slows everything down a lot and fills the logs with errors).

Also, "can deliver a huge performance increase; up to 3 or 4x faster than older versions". This is true for PHP itself, not for the website as a whole. PHP is actually pretty fast compared with network speeds, etc. so the performance increase is usually not noticeable. The users will be looking at website performance, not timing PHP :)

So thinking here it would be better to be a bit more modest. Perhaps something like "can deliver a performance increase that will benefit all visitors to your website."

azaozz:
Few more suggested changes:

- https://github.com/WordPress/wordpress.org/pull/95/files#diff-011f219680c25dbead7804217ea9b23de13e25c333d64c428b33fcfaa6c6f273R59 the "PHP is the coding language WordPress is built on" is not true. PHP is **one of the coding languages**.
- Same line, "its version is set at the server-level by your hosting company". Confusing? How can a version be "set". Better to say something like "it is installed on your server by the hosting company. Often there are several installed versions and you can choose which one is used for your website.".

wordpress.org/public_html/wp-content/themes/pub/wporg-support/helphub-update-php-strings.php
ironprogrammer: Thanks, @azaozz! These claims have been dampened down a bit in 441652d.

ironprogrammer:
In response to @azaozz:

> * https://github.com/WordPress/wordpress.org/pull/95/files#diff-011f219680c25dbead7804217ea9b23de13e25c333d64c428b33fcfaa6c6f273R59 the "PHP is the coding language WordPress is built on" is not true. PHP is **one of the coding languages**.
> * Same line, "its version is set at the server-level by your hosting company". Confusing? How can a version be "set". Better to say something like "it is installed on your server by the hosting company. Often there are several installed versions and you can choose which one is used for your website.".

Adjustments applied in ff6657b10 and 1397299c2.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/helphub-update-php-strings.php
```diff
@@ -56,20 +56,27 @@ function wporg_support_filter_update_php_content( $content ) {
 	$content .= '<p>';
 	$content .= sprintf(
 		/* translators: %s: link URL about keeping WordPress up to date */
-		__( 'PHP is the coding language WordPress is built on, and its version is set at the server-level by your hosting company. Whilst you may be familiar with the importance of <a href="%s">keeping WordPress, and your themes and plugins up-to-date</a>, keeping PHP up-to-date is just as important.', 'wporg-forums' ),
+		__( 'PHP is one of the programming languages that WordPress is built on. The PHP version used on your website is usually set by your hosting company, but may be an option you can adjust yourself. And while you may be familiar with the importance of <a href="%s">keeping WordPress, and your themes and plugins up-to-date</a>, keeping PHP up-to-date is just as important.', 'wporg-forums' ),
```
azaozz: > The PHP version used on your website is usually set by

Still a bit confusing here, perhaps. PHP is an application installed on the web server. How can the version of an application be set to something? I mean, usually cannot say things like "The version of Firefox on my computer was set to 105", right? :) 

Thinking that explaining availability of different PHP versions on the server would be easier if it uses familiar terms. For example: 

- PHP is (an application) installed on the server by the hosting company.
- There may be several versions installed.
- The site owner can choose which of these different PHP versions is used for their website.

Then perhaps further down can mention that if only one PHP version is installed on the server, the site owner may be able to request their site to be moved to another server with newer PHP version. Think that was a practice at some hosting companies, at least in the past, but good to confirm.

azaozz:
Yes, reads a lot better now imho. Good job!

Just one more comment about a version of an application being "set". Not sure this is clear enough, it is not an option to set the version of an app. The app has to be upgraded.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/helphub-update-php-strings.php
ironprogrammer: This has been reworded, removing the "set" terminology, and referring to the possibility that there may be multiple PHP versions available to the user (e6e229803761f8025efb1912e7d2edc044e1c72d).

ironprogrammer:
Thanks, @azaozz!

> Just one more comment about a version of an application being "set". Not sure this is clear enough, it is not an option to set the version of an app. The app has to be upgraded.

Also updated the "How to update..." section to remove reference to "set". Clarified that PHP is installed on the server (7a0d829544282d7439378eb5dc4ee2ed0b5e9ab6).

dd32:
@ironprogrammer Is this ready to go now?

ironprogrammer:
@dd32, yes, these updates are ready.

PR #98, opened 2022-10-15
Add redirection to Openverse standalone site

### LINKED ISSUE #308: Add toggleable 'redirection' mode that 301s the entire WordPress Theme
## Problem

As part of our iframe removal efforts, the Openverse WordPress theme located at https://github.com/WordPress/wordpress.org/tree/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse needs to be updated to redirect the entire site to our new domain, openverse.wordpress.net.

## Description

We need to add a way for the theme to 301 redirect entirely to a new URL. This is expected to be openverse.wordpress.net. It might be nice to put this redirect mode behind a customizer setting, as we did for the iFrame url already in the theme.

The redirects need to account for internationalization and query params need to be preserved. Some examples:

1. `https://ru.wordpress.org/openverse` should redirect to `openverse.wordpress.net/ru`
2. `https://wordpress.org/openverse/search/?q=dog` should redirect to `https://openverse.wordpress.net/search/?q=dog

## Alternatives

There is no alternative.

## Implementation

- [ ] 🙋 I would be interested in implementing this feature.

dhruvkb:
Fixes WordPress/openverse#308
Closes https://meta.trac.wordpress.org/ticket/6578

This PR updates the WordPress.org Openverse theme to add logic that redirects requests to https://wordpress.org/openverse to the new Openverse standalone site at https://openverse.wordpress.net (customisable). The redirection behaviour remains dormant unless enabled via a toggle.

It also updates the `README.md` file to include workarounds to some problems I faced during the setup.

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
@@ -8,8 +8,9 @@ Follow these steps to set up a local playground for the theme:
 
 0.  Install all the prerequisites.
 
-    1.  **Required:** Node.js 14.
-    2.  **Required:** Composer.
+    1.  **Required:** Node.js 14
+    2.  **Required:** Composer
+    3.  **Required:** Subversion
     3.  **Recommended:** Docker (to use the automatic setup)
 
 1.  Build the parent theme WordPress.org theme.
@@ -18,6 +19,12 @@ Follow these steps to set up a local playground for the theme:
         (i.e. `wordpress.org/wordpress.org/public_html/wp-content/themes/pub/wporg`).
     2.  Install all the required `npm` packages.
         ```bash
+        $ npm install
+        ```
+        If you face issues installing Sass, try the following command, and
+        `npm install` again.
+        ```bash
+        $ npm install node-sass@npm:sass
         $ npm install
         ```
     3.  Build the theme assets.
@@ -76,6 +83,12 @@ Follow these steps to set up a local playground for the theme:
         your WordPress install. This site will have the `wporg` (parent) and
         `wporg-openverse` (child) themes installed. For detailed instructions,
         please read [the wp-env docs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-env/).
+    4.  Edit the `.htaccess` file to prevent Apache 404 errors.
+        ```bash
+        $ wp-env run cli bash
+        bash-5.1$ printf "RewriteEngine on\nFallbackResource /index.php\n" > .htaccess
+        bash-5.1$ exit
+        ```
 
     **Manual:**  
     If you prefer a manual approach, you can also set up your own WordPress
@@ -96,7 +109,8 @@ Follow these steps to set up a local playground for the theme:
 
 6.  Activate and customize the theme.
 
-    1.  Log into `/wp-admin`.
+    1.  Log into `/wp-admin`. If you used `@wordpress/env`, your username will
+        be 'admin' and password will be 'password'.
     2.  Under Appearance > Themes, activate the theme 'WordPress.org Openverse'.
     3.  To change the embed URL, open the customizer at Appearance > Customize
         and update the value in the 'Openverse embed' panel.
@@ -106,6 +120,17 @@ Follow these steps to set up a local playground for the theme:
     1.  Change the Openverse embed to
         `/wp-content/themes/wporg-openverse/js/message_test.html`.
     2.  Visit the site and interactively test the messages.
+
+8. Test redirects.
+
+    1.  Visit some test URLs.
+        1.  https://ru.wordpress.org/openverse → ...
+        2.  https://wordpress.org/openverse/search/?q=dog → ...
+    2.  See the target redirect URL as a comment inside the dev tools.
+        1.  ... → https://openverse.wordpress.net/ru/
+        2.  ... → https://openverse.wordpress.net/search/?q=dog
+    3.  Change the language in Settings > General to see how the locale factors
+        into the redirect path.
 
 ## Related links
 
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
@@ -11,7 +11,7 @@ Follow these steps to set up a local playground for the theme:
     1.  **Required:** Node.js 14
     2.  **Required:** Composer
     3.  **Required:** Subversion
-    3.  **Recommended:** Docker (to use the automatic setup)
+    4.  **Recommended:** Docker (to use the automatic setup)
 
 1.  Build the parent theme WordPress.org theme.
 
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
@@ -253,11 +253,11 @@ if ( !defined( 'OPENVERSE_STANDALONE_URL' ) ) {
  * URL, the requested path and the current locale.
  *
  * Examples:
- * - https://ru.wordpress.org/openverse → {OPENVERSE_STANDALONE_URL}/ru/
- * - https://wordpress.org/openverse/search/?q=dog → {OPENVERSE_STANDALONE_URL}/search/?q=dog
+ * - https://ru.wordpress.org/openverse → {ov_redirect_url}/ru/
+ * - https://wordpress.org/openverse/search/?q=dog → {ov_redirect_url}/search/?q=dog
  */
 function get_target_url() {
-	$target_url = OPENVERSE_STANDALONE_URL;
+	$target_url = get_theme_mod( 'ov_redirect_url', OPENVERSE_STANDALONE_URL );
 
 	$curr_locale = get_locale();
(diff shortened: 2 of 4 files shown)


zackkrida:
Thanks, @dhruvkb. I'll review this today.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -203,3 +230,104 @@ function wporg_ov_customizer( $wp_customize ) {
...
+	Openverse standalone site redirection
+	=====================================
+ */
+
+/**
+ * This is the URL at which the standalone Openverse site is hosted. When
+ * redirect is enabled (see setting `ov_is_redirect_enabled`), the theme
+ * redirects all incoming requests to the right URL on this domain.
+ *
+ * Note: Do not put a trailing slash '/' in this URL, as it will cause problems.
```
sarayourfriend: Can you specify what the specific problems are? That way if in the future those problems are no longer relevant this comment can be deleted. The way it is right now it will be kind of scary for anyone trying to understand what the issue is, especially if they try it out themselves and are unable to reproduce any issues :open_mouth:

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -203,3 +230,104 @@ function wporg_ov_customizer( $wp_customize ) {
...
+ */
+if ( !defined( 'OPENVERSE_STANDALONE_URL' ) ) {
+	define( 'OPENVERSE_STANDALONE_URL', 'https://openverse.wordpress.net' );
+}
```
sarayourfriend, line 250: I wonder if this variable should be a theme variable instead? That way if we ever moved the location (say to `openverse.org`) we wouldn't need to redeploy the theme, just update the variable.

Edit: I see you actually already added it as a theme variable, but I don't see how it is used. Is there some magic going on that wires up the constant to the theme variable when the theme variable is defined?

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -203,3 +230,104 @@ function wporg_ov_customizer( $wp_customize ) {
...
+	$wp_customize->add_setting( 'ov_redirect_url', array(
+		'type' => 'theme_mod',
+		'capability' => 'edit_theme_options',
+		'default' => OPENVERSE_STANDALONE_URL,
+		'sanitize_callback' => function( $val, $setting ) {
+			if ( empty( $val ) ) {
+				return $setting->default;
+			}
+			return $val;
+		}
```
sarayourfriend: Might as well check for the trailing slash here as well to prevent mistakes/typos if we're using the theme variable.

sarayourfriend (commented):
Couple of questions about the theme variable and one chance to make things a little harder to mess up with a typo in the customizer. It looks good to me otherwise though :slightly_smiling_face: I haven't been able to test it locally yet.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -203,3 +230,104 @@ function wporg_ov_customizer( $wp_customize ) {
...
+		'priority' => 10,
+		'capability' => 'edit_theme_options',
+		'title' => 'Openverse Redirect',
+		'description' => 'Configure the redirection to the standalone Openverse site.'
+	) );
+
+	$wp_customize->add_setting( 'ov_redirect_url', array(
+		'type' => 'theme_mod',
+		'capability' => 'edit_theme_options',
+		'default' => OPENVERSE_STANDALONE_URL,
```
dhruvkb, line 296: Here we set the default value of the `ov_redirect_url` to the `OPENVERSE_STANDALONE_URL` constant.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
dhruvkb: There's no magic, it's explicitly set as the default value [here](https://github.com/WordPress/wordpress.org/pull/98#pullrequestreview-1146887519).

This default is only used if the value has not been modified or set to blank.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
dhruvkb: I don't remember what it was exactly, but iirc it was that paths already start with `/` so having a trailing slash here would cause them to appear twice.

```
https://openverse.wordpress.net/ + /path
```

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
sarayourfriend: Right... but where is the theme variable used? I only see the constant being used in the code, not the theme variable. That's where my confusion is coming from.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
sarayourfriend: Makes sense. Having a comment to that effect would be helpful :+1:

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
```diff
@@ -96,7 +109,8 @@ Follow these steps to set up a local playground for the theme:
 
 6.  Activate and customize the theme.
 
-    1.  Log into `/wp-admin`.
+    1.  Log into `/wp-admin`. If you used `@wordpress/env`, your username will
```
obulat, line 112: Thank you for this! :)

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/README.md
```diff
@@ -20,6 +21,12 @@ Follow these steps to set up a local playground for the theme:
         ```bash
         $ npm install
         ```
+        If you face issues installing Sass, try the following command, and
+        `npm install` again.
+        ```bash
+        $ npm install node-sass@npm:sass
```
obulat, line 27: Does this command replace `node-sass` with `sass` when running `npm install`? I was manually changing the `package.json` to get it to build before.

dhruvkb: Yeah, I don't know the specifics about that package but `node-sass` has problems running on M1 with Node.js 14. while `sass` works. Since this is only needed in dev, best to just use `sass` to get the base theme to build.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
dhruvkb: Good catch @sarayourfriend, thank you! Now fixed that in 0083dd5.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
dhruvkb: Added the comment in e390c58.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
dhruvkb: Added the check in e390c58.

obulat (approved):
I've tested the setup locally with:
- homepage
- search pages
- search pages with filter query params
- content pages
- `ru`/`ar` locales.
I always get the correct redirect comment at the top of the HTML in the devtools inspector.

The WP looks broken in the theme, but it's probably not related to this PR.

Screen Shot 2022-10-21 at 7 29 34 AM

zackkrida (approved):
Seeing the same as @obulat, LGTM. I hope we're not doing something wrong with getting the parent wp.org theme styles or something; but perhaps a contributor with more WordPress familiarity can take a look at this once we have the accompanying trac ticket.

dhruvkb:
@zackkrida this child theme inherits the header from the parent theme so if there is indeed a problem, it's very likely coming from the parent theme missing some plugins or stylesheets, which would not be the case in production 🤞.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -81,6 +84,21 @@ function disable_parse_request( $return, $wp ) {
...
+ * prefixed in the URL paths. It also returns a blank string if `$wp_locale` is
+ * not found in the locales list.
+ */
+function get_locale_slug ( $curr_locale ) {
+	if ( $curr_locale === 'en_US' ) {
+		return '';
+	}
+
+	$locale = get_locales()[$curr_locale];
+	return isset( $locale ) ? $locale->slug : '';
```
dd32, line 99: This is problematic as it'll cause a PHP Warning for missing `$curr_locale` key which shouldn't happen, but may.

This however can be written much easier as:
```suggestion
	return get_locales()[$curr_locale]->slug ?? '';
```

(No need to change anything here, I'm going to change it while I commit it)

dhruvkb: Thanks for the fix!

PR #99, opened 2022-10-18
Add a new plugin to filter some tags at make.wordpress.org/polyglots

amieiro:
This new plugin filters the posts by their tags at https://make.wordpress.org/polyglots, excluding the posts with the 'editor-requests', 'request', 'clpte', 'locale-requests' tags in the https://make.wordpress.org/polyglots/?exclude=requests URL.

To test this new plugin, you have to install it and activate it, and you have to access to 2 different URL:
- https://make.wordpress.org/polyglots where you will see all posts.
- https://make.wordpress.org/polyglots/?exclude=requests where you will see the main posts, not under the 'editor-requests', 'request', 'clpte', 'locale-requests' tags.

Once we release the plugin, we have to add a link to https://make.wordpress.org/polyglots/?exclude=requests in the right sidebar.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-polyglots-posts/wporg-polyglots-posts.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-polyglots-posts/wporg-polyglots-posts.php
@@ -1,0 +1,46 @@
+<?php
+namespace WordPressdotorg\Plugin\Polyglots_Posts;
+
+/**
+ * Plugin Name: WordPress.org Polyglots Posts.
+ * Description: Filters the posts by their tags, excluding the posts with the
+ * 'editor-requests', 'request', 'clpte', 'locale-requests' tags in the
+ * https://make.wordpress.org/polyglots/?exclude=requests URL.
+ * License: GPLv2 or later
+ */
+
+/**
+ * The ids of the tags to be excluded.
+ * 'editor-requests', 'request', 'clpte', 'locale-requests' id tags.
+ *
+ * @var array
+ */
+const TAG_IDS = array( 1453, 307, 6039, 1999 );
+
+/**
+ * The URL where this plugin will work.
+ *
+ * @var string
+ */
+const URL = 'https://make.wordpress.org/polyglots/?exclude=requests';
+
+/**
+ * Excludes the tags from the main query.
+ *
+ * @param $query
+ *
+ * @return void
+ */
+function pre_get_posts( $query ) {
+	global $wp;
+
+	if ( URL != home_url( add_query_arg( $_GET, $wp->request ) ) ) {
+		return;
+	}
+
+	if ( $query->is_main_query() ) {
+		$query->set( 'tag__not_in', TAG_IDS );
+	}
+}
+
+add_filter( 'pre_get_posts', __NAMESPACE__ . '\\pre_get_posts' );


wordpress.org/public_html/wp-content/plugins/wporg-polyglots-posts/wporg-polyglots-posts.php
```diff
@@ -0,0 +1,46 @@
...
+ * Excludes the tags from the main query.
+ *
+ * @param $query
+ *
+ * @return void
+ */
+function pre_get_posts( $query ) {
+	global $wp;
+
+	if ( URL != home_url( add_query_arg( $_GET, $wp->request ) ) ) {
```
dd32, line 37: Using absolute URLs here doesn't seem worthwhile, just check the GET variable is set or not IMHO.

```suggestion
	if ( empty( $_GET['exclude'] ) || 'requests' !== $_GET['exclude'] ) {
```

wordpress.org/public_html/wp-content/plugins/wporg-polyglots-posts/wporg-polyglots-posts.php
```diff
@@ -0,0 +1,46 @@
...
+		return;
+	}
+
+	if ( $query->is_main_query() ) {
+		$query->set( 'tag__not_in', TAG_IDS );
+	}
+}
+
+add_filter( 'pre_get_posts', __NAMESPACE__ . '\\pre_get_posts' );
```
dd32, line 46: I'd make this work with the `p2-resolved-posts` plugin, and remove the need to store a static list of IDs, which is bad for long-term maintainability.

Eg, this does the same thing, but works with the above plugin and doesn't have magic IDs:
(note: I've switched from `pre_get_posts` as it doesn't play nice with p2 plugins that work on `request`.)
```suggestion
function request( $query_vars ) {

	if ( empty( $_GET['exclude'] ) || 'requests' !== $_GET['exclude']  ) {
		return $query_vars;
	}

	if ( ! isset( $query_vars['tax_query'] ) ) {
		$query_vars['tax_query'] = [];
	}

	$query_vars['tax_query'][] = array(
		'taxonomy' => 'post_tag',
		'field'    => 'slug',
		'operator' => 'NOT IN',
		'terms'    => [ 'editor-requests', 'request', 'clpte', 'locale-requests' ],
	);

	return $query_vars;
}

add_filter( 'request', __NAMESPACE__ . '\\request' );
```

dd32, line 46:
```suggestion

add_action( 'pre_get_posts', __NAMESPACE__ . '\\pre_get_posts' );
```
Just noting that this is an action, not a filter, so should be `add_action` unless the above change to `request` is made.

dd32:
In addition to the above, I can't recall where the logic is (I don't think `?tags=` is core? I think it's a P2 thing?), but this URL will also filter to exclude those tags, which could just be linked to instead of `?exclude=requests` I guess.

https://make.wordpress.org/polyglots/?tags=-editor-requests,-request,-clpte,-locale-requests

amieiro:
Thank you for the code review, @dd32. For what I want to do, I can use the link that excludes the tags. I have already added it in a widget at https://make.wordpress.org/polyglots/, in the sidebar. I close the PR. 


<image src="https://user-images.githubusercontent.com/1667814/196614059-558afdaf-f4e0-4eb5-b61b-b2f48aff6d01.png" width=300>

PR #108, opened 2022-11-03
Add a plugin to create virtual projects from the patterns project

amieiro:
When a translator translates the patterns, all the strings are under the same project, so she can't focus in translating one pattern easily.

image

This PR creates "virtual projects" for the patterns, so each pattern has its own project. This approach try to improve the individual translation for patterns.

image

All the original strings are under the same project (project_id 473698) and the `references` field has the different URLs (patterns) in which the string in present.

This approach has some limitations:
- The virtual projects are alphabetically ordered.
- The search and the filter are not working.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-routes/inc/routes/class-locale.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-routes/inc/routes/class-locale.php
@@ -106,6 +106,13 @@ class Locale extends GP_Route {
 			)
 		);
 
+		$data['project']                = $project;
+		$data['sub_projects']           = $paged_sub_projects['projects'];
+		$data['pages']                  = $paged_sub_projects['pages'];
+		$data                           = apply_filters( 'gp_subprojects', $data );
+		$paged_sub_projects['projects'] = $data['sub_projects'];
+		$paged_sub_projects['pages']    = $data['pages'];
+
 		if ( ! $paged_sub_projects ) {
 			return $this->die_with_404();
 		}
@@ -588,6 +595,11 @@ class Locale extends GP_Route {
 			$status->untranslated_count = 0;
 			$status->all_count          = 0;
 			$status->percent_complete   = 0;
+		}
+
+		$status = apply_filters( 'gp_get_project_status', $status, $project, $locale, $set_slug );
+		if ( 0 != $status->sub_projects_count ) {
+			return $status;
 		}
 
 		$set = GP::$translation_set->by_project_id_slug_and_locale(
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-virtual-projects/wporg-gp-virtual-projects.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-virtual-projects/wporg-gp-virtual-projects.php
@@ -1,0 +1,181 @@
+<?php
+/**
+ * Plugin Name: WPORG Virtual projects for patterns.
+ * Description: Creates virtual projects for the patterns.
+ * Author: the WordPress team.
+ */
+class WPorg_Virtual_Projects {
+
+	/**
+	 * Adds filters to subprojects and to project status.
+	 */
+	public function __construct() {
+		add_filter( 'gp_subprojects', array( $this, 'add_virtual_projects' ), 10, 1 );
+		add_filter( 'gp_get_project_status', array( $this, 'gp_get_project_status' ), 10, 4 );
+	}
+
+	/**
+	 * Adds virtual projects for the patterns, so each pattern is shown as a different project.
+	 *
+	 * The received and returned data is an array with:
+	 * - $data['project']       The main project.
+	 * - $data['sub_projects']  The subprojects.
+	 * - $data['pages']         Pagination information.
+	 *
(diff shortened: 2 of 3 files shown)


wordpress.org/public_html/wp-content/plugins/wporg-gp-virtual-projects/wporg-gp-virtual-projects.php
```diff
@@ -0,0 +1,181 @@
...
+				$name = $row;
+			}
+			$subproject                      = new GP_Project();
+			$name                            = rtrim( $name, '/' );
+			$subproject->slug                = $name;
+			$subproject->id                  = $name;
+			$name                            = ucfirst( str_replace( '-', ' ', $name ) );
+			$subproject->name                = $name;
+			$subproject->description         = $name;
+			$subproject->path                = 'patterns/' . $subproject->slug;
```
akirk, line 82: This leads to a 404. The idea was to directly link to a reference search like this:
```suggestion
			$subproject->path                = '/../../../projects/patterns/core/' . $locale_slug . '/' . $set_slug. '/?filters[term_scope]=scope_references&filters[term]=' . urlencode(    $url ) . '&';
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-virtual-projects/wporg-gp-virtual-projects.php
```diff
@@ -0,0 +1,181 @@
...
+ * Description: Creates virtual projects for the patterns.
+ * Author: the WordPress team.
+ */
+class WPorg_Virtual_Projects {
+
+	/**
+	 * Adds filters to subprojects and to project status.
+	 */
+	public function __construct() {
+		add_filter( 'gp_subprojects', array( $this, 'add_virtual_projects' ), 10, 1 );
```
akirk, line 13:
```suggestion
		add_filter( 'gp_subprojects', array( $this, 'add_virtual_projects' ), 10, 3 );
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-virtual-projects/wporg-gp-virtual-projects.php
```diff
@@ -0,0 +1,181 @@
...
+	 * The received and returned data is an array with:
+	 * - $data['project']       The main project.
+	 * - $data['sub_projects']  The subprojects.
+	 * - $data['pages']         Pagination information.
+	 *
+	 * @param array $data The project, subprojects and pages.
+	 *
+	 * @return array The project, subprojects (or virtual projects) and pages.
+	 */
+	public function add_virtual_projects( $data ) {
```
akirk, line 29:
```suggestion
	public function add_virtual_projects( $data, $locale_slug, $set_slug ) {
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-routes/inc/routes/class-locale.php
```diff
@@ -106,6 +106,13 @@ public function get_locale_projects( $locale_slug, $set_slug = 'default', $proje
 			)
 		);
 
+		$data['project']                = $project;
+		$data['sub_projects']           = $paged_sub_projects['projects'];
+		$data['pages']                  = $paged_sub_projects['pages'];
+		$data                           = apply_filters( 'gp_subprojects', $data );
```
akirk, line 112:
```suggestion
		$data                           = apply_filters( 'gp_subprojects', $data, $locale_slug, $set_slug );
```

akirk (changes requested):
Nice. Just the URLs of the projects were broken, I suggested the changes necessary to make it work.

virtual-projects

dd32:
Is there any reason why we shouldn't just create real sub-projects?

The decision to not create sub-projects for each pattern was made when it was assumed we'd not have as many strings as we've ended up with, and/or that it'd be short-lived (ie. We'd end up having translations done within the Pattern Creator, rather than being done externally)

akirk:
We thought it would be useful to keep them all in the same project because of re-used strings resp. strings used in multiple patterns. Some patterns are just slight variations of others with some additional or replaced strings. It's quite useful to automatically be able to reuse those translations.

e.g.
Screenshot 2022-11-23 at 10 43 36

akirk:
Overall I think it's worth having this as a first step and we'll research if we can change it to do it with multiple projects.

It will need changes to how the patterns are imported to GlotPress (which I don't know how it's done) and to how translations are loaded.

dd32:
Thanks for the context @akirk.

We do have code on translate.w.org already for syncing translations between projects where the source-string remains the same, although that's limited to between plugin stable<->dev at present due to translator issues translating it between different plugins. The context difference between two blocks however seems much smaller, so your current approach seems reasonable.

My suggestion is just to not go down the path too far of attempting to make it work with filtering / subscriptions / favouriting / etc.. syncing between projects is going to be easier in the long-term than reinventing it all for this.

amieiro:
I have deployed this PR and I have written a [post](https://make.wordpress.org/polyglots/2023/02/16/virtual-projects-for-the-patterns-project/) to inform to the community, so I close this PR. Thank you for the review and feedback.

PR #111, opened 2022-11-29
Translate: Add CSS to convert translate.wordpress.org to full width

amieiro:
We are working to [bring some new tabs in the GlotPress right sidebar](https://github.com/GlotPress/gp-translation-helpers/pull/148), so we think it is interesting to convert translate.wordpress.org to full width. Below, you can see some screenshot from the new draft design.

These screenshots were taken in a screen with a 1920x1080 resolution.

image

image

image

image

image

image

image

image

image

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/style.css
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/style.css
@@ -176,8 +176,8 @@ select.is-small {
 
 .site-header .site-branding {
 	margin: 0 auto;
-	max-width: 1200px;
-	padding: 0 10px;
+	max-width: 100%;
+	padding: 0 20px;
 }
 
 .site-header.home {
@@ -238,7 +238,7 @@ p.site-title a {
 
 .navigation-main {
 	background: transparent;
-	max-width: 1200px;
+	max-width: 100%;
 	height: 32px;
 	margin: -32px auto 0 auto;
 	text-align: right;
@@ -478,8 +478,8 @@ table.translations tr.editor:hover {
 
 .gp-content {
 	margin: 20px auto;
-	max-width: 1200px;
-	padding: 0 10px;
+	max-width: 100%;
+	padding: 0 20px;
 }
 
 .gp-content h2 {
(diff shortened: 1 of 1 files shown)


dd32:
For certain things, full width makes sense (Such as those stat tables).

For others, full width can severely reduce readability (For example, the consistency tool, the project card listings, and some of the translation interfaces), especially when the content is not suited to a full width, which is why there's a hard-limit imposed in the first place.
Increasing that limit to account for a sidebar would however make sense, as would increasing the width less-than-fullwidth to account for any extra data that's been added.

Just my 2c, from not-a-UX-designer. Might be worth asking for a UX review of the GlotPress interface and this proposal

amieiro:
I have [deployed](https://meta.trac.wordpress.org/changeset/12287) a test at https://translate.wordpress.org/ to review this with the community.

pedro-mendonca:
I agree with @dd32.
It's more readable up to 1600px wide, keeping some margins. Still, it would be nice to have a designer review on this.

webaxones:
French translators and editors use the [SPTE webextension](https://github.com/webaxones/SPTE) allowing them to see all typographical errors related to the language. There are a few additional options on this webextension, including 2 allowing them both to switch to fullwidth, and at the same time to define a max-width because the screen definitions are numerous, one in wide screen, the other in 27" fullscreen, the 3rd on a laptop, etc.
For these 2 options I just override the ".gp-content" class, and ask for a simple percentage.
As this is a French extension, the interface is in French, but I have translated the options to show you what it looks like.

image

So you could just switch the site to full-width as this PR would allow and add a max-width option to the Translate settings to suit all cases.

nicomollet:
This PR needs to consider bigger screens. 
Screenshots in setups like 3840 x 2160 resolutions would make it easier to accept or not.

A maximum width should be set, in my opinion.

fxbenard:
Looks nice on my laptop, even if some margins will look nicer I guess… For my bigger screen it becomes much to big on the other hand. What about the max-width idea?

webaxones:
About the idea of the setting: it could have a default value. Eg 1600px if the setting is empty, but the user can decide to go to 80% of maximum resolution.

amieiro:
Ticket https://meta.trac.wordpress.org/ticket/6594.

amieiro:
I close this PR, because we have set a maximum of 1600px [here](https://github.com/WordPress/wordpress.org/pull/117).

PR #112, opened 2022-11-29
Add a "demo" tab to the plugin directory.

adamziel:
## Description

Adds a "demo" tab to a few selected plugins:

CleanShot 2022-11-29 at 21 43 35@2x

When loaded, it displays an iframe loading the [WordPress sandbox](https://github.com/WordPress/wordpress-wasm) on wasm.wordpress.net with the correct plugin configuration and a manually configured redirection URL.

## Shortcomings

This early version uses a hardcoded configuration to get things off the ground. Ideally, the configuration would be stored in the `plugin` post fields.

cc @tellyworth @dd32 @StevenDufresne


adamziel:
Let's consider only enabling the demo tab when explicitly requested, e.g. via `?enable_demo=1`. It's an experimental feature that can't support every plugin from day one, so perhaps that would be more fair. It's not a definitive decision, though, just a thought.

PR #114, opened 2022-12-07
Support profile - move security settings onto separate screen

### LINKED ISSUE #26: Move Account settings on to separate screen
> I think it would make sense to split the "Account" subsection that has Email/Password onto a new tab, and list that along with 2FA on that page.
> https://github.com/WordPress/wporg-two-factor/issues/15#issuecomment-1316376334

Breaking this off of #15 since it's a distinct task, and something that needs to be done in the `wporg-support` theme.

- [x] Create new Edit Account screen at https://wordpress.org/support/users/me/edit/account/
- [x] Populate screen with custom UI
- [x] Update Rosetta forums with same changes as main forums
- [x] Move Account settings from https://wordpress.org/support/users/me/edit/ to new screen -- staged in https://github.com/WordPress/wordpress.org/pull/114 but needs to wait until we launch for all users. Or maybe move it all to Profiles instread? Will still need to have the password/lang/etc settings removed from forums, unless ignoring them for now.
- [x] Update `Your password can be changed via...` link on `https://profiles.wordpress.org/{username}/profile/edit/group/1/` -- `wp-content/themes/profiles.wordpress.org/members/single/profile/edit-custom-general.php`

iandunn:
WIP

Fixes https://github.com/WordPress/wporg-two-factor/issues/26

### Regular Profile:

Screenshot 2023-09-18 at 12 55 55 PM

### Forum Profile

Screenshot 2023-09-18 at 12 57 31 PM

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-support/bbpress/form-user-edit.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-support/bbpress/form-user-edit.php
@@ -1,0 +1,184 @@
+<?php
+
+/**
+ * The Account section has been modified so that email and password link to the `/edit/account/` page, which is
+ * the canonical locations for those settings.
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+?>
+
+<form id="bbp-your-profile" method="post" enctype="multipart/form-data">
+
+	<h2 class="entry-title"><?php esc_html_e( 'Name', 'bbpress' ); ?></h2>
+
+	<?php do_action( 'bbp_user_edit_before' ); ?>
+
+	<fieldset class="bbp-form">
+		<legend><?php esc_html_e( 'Name', 'bbpress' ); ?></legend>
+
+		<?php do_action( 'bbp_user_edit_before_name' ); ?>
+
+		<div>
+			<label for="first_name"><?php esc_html_e( 'First Name', 'bbpress' ); ?></label>
+			<input type="text" name="first_name" id="first_name" value="<?php bbp_displayed_user_field( 'first_name', 'edit' ); ?>" class="regular-text" />
+		</div>
+
+		<div>
+			<label for="last_name"><?php esc_html_e( 'Last Name', 'bbpress' ); ?></label>
+			<input type="text" name="last_name" id="last_name" value="<?php bbp_displayed_user_field( 'last_name', 'edit' ); ?>" class="regular-text" />
+		</div>
+
+		<div>
+			<label for="nickname"><?php esc_html_e( 'Nickname', 'bbpress' ); ?></label>
+			<input type="text" name="nickname" id="nickname" value="<?php bbp_displayed_user_field( 'nickname', 'edit' ); ?>" class="regular-text" />
+		</div>
+
+		<div>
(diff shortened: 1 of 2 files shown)


iandunn:
@dd32, @pkevan, @tellyworth :  this pr creates a new `https://wordpress.org/support/users/{username}/edit/account/` page.  that works in conjunction with the [ui-stub](https://github.com/WordPress/wporg-two-factor/tree/ui-stub) branch in `wporg-two-factor`, which renders the `wporg-two-factor/settings` block into new page. that block would use REST API requests like Dion is building in https://github.com/WordPress/two-factor/pull/504 to save the 2FA settings, and  Core `users` endpoint to save email/password/language.

the text is just rough stubs and there's no design, but take a look at the code for both branches and let me know if you think this is a good architecture to build on

pkevan:
I'm not deeply familiar with the profiles part, but seems like a sensible way to approach this initially.

iandunn:
I've committed the parts that we can deploy now in 4db5f38 and `r19711-dotorg`, so now the branch just has the parts that need to wait until we launch everything for users.

StevenDufresne:
I think we may need another path to security settings.

Logged in, when I click "Edit Profile", I land on https://profiles.wordpress.org/{username}/profile/edit/ which doesn't have any references to security settings. I would have to already know the settings is in our "Forum profiles".

Should we add a section to the General tab that leads directly to 2fa?


Screenshot 2023-09-15 at 2 06 45 PM

I'm not sure if that's the best language or link treatment :).

pkevan:
We probably also want to change the label/link in the Forum profile from `Edit Account` to be more descriptive too
Screenshot 2023-09-15 at 08 23 50

iandunn:
> change the label/link in the Forum profile from Edit Account to be more descriptive too

🤔 Do you have any ideas? It was named "Account" because previously the email/password settings were in a section of the profile page named "Account"

Renaming it to "Security" makes sense for the 2FA settings, but I don't think users would expect that for email. Maybe "Account Settings" instead of "Edit Account" ?

iandunn:
> Should we add a section to the General tab that leads directly to 2fa?

What do you think about this?

Screenshot 2023-09-15 at 10 01 34 AM

That way we include "Security", with might catch folks eyes while they're scanning, but it still has "password" and "two-factor authentication". I'm also open to having a sections for each, though.

StevenDufresne:
Yeah, that looks good to me.

pkevan:
> > change the label/link in the Forum profile from Edit Account to be more descriptive too
> 
> 🤔 Do you have any ideas? It was named "Account" because previously the email/password settings were in a section of the profile page named "Account"
> 
> Renaming it to "Security" makes sense for the 2FA settings, but I don't think users would expect that for email. Maybe "Account Settings" instead of "Edit Account" ?

It's just labelled `Security` on WP.com, although I agree it's somewhat confusing when it contains email, although don't think it's a big issue - might be good for consistency.

iandunn:
🤔 Changing the menu label would also entail changing the rewrite rule and links to the page.

I don't personally think it's worth it, but I'm not opposed to it either. I'd rather not tack that onto this PR, though, do you mind opening another issue for it?

PR #116, opened 2022-12-21
Translate: Cron job to store translation stats daily

trymebytes:
In this PR, when the daily stats cron job is run we also store the  translation stats in the new polyglot_stats table (see https://meta.trac.wordpress.org/ticket/6571).

We will also create a script that can display the latest stats for copy & paste for the Weekly Polyglots Meetings.


wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-stats.php
```diff
@@ -162,9 +176,230 @@ public function __invoke( bool $echo_the_values = false ) {
...
+
+		$current_date_data = $this->get_data_for_date( $current_date );
+		$old_date_data     = $this->get_data_for_date( $old_date );
+
+		if ( ! $current_date_data || ! $old_date_data ) {
+			return;
+		}
+		$stats_diff = array();
+		foreach ( $current_date_data as $key => $value ) {
+			$stats_diff[ $key ] = $value - $old_date_data->$key;
```
akirk: I think it'd be better for consistency to also put this into an object.
```suggestion
			$stats_diff->$key = $value - $old_date_data->$key;
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-stats.php
```diff
@@ -162,9 +176,230 @@ public function __invoke( bool $echo_the_values = false ) {
...
+		}
+		$old_date = is_null( $old_date ) ? gmdate( 'Y-m-d', strtotime( '-1 week' ) ) : $old_date;
+
+		$current_date_data = $this->get_data_for_date( $current_date );
+		$old_date_data     = $this->get_data_for_date( $old_date );
+
+		if ( ! $current_date_data || ! $old_date_data ) {
+			return;
+		}
+		$stats_diff = array();
```
akirk:
```suggestion
		$stats_diff = new stdClass;
```

or 
```suggestion
		$stats_diff = (object) array();
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-stats.php
```diff
@@ -162,9 +176,249 @@ public function __invoke( bool $echo_the_values = false ) {
...
+			$this->print_wpcli_heading( 'Summary for weekly stats' );
+		}
+		$stats_diff = new \stdClass();
+		foreach ( $current_date_data as $key => $value ) {
+			$stats_diff->$key = $value - $old_date_data->$key;
+		}
+		$all_locales_data = get_locales_data();
+		$stats_data       = $all_locales_data['status_counts'];
+
+		$code  = 'Summary for weekly stats' . PHP_EOL;
```
akirk: Should this be a headline? Looking at the rest of the source you should be using the `create_gutenberg_heading()` function for that one.

akirk: See https://github.com/WordPress/wordpress.org/blob/26bfc5b422e3b00221bb5208dc135e7f79191239/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-stats.php#L248-L252

akirk:
This is good to go for me!

PR #120, opened 2023-02-15
Sync fuzzy, rejected and changes requested translations in both dev and stable version

trymebytes:
Ticket - https://meta.trac.wordpress.org/ticket/5679

At the moment, translations are synced to both dev and stable versions only when it is approved on either versions. This implies that when a translation is approved and set to current in dev, the stable version is automatically updated with that approved translation but when a translation is rejected or set to fuzzy or changes requested on dev, the stable version is not updated and vice versa.

With this PR, when a translation is marked as fuzzy, rejected or changes requested on dev version, the same is applied/synced in the stable version and vice versa.

**How to test**
1. Suggest a translation on the dev branch if there are no waiting strings for a project.
2. As a PTE/GTE, set selected waiting translations to rejected, fuzzy or changes requested.
3. The translation rejected on the dev branch should also be rejected on the stable branch.
4. Repeat steps 1-3 starting with the stable branch in step 1.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
@@ -141,8 +141,6 @@ class Translation_Sync {
 			return;
 		}
 
-		do_action( 'wporg_translate_translation_synced', $translation );
-
 		$project = GP::$project->one(
 			"SELECT p.* FROM {$wpdb->gp_projects} AS p JOIN {$wpdb->gp_originals} AS o ON o.project_id = p.id WHERE o.id = %d",
 			$translation->original_id
@@ -267,6 +265,8 @@ class Translation_Sync {
 		if ( ! $translation ) {
 			return false;
 		}
+
+		do_action( 'wporg_translate_translation_synced', $copy, $translation );
 
 		$translation->set_status( $copy->status );
 		gp_clean_translation_set_cache( $new_translation_set->id );
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
@@ -135,10 +135,13 @@ class Translation_Sync {
 	public function queue_translation_for_sync( $translation ) {
 		global $wpdb;
 
-		// Only propagate current translations without warnings.
-		if ( 'current' !== $translation->status || ! empty( $translation->warnings ) ) {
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
```diff
@@ -135,8 +139,9 @@ public function sync_plugin_translations_on_commit( $args ) {
 	public function queue_translation_for_sync( $translation ) {
 		global $wpdb;
 
-		// Only propagate current translations without warnings.
-		if ( 'current' !== $translation->status || ! empty( $translation->warnings ) ) {
+		$allowed_statuses = array( 'current', 'rejected', 'changesrequested', 'fuzzy' );
```
akirk, line 138: If we sync the `changesrequested` status, I think the rejection reason will not be visible. We could implement something like on bulk reject where there is a "linking comment". We shouldn't do this here but we could add a `do_action( 'translation_synced', $translation );` in anticipation of that.

akirk: Probably some additional parameters would be useful like source and target project.

wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
```diff
@@ -9,23 +9,23 @@
 
 class Translation_Sync {
 
-	private $queue = [];
+	private $queue = array();
```
ocean90: Can you please revert all these `[]` to `array()` changes to the keep the PR clean?

wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
```diff
@@ -135,11 +139,14 @@ public function sync_plugin_translations_on_commit( $args ) {
...
 
-		// Only propagate current translations without warnings.
-		if ( 'current' !== $translation->status || ! empty( $translation->warnings ) ) {
+		$allowed_statuses = array( 'current', 'rejected', 'changesrequested', 'fuzzy' );
+		// Do not propagate waiting translations and other translations with warnings.
+		if ( ! in_array( $translation->status, $allowed_statuses ) || ! empty( $translation->warnings ) ) {
 			return;
 		}
 
+		do_action( 'translation_synced', $translation );
```
ocean90:
```suggestion
		do_action( 'wporg_translate_translation_synced', $translation );
```

This should get a prefix so it's clear that this is a custom hook.

akirk:
Looks good me now!

PR #123, opened 2023-02-17
Add robots header for hosting test results

tellyworth:
Fixes https://meta.trac.wordpress.org/ticket/5398

This adds noindex on any p2 pages with a `rpage=` query string, where rpage > 1. That probably only affects the hosting page, but even if it affects something else that might be a good thing.

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-breathe/functions.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-breathe/functions.php
@@ -51,6 +51,11 @@ function no_robots( $noindex ) {
 		$noindex = true;
 	}
 
+	// This is used by https://github.com/WordPress/phpunit-test-reporter/blob/master/src/class-display.php on the test reporter page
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/themes/pub/wporg-breathe/functions.php
```diff
@@ -51,6 +51,11 @@ function no_robots( $noindex ) {
 		$noindex = true;
 	}
 
+	// This is used by https://github.com/WordPress/phpunit-test-reporter/blob/master/src/class-display.php on the test reporter page
+	if ( isset( $_GET['rpage'] ) && $_GET['rpage'] > 1 ) {
```
dd32, line 55:
```suggestion
	if ( isset( $_GET['rpage'] ) ) {
```

I think we can just check to see if it's set, since the ?rpage=1 url shouldn't really exist, but is "incorrectly" linked to by the page navigation.

dd32 (approved):
Seems reasonable to me, noting that probably don't need the `rpage > 1` part though.

PR #126, opened 2023-02-22
Fix fuzzy rejection sync

trymebytes:
Trac ticket: https://meta.trac.wordpress.org/ticket/6788

When a fuzzy translation is rejected in one project(e.g dev) it appears in the other project(e.g stable) as current.

This PR fixes the issue and ensure that statuses are constant across projects even if the translation had been existing on either dev or stable.

To test, set a translation to fuzzy on dev(or stable), then reject that fuzzy translation on stable or (dev), the translation should appear as `rejected` on both dev and stable.


akirk:
I tested approving and rejecting fuzzies and current translations in stable checked dev and vice versa. It always synced it properly for me. So 👍 from me.

PR #127, opened 2023-03-01
Reduce scope of rewrite rules

pkevan:
Avoids too many redirect for URLs containing more common strings (faq,stats).

See: https://meta.trac.wordpress.org/ticket/6819


johnbillion:
This might have just broken something: https://meta.trac.wordpress.org/ticket/6824

PR #131, opened 2023-03-15
Add support for listing component Slack Channels.

desrosj:
This adds a new custom field for a Slack channel ID of a component. When one is present, a Slack logo will display in a new column on the Components page that links off to that room.

Trac ticket: https://meta.trac.wordpress.org/ticket/6862

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/trac-notifications/trac-components.php
+++ b/wordpress.org/public_html/wp-content/plugins/trac-notifications/trac-components.php
@@ -41,6 +41,10 @@ class Make_Core_Trac_Components {
 		return ucfirst( $this->trac );
 	}
 
+	function slack_channel_url( $channel_id ) {
+		return 'https://wordpress.slack.com/archives/' . sanitize_text_field( $channel_id );
+	}
+
 	function init() {
 		add_shortcode( 'trac-select', array( $this, 'shortcode_select' ) );
 		add_shortcode( 'logged-in', array( $this, 'shortcode_logged_in' ) );
@@ -151,8 +155,11 @@ class Make_Core_Trac_Components {
 		if ( ! $this->page_is_component( $post ) ) {
 			return;
 		}
-		$value = get_post_meta( $post->ID, '_active_maintainers', true );
-		echo '<p><label for="active-maintainers">Active maintainers (WP.org usernames, comma-separated)</label> <input type="text" class="large-text" id="active-maintainers" name="active-maintainers" value="' . esc_attr( $value ) . '" />';
+		$active_maintainers = get_post_meta( $post->ID, '_active_maintainers', true );
+		echo '<p><label for="active-maintainers">Active maintainers (WP.org usernames, comma-separated)</label> <input type="text" class="large-text" id="active-maintainers" name="active-maintainers" value="' . esc_attr( $active_maintainers ) . '" />';
+
+		$slack_channel = get_post_meta( $post->ID, '_slack_channel', true );
+		echo '<p><label for="slack-channel">Slack channel ID</label> <input type="text" class="large-text" id="slack-channel" name="slack-channel" value="' . esc_attr( $slack_channel ) . '" />';
 	}
 
 	function save_post( $post_id, $post ) {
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/plugins/trac-notifications/trac-components.php
```diff
@@ -423,10 +434,28 @@ function the_content( $content ) {
...
+
+		echo '<h3>Help maintain this component</h3>';
+
+		echo "Want to help maintain this component? There's several ways to get involved." . "\n\n";
+
+		echo '<ul>';
+
+		echo '<li><a href="' . $this->trac_url() . '/component/' . esc_attr( rawurlencode( $component ) ) . '">Find some open tickets to work on in Trac.</a>';
+
+		$slack_channel = get_post_meta( $post->ID, '_aslack_channel', true );
```
dd32:
```suggestion
		$slack_channel = get_post_meta( $post->ID, '_slack_channel', true );
```

typo

wordpress.org/public_html/wp-content/plugins/trac-notifications/trac-components.php
```diff
@@ -732,6 +761,15 @@ function component_table_row( $post ) {
 			echo '<td></td>';
 		}
 
+		$slack_channel = get_post_meta( $post->id, '_slack_channel', true );
```
dd32:
```suggestion
		$slack_channel = get_post_meta( $post->ID, '_slack_channel', true );
```
typo

wordpress.org/public_html/wp-content/plugins/trac-notifications/trac-components.php
```diff
@@ -693,7 +722,7 @@ function component_table_row( $post ) {
 		static $once = true;
 		if ( $once ) {
 			$once = false;
-			echo '<thead><tr><td>Component</td><td>Tickets</td><td>7 Days</td><td>0&nbsp;Replies</td><td>Maintainers</td></tr></thead>';
+			echo '<thead><tr><td>Component</td><td>Tickets</td><td>7 Days</td><td>0&nbsp;Replies</td><td>Slack</td><td>Maintainers</td></tr></thead>';
```
dd32: Adding this extra column unfortunately doesn't really work due to the width of the column:

_edit: I missed the mention on Trac that this text should be hidden with just the icon._

| PR | With icon only | Alternative |
| --- | --- | ---|
| Screenshot 2023-03-16 at 12 48 55 pm | Screenshot 2023-03-16 at 12 57 54 pm | Screenshot 2023-03-16 at 1 02 07 pm |

dd32 (commented):
As noted in the review, I don't think there's room for a dedicated Slack channel column with the current components overview table.

I don't see any issue with the rest of the PR though.

wordpress.org/public_html/wp-content/plugins/trac-notifications/trac-components.php
dd32: The alternative screenshot above was done with this code, the inline `float: right` should probably be a `class="slack-float-right"` or something i guess.
```diff
 		}

 		echo '<tr>';
+		echo '<td>';
 		if ( $post->post_parent ) {
-			echo '<td>&mdash; <a href="' . get_permalink() . '">' . $post->post_title . '</a></td>';
+			echo '&mdash; <a href="' . get_permalink() . '">' . $post->post_title . '</a>';
 		} else {
-			echo '<td><a href="' . get_permalink() . '"><strong>' . $post->post_title . '</strong></a></td>';
+			echo '<a href="' . get_permalink() . '"><strong>' . $post->post_title . '</strong></a>';
 		}

+		$slack_channel = get_post_meta( $post->ID, '_slack_channel', true );
+		if ( ! empty( $slack_channel ) ) {
+			echo '<span style="float: right"><a href="' . $this->slack_channel_url( $slack_channel ) . '"><span class="screen-reader-text">Join the ' . $component . ' Slack channel</span></a></span>';
+		}
+
+		echo '</td>';
+
 		$open_tickets = 0;
 		if ( ! empty( $this->breakdown_component_type[ $component ] ) ) {
 			$open_tickets = array_sum( $this->breakdown_component_type[ $component ] );
```

desrosj: I like it. Made this change (including adding a CSS class to avoid the `float: right` inline) in 236a4eaecbf34c52c6098000959a3cc803b4e4ca.

PR #132, opened 2023-03-17
Fix multi-status problem

trymebytes:


wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
```diff
@@ -249,7 +249,18 @@ private function copy_translation_into_set( $translation, $new_translation_set,
...
-				$_existing_translation->set_status( $translation->status );
+				if ( 'current' !== $translation->status && 'changesrequested' != $translation->status ) {
+					$_existing_translation->set_status( $translation->status );
+				}
+
+				if ( 'current' == $translation->status ) {
+					$_existing_translation->set_as_current();
+				}
+
+				if ( 'changesrequested' == $translation->set_as_changesrequested() ) {
```
akirk: The method returns a boolean or null, not a string, see https://github.com/GlotPress/GlotPress/blob/82d29e3b7638a662fc0bc9d7f1dc247c3ad32a57/gp-includes/things/translation.php#L651
```
public function set_as_changesrequested(): ?bool {
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-plugin-directory/inc/sync/class-translation-sync.php
```diff
@@ -249,7 +249,18 @@ private function copy_translation_into_set( $translation, $new_translation_set,
...
+					$_existing_translation->set_as_current();
+				}
+
```
akirk:
```suggestion
				if ( 'current' == $translation->status ) {
					$_existing_translation->set_as_current();
				} elseif ( 'changesrequested' == $translation->status ) {
					$_existing_translation->set_as_changesrequested()
				} else {
					$_existing_translation->set_status( $translation->status );
				}

```

PR #130, opened 2023-03-14
Add UI for ChatGPT settings

trymebytes:
In this PR, we add the UI for storing a user's OpenAI credentials to be used for generating translation suggestions and for other implementations.

Screenshot 2023-03-14 at 16 00 45

PR #134, opened 2023-03-21
Experiment: Add OpenAI (ChatGPT) and Deepl support in the Translation Memory

amieiro:
## What?

We are testing [ChatGPT](https://openai.com/blog/chatgpt) and [DeepL](https://www.deepl.com/) as helpers to improve the translator's work.

image

This PR adds some proposed translations (suggestions) as new rows at the bottom of the "Translation Memory". The end user can activate one of this, or both, in the [translation settings](https://translate.wordpress.org/settings/).

image

### ChatGPT

The user can adjust the **prompt** and the **temperature** for ChatGPT. This PR is using:
* The glossary for the target language.
* The gpt-3.5-turbo [model](https://platform.openai.com/docs/api-reference/completions/create#completions/create-model).
* A maximum number of 1000 [tokens](https://platform.openai.com/docs/api-reference/completions/create#completions/create-max_tokens).
* [n](https://platform.openai.com/docs/api-reference/completions/create#completions/create-n) (how many completions to generate for each prompt) to 1.
* A 20 s. timeout to query the OpenAI API.

### Deepl

This PR is using the [free API](https://www.deepl.com/account/).

We always translate from EN-US - English (American). Deepl only supports [these target languages](https://www.deepl.com/es/docs-api/translate-text/translate-text/):
* BG - Bulgarian
* CS - Czech
* DA - Danish
* DE - German
* EL - Greek
* EN-GB - English (British)
* ES - Spanish
* ET - Estonian
* FI - Finnish
* FR - French
* HU - Hungarian
* ID - Indonesian
* IT - Italian
* JA - Japanese
* KO - Korean
* LT - Lithuanian
* LV - Latvian
* NB - Norwegian (Bokmål)
* NL - Dutch
* PL - Polish
* PT-BR - Portuguese (Brazilian)
* PT-PT - Portuguese (Portugal Portuguese)
* RO - Romanian
* RU - Russian
* SK - Slovak
* SL - Slovenian
* SV - Swedish
* TR - Turkish
* UK - Ukrainian
* ZH - Chinese (simplified)

This PR is using the UI from https://github.com/WordPress/wordpress.org/pull/130, with some updates.


wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/settings-edit.php
```diff
@@ -60,7 +60,39 @@
...
+			</label></th>
+		<td><input type="text" class="openai_api_key" id="default_sort[openai_api_key]" name="default_sort[openai_api_key]" value="<?php echo esc_html( gp_array_get( $gp_default_sort, 'openai_api_key', '' ) ); ?>" placeholder="Enter your OpenAI API key" /></td>
+	</tr>
+	<tr>
+		<th><label for="default_sort[openai_custom_prompt]"><?php esc_html_e( 'Custom Prompt', 'glotpress' ); ?></label></th>
+		<td><textarea class="openai_custom_prompt" id="default_sort[openai_custom_prompt]" name="default_sort[openai_custom_prompt]" placeholder="Enter your custom prompt for ChatGPT translation suggestions"><?php echo esc_html( gp_array_get( $gp_default_sort, 'openai_custom_prompt', '' ) ); ?></textarea></td>
+	</tr>
+	<tr>
+		<th><label for="default_sort[openai_temperature]"><?php esc_html_e( 'Temperature', 'glotpress' ); ?></label></th>
+		<td><input type="number" min="0" max="2" step=".1" class="openai_temperature" id="default_sort[openai_temperature]" name="default_sort[openai_temperature]" value="<?php echo esc_html( gp_array_get( $gp_default_sort, 'openai_temperature', 0 ) ); ?>" placeholder="Enter your OpenAI key" /></td>
```
akirk, line 126: Should we change the default temperature to something not 0?

amieiro: I was making some tests with different temperatures with my 2 mother languages (Galician and Spanish) and the best value for these languages is 0, so I think it is the most appropriate value to set as default. Maybe be another different values will be more accurate with another languages.

wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/routes/class-translation-memory.php
```diff
@@ -31,13 +32,196 @@ public function get_suggestions( $project_path, $locale_slug, $set_slug ) {
...
+		}
+		$deepl_response = wp_remote_post(
+			$free_url,
+			array(
+				'timeout' => 20,
+				'body'    => array(
+					'auth_key'    => $deepl_api_key,
+					'text'        => $original_singular,
+					'source_lang' => $source_lang,
+					'target_lang' => $target_lang,
```
akirk, line 202: Could we extend the mapping so that it can also include support [for the `formality` parameter](https://www.deepl.com/docs-api/translate-text/translate-text/)? Without it, it is unusable in German since we use informal German in WordPress but DeepL creates formal German.

amieiro: Done! I have tested with Spanish (Spain) and works fine! Thank you for the suggestion. 

image

akirk: Hm, actually I thought we would just add it to the mapping since inside of translate.wordpress.org it's quite clear which language has which formality. Maybe we can do both to start with but add it to the mapping in the end?

akirk:
I've played around with it and it works great. Could we keep some stats? Maybe we can just store them in user meta?
Specifically I think these will be interesting:
- Number of queries (= one per get-other-language-suggestions endpoint invocation)
- Number of tokens/characters used. [OpenAI reports](https://platform.openai.com/docs/api-reference/completions/create) `['usage']['total_tokens']` in its response, for [DeepL we need to count the English UTF8 codepoints](https://www.deepl.com/docs-api/general/get-usage/) (so `mb_strlen()`).
- Number of times the provider was used unmodified.
- Number of times the provider was used modified.

So we'd add stats to the endpoint and for saving the latter we'd need to extend GlotPress a little. Save in JS that a button was clicked and with what text and then when submitting the translation compare the text to the actual submission so that we can categorize it in the buckets.

wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/routes/class-translation-memory.php
amieiro: I have moved the select item for the language formality from the user preferences to a method with the values hard-coded for each language. I have [pinged](https://wordpress.slack.com/archives/C02RP50LK/p1679476128686079) to the GTE for each target language supported in Deepl, so this array should be updated as we receive information from the GTEs.

image

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/settings-edit.php
```diff
@@ -96,13 +116,28 @@
...
+							echo ' ' . esc_html( sprintf( _n( '(%d used without modifications)', '(%d used without modifications)', 'glotpress' ), gp_array_get( $gp_external_translations, 'deepl_same_translations_used', 0 ) ) );
+						}
+						echo '</small>';
+					}
+					?>
+				<br>
 				<a href="https://www.deepl.com/account/usage" target="_blank"><small>
 						<?php
 						/* translators: %d: number of chars translated with Deepl. */
 						echo esc_html( sprintf( __( '(Chars translated with Deepl: %d)', 'glotpress' ), gp_array_get( $gp_default_sort, 'deepl_chars_used', 0 ) ) );
```
akirk: These parenthesis should also be removed for consistency.

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/settings-edit.php
```diff
@@ -96,13 +116,28 @@
 	<tr>
 		<th><label for="default_sort[deepl_api_key]">
 				<?php esc_html_e( 'Deepl Free API Key', 'glotpress' ); ?>
+			</label>
+					<?php
+					if ( gp_array_get( $gp_external_translations, 'deepl_translations_used', 0 ) > 0 ) {
+						echo '<br>';
+						echo '<small>';
+						/* translators: %d: number of OpenAI translations used. */
+						echo esc_html( sprintf( _n( '%d Deepl translation used', '%d Deepl translations used', 'glotpress' ), gp_array_get( $gp_external_translations, 'deepl_translations_used', 0 ) ) );
```
akirk: They themselves always spell DeepL with a capital L at the end. Could you replace all occurences of outputted text so that it conforms to that?

akirk: Also, could we, to make it more clear add a "with modifications" here?

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/js/editor.js
```diff
@@ -69,6 +71,48 @@
...
+	/**
+	 * Saves the number of external suggestions used and used without modification.
+	 *
+	 * @return {void}
+	 **/
+	function saveExternalSuggestions() {
+		var $button = $(this);
+		var $row = $button.closest('tr.editor');
+		var $originalId = $row.attr('id').substring(7);
+		var $translation = $row.find('.lt-mirror__canvas').text().trim();
```
akirk: That selector doesn't seem right. It causes for me an empty translation to be reported.

akirk (changes requested):
Very nice! It almost works! The actually used translation is not reported for me, hence all my translations were recorded as "modified". Also a small number of inconsistencies. But otherwise, really great, also that their stats are transparently exposed to the user!

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/js/editor.js
amieiro: Done. Thank you for catching this!

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/settings-edit.php
amieiro: Done!

wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/templates/translation-memory-suggestions.php
```diff
@@ -23,5 +22,43 @@
...
+			echo '<span aria-hidden="true" class="translation-suggestion__translation-raw">' . esc_translation( $suggestion['translation'] ) . '</span>';
+
+			echo '<button type="button" class="button is-small copy-suggestion">Copy</button>';
+		echo '</div>';
+		echo '</li>';
+	}
+	foreach ( $deepl_suggestions as $suggestion ) {
+		echo '<li>';
+		echo '<div class="translation-suggestion with-tooltip deepl" tabindex="0" role="button" aria-pressed="false" aria-label="Copy translation">';
+			echo '<span class="deepl-suggestion__score">Deepl</span>';
```
akirk: You forgot to change this one, or update it in the PR:
```suggestion
			echo '<span class="deepl-suggestion__score">DeepL</span>';
```

akirk (approved):
This works well now! One observation is that it now takes longer for the TM to appear. Maybe we can decouple them in the next iteration?

amieiro:
This PR has been deployed ([#12510](https://meta.trac.wordpress.org/changeset/12510), [12511](https://meta.trac.wordpress.org/changeset/12511)), so I close it.

PR #137, opened 2023-04-03
Make all the requests async in the Translation Memory

amieiro:
## Why?

In #134 we add the integration with OpenAI (ChatGPT) and DeepL. In that PR, the API calls were synchronous, so if any API takes a lot of time to respond, the whole response will be affected.

## What?

In this PR I have improved this approach, splitting the existing AJAX call in 3 calls (internal translation memory, OpenAI and DeepL). 

## Testing

1. Test with empty and wrong API keys for both external systems. You should get only the suggestions from the internal TM.
2. Add a good OpenAI key. You should get the suggestions from the internal TM and from OpenAI.
3. Add a good DeepL key. You should get the suggestions from the 3 systems: internal TM, from OpenAI and from DeepL.
4. Change the OpenAI key to a bad one. You should get the suggestions from the internal TM and from DeepL.
5. Change the DeepL key to a bad one. You should get the suggestions from the internal TM.

## Design decisions 
- If the response is incorrect or empty, the AJAX query responses with "No suggestions.".
- The new AJAX calls (OpenAI and DeepL) are only made if the user has a non-empty API key in her settings.
- If the user has an incorrect API key, the external query is made. 
- If all the responses are incorrect, I remove the "No suggestions." duplicates. 
- If at least one of the 3 queries has some content in the response, I remove the "No suggestions." replies from the other systems.

image

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-plugin.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-plugin.php
@@ -127,16 +127,18 @@ class Plugin {
 		);
 
 		$gp_default_sort         = get_user_option( 'gp_default_sort' );
-		$get_openai_translations = empty( trim( gp_array_get( $gp_default_sort, 'openai_api_key' ) ) ) ? 'false' : 'true';
-		$get_deepl_translations  = empty( trim( gp_array_get( $gp_default_sort, 'deepl_api_key' ) ) ) ? 'false' : 'true';
+		$get_openai_translations = ! empty( trim( gp_array_get( $gp_default_sort, 'openai_api_key' ) ) );
+		$get_deepl_translations  = ! empty( trim( gp_array_get( $gp_default_sort, 'deepl_api_key' ) ) );
 
 		wp_localize_script(
 			'gp-translation-suggestions',
 			'gpTranslationSuggestions',
 			array(
-				'nonce'                   => wp_create_nonce( 'gp-translation-suggestions' ),
-				'get_openai_translations' => $get_openai_translations,
-				'get_deepl_translations'  => $get_deepl_translations,
+				'nonce'                     => wp_create_nonce( 'gp-translation-suggestions' ),
+				'get_external_translations' => array(
+					'get_openai_translations' => $get_openai_translations,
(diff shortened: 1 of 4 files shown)


wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-plugin.php
```diff
@@ -112,25 +114,41 @@ public function pre_tmpl_load( $template, $args ) {
...
 			'gp-translation-suggestions',
 			plugins_url( './js/translation-suggestions.js', PLUGIN_FILE ),
-			[ 'gp-editor' ],
+			array( 'gp-editor' ),
 			'20190510'
 		);
 
+		$gp_default_sort         = get_user_option( 'gp_default_sort' );
+		$get_openai_translations = empty( trim( gp_array_get( $gp_default_sort, 'openai_api_key' ) ) ) ? 'false' : 'true';
+		$get_deepl_translations  = empty( trim( gp_array_get( $gp_default_sort, 'deepl_api_key' ) ) ) ? 'false' : 'true';
```
akirk: It works but I am curious why you're using strings here instead of booleans.

akirk (approved):
I tested it and it worked fine. Not a blocker but I noticed that if for example the TM comes back first with no suggestions it for a brief moment shows "No suggestions" only to be replaced with a result from DeepL or OpenAI. Maybe we should show multiple loading indicators?

wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-plugin.php
amieiro: I have resolved this in [this commit](https://github.com/WordPress/wordpress.org/pull/137/commits/532d018257b76097aa340fb05efc2ae8cdb277cb).

amieiro: Related with the "No suggestions" situation, I think the best approach is to show the translations when we receive them and finally, if we don't have any translation, to show the "No suggestions" message. I will do this in a different PR.

amieiro:
This PR has been [committed to the SVN](https://meta.trac.wordpress.org/changeset/12527) and deployed, so I close it.

PR #139, opened 2023-04-18
Add the GlotPress playground

akirk:
This adds playground.wordpress.net to translate.wordpress.org:
Screenshot 2023-04-18 at 16 10 24

After translating, you can directly submit back to translate.wordpress.org:
wpgp-glotpress-contribute-back-on-dot-org


akirk:
Shipped

PR #141, opened 2023-06-01
Translate: Preload the suggestions

amieiro:
Currently, in the translation environment (translate.w.org), when we translate a project, we have suggestions from our own translation memory (TM), from OpenAI (ChatGPT) and from DeepL. We have the translations from other languages, too.

This PR adds a local cache for the previous suggestions, in 4 variables, so it caches the suggestions and translations from other languages:
- For the first row when the page loads, so when the user opens the first row, she gets the suggestions immediately.
- For the next row, when she opens a row, so when the user opens the next row, she gets the suggestions immediately.

We only cache what we believe to be the most likely row to be opened next, not all of them, as this avoids making excessive API calls, with the consequent economic cost (OpenAI and DeepL).

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/js/translation-suggestions.js
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/js/translation-suggestions.js
@@ -48,6 +48,8 @@
 	function fetchSuggestions( $container, apiUrl, originalId, translationId, nonce, type ) {
 		var cachedSuggestion = getTheSuggestionFromTheCache( type, originalId );
 		if ( cachedSuggestion ) {
+			$container.removeClass( 'fetching' );
+			$container.find( '.suggestions__loading-indicator' ).remove();
 			if ( isThisTypeOfSuggestionInTheContainer( $container, type ) ) {
 				return;
 			}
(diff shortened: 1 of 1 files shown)


trymebytes (commented):
I tested and it works as expected. When I jumped from row 1 to say row 5, I found out that it sent 8 AJAX requests, 4 for the row 5 I opened and another for the row 6 I have yet to open. If I decide not to open row 6 and jump to row 8, the same 8 AJAX requests would be made.

I was wondering if we can and need to reduce the occurrence of these multiple requests.

amieiro:
> I tested and it works as expected. When I jumped from row 1 to say row 5, I found out that it sent 8 AJAX requests, 4 for the row 5 I opened and another for the row 6 I have yet to open. If I decide not to open row 6 and jump to row 8, the same 8 AJAX requests would be made.
> 
> I was wondering if we can and need to reduce the occurrence of these multiple requests.

**This is the expected work**. I only preload the first row when the page is loaded, with the closed table, because I expect the translator will start translating in the first row. And when the translator opens a row, I preload the next one, because I expected the translator will continue translating the next row.

I know the user can jump between alternate rows, and the requests will work fine, but without preloading. I don't want to preload all the rows, because this will consume a lot of API tokens, so I only preload the row that is most likely to be opened.

trymebytes (approved):

akirk (changes requested):
Oops, this was for the other PR. I still need to test this one and update here.

akirk (approved):
This works well for me.

amieiro:
Released [here](https://github.com/WordPress/wordpress.org/commit/95bd5ada1360e9be77712d7f1e2855d80b7fb43c).

PR #142, opened 2023-06-06
Plugins: Add a filter bar to switch between business models

ryelle:
This adds "Commercial" and "Community" (and "All") links to the top of the plugins grid. When clicked, it filters the current archive by the selected business model.

This taxonomy already existed as "Business model". The filter URLs already work on production: for example, [all community plugins](https://wordpress.org/plugins/?plugin_business_model=community) or [commercial block directory plugins](https://wordpress.org/plugins/browse/block/?plugin_business_model=commercial).

When the archive title is generated, it's "Business Model: Commercial" & "Business Model: Community" — do we want to change that at all?

The filters exist on all archives except the "beta" view, since these are likely all "community" (or eventually "canonical"?). Should they apply to "My Favorites" (see screenshot below)?

The mockups also include a "Canonical" link, which does exist on the site, but there are no plugins in it yet, so I did not add it here.

cc @WordPress/meta-design 

| Before | After |
|---|---|
| before | home |

All community plugins (https://wordpress.org/plugins/?plugin_business_model=community)
filtered

My Favorites — currently has the filters, but we could remove if this seems unnecessary. I removed the curation filters from favorites on [the Pattern Directory](https://github.com/WordPress/pattern-directory/pull/584), but that is a slightly different use case.

favorites

The filters also apply to rosetta sites:

| Arabic | Korean |
|---|---|
| rosetta-ar | rosetta-ko |

Ticket: https://meta.trac.wordpress.org/ticket/7035


wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/header.php
```diff
@@ -19,7 +19,21 @@
...
+$local_nav_items = array(
+	'' => __( 'All', 'wporg-plugins' ),
+	'community' => __( 'Community', 'wporg-plugins' ),
+	'commercial' => __( 'Commercial', 'wporg-plugins' ),
+);
+
+global $wp_query;
+$is_beta = 'beta' === $wp_query->get( 'browse' );
+// The filter bar should not be shown on:
+// - search: wp-query filters not compatible with ES search.
```
StevenDufresne, line 32: This is unfortunate. I would expect it to be most valuable on the search page since browsing the directory isn't a great experience. 

During the redesign, we should make sure to get this working or at least scoped.

StevenDufresne:
> When the archive title is generated, it's "Business Model: Commercial" & "Business Model: Community" — do we want to change that at all?

Maybe we can go with:

- `Commercial Plugins`
- `Community Plugins`

We should also probably update the copy on the filtered view before launch. They have pluralization problems:

`This plugin is developed and supported by a community.` should be `These plugins are developed and supported by a community.`
| Screenshot |
| --- |
| Screenshot 2023-06-07 at 10 49 16 AM |

StevenDufresne:
> The filters exist on all archives except the "beta" view, since these are likely all "community" (or eventually "canonical"?). Should they apply to "My Favorites" (see screenshot below)?

I don't think it's necessary for that view.

wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/header.php
```diff
@@ -47,16 +61,50 @@
...
+						( ! $slug && false === strpos( $_SERVER['REQUEST_URI'], 'plugin_business_model' ) ) // phpcs:ignore
+					) {
+						$class = 'is-active';
+					}
+					$url = home_url( '/' );
+					if ( $slug ) {
+						$url = add_query_arg( array( 'plugin_business_model' => $slug ) );
+					} else if ( ! is_home() ) {
+						$url = remove_query_arg( 'plugin_business_model' );
+					}
```
dd32, line 100: This should be achievable just by using `get_term_link()` I'd have thought? We probably want to enable pretty-permalinks for these anyway, and/or use `/browse/community/`.

I'm happy to see the design/functionality launch as-is **before** that happens though, it can be resolved afterwards.

jasmussen:
Looks good, thanks for working on this.

wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/header.php
ryelle: I went with this so that we could easily build up filtered queries like `https://wordpress.org/plugins/browse/block/?plugin_business_model=commercial`, because I was under the impression that these are filters, not just static links (@jasmussen can confirm).

wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/header.php
ryelle: It is _possible_ to add it, assuming the value is synced to the elasticsearch database (i'm not sure where to confirm this), it would need to be added in before the ES query is built ([here, I think](https://github.com/WordPress/wordpress.org/blob/0860b124ae32e0b65129abca6787383bfde4f270/wordpress.org/public_html/wp-content/plugins/plugin-directory/libs/site-search/jetpack-search.php#L444-L457)).

So we can add the filter to search too, on the redesign.

ryelle:
I've removed the filters from the "Favorites" page, and updated the title & description on the archive pages, so those now look like this:

| Community | Commercial |
|---|---|
| Screenshot 2023-06-07 at 4 53 47 PM | Screenshot 2023-06-07 at 4 54 08 PM |

wordpress.org/public_html/wp-content/themes/pub/wporg-plugins/header.php
StevenDufresne: Sounds good.

ryelle:
Committed in [r12643](https://meta.trac.wordpress.org/changeset/12643).

PR #144, opened 2023-06-08
Save stats to database for website quality tracking.

StevenDufresne:
This PR, although not in the right place (will move it elsewhere) adds a rest endpoint so we can save quality metrics to the database.

### DIFF
--- a/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-quality-stats.php
+++ b/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-quality-stats.php
@@ -1,0 +1,132 @@
+<?php
+
+namespace WordPressdotorg\MU_Plugins\REST_API\Site_Quality_Stats;
+
+use WP_Error;
+use WP_REST_Controller, WP_REST_Server, WP_REST_Response;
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-quality-stats.php
```diff
@@ -0,0 +1,132 @@
...
+			);
+		}
+
+		$results = json_decode( urldecode( $body ), true );
+
+		foreach ( $results as $item ) {
+			foreach ( $item['summary'] as $key => $value ) {
+				$result = $this->save_stat( $item['url'], $key, (int) ( $value * 100 ) );
+
+				//TO DO: Where should I log failures?
```
dd32, line 124: If you need to log errors (And not just let it silently error) you could use a `trigger_error( E_USER_WARNING )`.

StevenDufresne: Yeah, Maybe i'll let it fail silently since it's supplementary data. Thanks.

PR #145, opened 2023-06-08
Add playground embedder

jdevalk:
Fixes https://meta.trac.wordpress.org/ticket/7040


wordpress.org/public_html/wp-content/plugins/playground-embedder/playground-embedder.php
```diff
@@ -0,0 +1,64 @@
...
+			],
+			$attributes
+		);
+		$query_parameters = array_filter( $query_parameters, static function( $element ) {
+			return ! empty( $element );
+		} );
+
+		$url = add_query_arg( $query_parameters, 'https://playground.wordpress.net/' );
+
+		return sprintf( '<iframe src="%1$s" width="800" height="600"></iframe>', $url );
```
adamwoodnz: The width and height attributes should likely be pulled from the shortcode attributes, but this could be addressed in a later PR.

adamwoodnz: Also $url should probably be escaped, I've tested and `esc_url( $url )` works as expected

PR #146, opened 2023-06-08
Add GTM mu-plugin to buddypress

jdevalk:


dd32:
Closing this, as the MU-plugin is already within the environment, just loading only for the `profiles.wordpress.org` hostname.

Another BuddyPress.org/bbPress.org plugin is in use which loads GTM tags for those sites.

I'll enable them as-needed once my questions in Slack are answered.

PR #148, opened 2023-06-08
Theme Directory: Update packages & build process.

ryelle:
This updates the Gruntfile, to switch from node-sass (deprecated) to sass, removes the pixrem polyfill, and removes the es6-promise polyfill. It also updates all the dependencies to their latest versions.

See 00fd9cd87b7a71eea98ddf9b5c4ff0b8101fce39.

To test:

- Run `npm install` in the wporg-themes folder
- Run `npm run grunt build`
- The CSS files `css/style.css`, `css/style-rtl.css`, and `css/style.css.map` should be generated
- The JS file `js/theme.min.js` should be generated

There should be no functionality changes on the site.

PR #149, opened 2023-06-08
Theme Directory: Add a filter bar to switch between business models

ryelle:
See https://github.com/WordPress/wordpress.org/pull/142 for the Plugin Directory version of this PR.

This updates the navigation on the Theme Directory to add "Commercial" and "Community" links, which show archives of themes using that categorization. It creates new "browse" pages, `wordpress.org/themes/browse/commercial/` and `wordpress.org/themes/browse/community/`.

⚠️  Due to the way the theme's JS routing works, I can't figure out a great way to combine filters (ex, we can't sort "Community" by popular, nor can we filter out "Community" + "Block Themes"). But this matches the current behavior anyway - you can't sort Block Themes by popular, either.

This is built on #148 but I can't seem to chain them since that branch lives on my own fork, but I intended for them to be separate commits.

cc @WordPress/meta-design

**Screenshots:**

I also added some responsive styles so the nav isn't so awkward on smaller screens. it's not great, but should be fine until this is redesigned.

| Large | Medium | Small |
|---|---|---|
| themes-community-large | themes-commercial-med | themes-commercial-small |
| themes-block-theme-large | themes-block-theme-med | themes-block-theme-small |
| themes-filters-open | themes-filters-open-med | themes-filters-open-small |

Ticket: https://meta.trac.wordpress.org/ticket/7036


StevenDufresne:
Can we move the margin to the padding declaration to increase the indicator width?
https://github.com/WordPress/wordpress.org/blob/ab81d3e771247cdb7d8f83bf33697f6c4d3921cd/wordpress.org/public_html/wp-content/themes/pub/wporg-themes/css/components/_main.scss#L1446
https://github.com/WordPress/wordpress.org/blob/ab81d3e771247cdb7d8f83bf33697f6c4d3921cd/wordpress.org/public_html/wp-content/themes/pub/wporg-themes/css/components/_main.scss#L1447

| Before | After |
| --- | --- |
| Screenshot 2023-06-09 at 10 22 10 AM | Screenshot 2023-06-09 at 10 21 58 AM |

StevenDufresne:
Is it possible my patch didn't apply correctly? It seems to load the relevant view on `page load` but clicking the menu items returns empty from the endpoint?

Screen Capture on 2023-06-09 at 10-29-29

dd32:
> Is it possible my patch didn't apply correctly?

@StevenDufresne Make sure you have the API sandboxed, but there's also caching which you may have run into.  
If you look for `if ( 'local' === wp_get_environment_type() ) {` in wporg-themes/functions.php and force-enable that branch it should bypass caching.

dd32:
I'm personally not a fan of replacing the `Latest` and `Popular` filters with the dropdown and treating it like a Sort - Because they're not sorts, they're views. (The code calls it a 'sorter' but that's a misnomer, and relates to the original code that was pulled from wp-admin theme browser)

If we want to convert them to sorts instead, then that requires changing them from `browse` to being `orderby`s in the query handlings.

dd32 (approved):
Other than my above concerns over the views being changed to sorts, this looks good to me and works in my testing.

jasmussen:
Ideally the filter bar could visually be similar to that of the pattern and plugin directory "tab-bars", i.e. just a blue background on the active state. The theme directory here is a bit the odd one out, with a bordered box, white background, and bottom border.

But given these extend sections are all due for a more coherent visual refresh in the future, this is more of a future concern, where I would expect filter-bars across to be unified, and lean closer to the pattern the plugin and pattern directories use. So on these, I'll defer to you on the dev side to choose. Functionally, this looks right. Thank you!

ryelle:
> I'm personally not a fan of replacing the Latest and Popular filters with the dropdown and treating it like a Sort - Because they're not sorts, they're views. (The code calls it a 'sorter' but that's a misnomer, and relates to the original code that was pulled from wp-admin theme browser)

Yeah, I was halfway to updating that (sorter term) when I realized it was a bigger change, and I was aiming for the smallest change to ship the new filters. (that's the same reason I didn't adjust the active state look, @jasmussen)

Instead of the dropdown implying sort, we could keep the tab look, which would clear up that confusion:

Screenshot 2023-06-09 at 12 28 41 PM

What do you think, @jasmussen ?

ryelle:
I updated the spacing on links, which also fixes the slightly awkward focus state 👍🏻 

| Before | After |
|---|---|
| Screenshot 2023-06-09 at 5 17 01 PM | Screenshot 2023-06-09 at 5 08 05 PM |

> Is it possible my patch didn't apply correctly? It seems to load the relevant view on page load but clicking the menu items returns empty from the endpoint?

The PR has changes to both `/themes/pub/wporg-themes/` and `/plugins/theme-directory/`— I'm not sure how you apply patches but I need to sync those separately, and if you miss the plugin, the endpoint doesn't work.

jasmussen:
Thanks for the clarity. The tabbed approach seems like the most pragmatic approach. I personally think that can be okay to go with, so for speed that seems fine to me.

I would love a general sanity check by @richtabor and @fcoveram, probably doesn't need to block anything here, mostly for followup.

ryelle:
Updated to use tabs instead of the sort dropdown, and committed in [r12645](https://meta.trac.wordpress.org/changeset/12645) and [r12646](https://meta.trac.wordpress.org/changeset/12646).

PR #150, opened 2023-06-09
Translate: Add Warnings for Special Originals (i.e. quasi-Settings) in WordPress Core

akirk:
There are some originals in WordPress core that need to be translated in special ways, see @zodiac1978's [gist](https://gist.github.com/Zodiac1978/1d1bd363d5ba7d1ea68e3ffe352008f1).

A way to solve this would be to add specialized warnings in GlotPress which are implemted in this PR, pending some changes in GlotPress core https://github.com/GlotPress/GlotPress/pull/1636.
Screenshot 2023-06-09 at 12 50 55


akirk:
Can be closed after https://github.com/WordPress/wordpress.org/pull/162 is shipped

PR #152, opened 2023-06-12
Update event list UI: Reorder event information and display event type

### LINKED ISSUE #894: Enhancement of Event List UI and Redirection Update for Events Domain
To help with the promotion of Pilot Events, we need to add them to the [Community Events](https://href.li/?https://make.wordpress.org/community/events/) list. The usability and user experience of the Community Event Lists can also be enhanced by this action.

- [x] Redirect the URL `events.wordpress.org` to https://make.wordpress.org/community/events/. This will simplify access to the community events list for users.

- [x] Improve the event list user interface to make it easier for users to identify the type of each event. Currently, the events list's layout can be difficult to read and needs some adjustments.

**Potential Option for UI Renewal**
| Before | After |
| ------ | ------|
| image | image |

Any further thoughts or suggestions on either the UI or the redirection?

outdoor2kode:
See https://github.com/WordPress/wordcamp.org/issues/894

This PR aims to improve the event list user interface by rearranging the information for each event and displaying the event type prominently. The changes include:

This update is being made in order to support the [The Next Generation of WordCamps](https://make.wordpress.org/community/2023/05/08/the-next-generation-of-wordcamps/) pilot event program. More traffic will be directed to this page (`community/events`) and clearly communicating the event type has been identified as a needed improvement.

- Moving the time information to the first line.
- Keeping the event name in the second line.
- Placing the location information in the third line.
- Introducing the event type and putting it before the location for better identification.

## Screenshots

| Before | After1 | After2 |
|   --       | --       |        -- |
| image | Screen Shot 2023-06-13 at 2 36 33 AM | Screen Shot 2023-06-13 at 2 36 20 AM |

## Note

1. `mb_convert_case` was used because even though we don't seem to translate this page into any other languages, I guess we will in the future and this function will fit the case while `ucwords` won't. Also, I assume `type` in the database is specified as `UTF-8`.

2. ~I'm not sure if it's worth doing (or best to do) more on our current layout. Perhaps adding some labels before each piece of information, like `Time: 2:00 p.m.`, or adding a table for it, could be beneficial? What do you all think? And Is now a good time to bring the meta design into this discussion, or someone else should be tagged instead?~ ✅ 

## Testing Steps

Go to https://make.wordpress.org/community/events/ while sandboxed and check if the layout is the same as the screenshots above.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-events-online.js
+++ b/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-events-online.js
@@ -12,6 +12,7 @@ function getSortedEvents( events ) {
 		// Inject a human-readable start time.
 		event.startTime = wp.date.format( 'g:i a', d );
 		event.timezone = timezone;
+		event.type = event?.type === 'wordcamp' ? 'WordCamp' : toTitleCase(event.type);
 		if ( sortedEvents.hasOwnProperty( key ) ) {
 			sortedEvents[ key ].push( event );
 		} else {
@@ -19,6 +20,17 @@ function getSortedEvents( events ) {
 		}
 	} );
 	return sortedEvents;
+}
+
+/**
+ * Convert a string to title case.
+ *
+ * @param {string} str
+ */
+function toTitleCase(str = '') {
+    return str.replace(/\w\S*/g, function(word) {
+        return word.charAt(0).toUpperCase() + word.substring(1).toLowerCase();
+    });
 }
 
 jQuery( function( $ ) {
--- a/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
+++ b/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
@@ -9,19 +9,23 @@
 		<ul class="ofe-event-list">
 			<?php foreach ( $day_events as $event ) : ?>
 				<li>
-					<?php if ( $event->location ) : ?>
-						<?php echo esc_html( 'online' === $event->location ? 'Online' : $event->location ); ?>
-						<span class="owe-separator"></span>
+					<?php if ( $event->start_timestamp ) : ?>
+						<?php echo date( 'g:i a', $event->start_timestamp ); ?>
+						<br>
 					<?php endif; ?>
 
 					<a href="<?php echo esc_url( $event->url ); ?>">
 						<?php echo esc_html( $event->title ); ?>
+						<br>
 					</a>
-
-					<?php // WordCamp.org doesn't collect start times, so only display those for meetups ?>
-					<?php if ( 'wordcamp' != $event->type ) : ?>
+					
+					<?php if ( $event->type ) : ?>
+						<?php echo esc_html( 'wordcamp' === $event->type ? 'WordCamp' : mb_convert_case( $event->type, MB_CASE_TITLE, 'UTF-8' ) ); ?>
 						<span class="owe-separator"></span>
-						<?php echo date( 'g:i a', $event->start_timestamp ); ?>
+					<?php endif; ?>
+
+					<?php if ( $event->location ) : ?>
+						<?php echo esc_html( 'online' === $event->location ? 'Online' : $event->location ); ?>
 					<?php endif; ?>
 				</li>
 			<?php endforeach; ?>
--- a/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-events-online.php
+++ b/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-events-online.php
@@ -20,7 +20,7 @@ function get_synced_online_events() {
 	// Include yesterday's events because server timezone may be ahead of user's timezone.
 	// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
 	$raw_events = $wpdb->get_results(
-		"SELECT title, url, date_utc, date_utc_offset
+		"SELECT title, url, date_utc, date_utc_offset, type
 		FROM `{$table}`
 		WHERE
 			date_utc >= SUBDATE( CURRENT_DATE(), 1 ) AND
@@ -43,6 +43,7 @@ function get_synced_online_events() {
 			'title'           => $event->title,
 			'url'             => $event->url,
 			'start_timestamp' => $timestamp,
+			'type'            => $event->type,
 		);
 	}
 
--- a/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events-online.php
+++ b/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events-online.php
@@ -10,11 +10,13 @@
 
 <script type="text/template" id="tmpl-official-event">
 	<li>
+		{{ data.startTime }}
+		<span class="owe-timezone">{{ data.timezone }}</span>
+		<br>
 		<a href="{{ data.url }}">
 			{{ data.title }}
 		</a>
-		<span class="owe-separator"></span>
-		{{ data.startTime }}
-		<span class="owe-timezone">{{ data.timezone }}</span>
+		<br>
+		{{ data.type }}
 	</li>
 </script>


wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
```diff
@@ -9,20 +9,26 @@
...
 					<a href="<?php echo esc_url( $event->url ); ?>">
 						<?php echo esc_html( $event->title ); ?>
+						<br>
 					</a>
-
-					<?php // WordCamp.org doesn't collect start times, so only display those for meetups ?>
-					<?php if ( 'wordcamp' != $event->type ) : ?>
+					
+					<?php if ( $event->type ) : ?>
+						<?php echo esc_html( mb_convert_case( $event->type, MB_CASE_TITLE, "UTF-8") ); ?>
```
iandunn: `WordCamp` is technically spelled with a capital `C`, so we may need something here that converts the types to hardcoded labels.

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
```diff
@@ -9,20 +9,26 @@
 		<ul class="ofe-event-list">
```
iandunn: I'm guessing we should also apply these changes to `template-events-online.php` (which generates https://make.wordpress.org/community/events/online/)

iandunn:
> Is now a good time to bring the meta design into this discussion

That seems like a good idea, they might have ideas about the labels, etc 👍🏻 

@WordPress/meta-design

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
```diff
@@ -9,20 +9,26 @@
...
 			<?php foreach ( $day_events as $event ) : ?>
 				<li>
-					<?php if ( $event->location ) : ?>
-						<?php echo esc_html( 'online' === $event->location ? 'Online' : $event->location ); ?>
-						<span class="owe-separator"></span>
+					<?php // WordCamp.org doesn't collect start times, so only display those for meetups ?>
+						<?php if ( 'wordcamp' != $event->type ) : ?>
+							<?php echo date( 'g:i a', $event->start_timestamp ); ?>
```
iandunn: What do you think about testing for the presence of a valid timestamp, rather than the event type? We don't know yet if pilot events will provide a start time or not, so that'd be a bit more future proof.

iandunn (commented):
This is looking good, but may need a few tweaks before merge.

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
outdoor2kode: I'm wondering is this page still in use? I didn't seem able to find a link redirecting to it, and also found that in `/events`, online events were already included. Perhaps there's a post/thread discussing about this, I'll see if I can find it.

jasmussen:
Seems like a fine micro improvement to what exists.

Provided you'd be up for potentially new tweaks at some point in the future, at some point we likely need to refresh these sections holistically to the new design. So it seems fine to go with an iterative improvement for now. What do you think?

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
outdoor2kode: Updated in 39127b32d7b047fd7eb0f2d77149f5f207a2d4e2

image

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
outdoor2kode: I'm curious as to why we didn't use `$event->timestamp` as the checking condition in the first place. Do you happen to know the history of that?

As I noticed that a WordCamp also has a time display, like in this example:
image
But the time seems to be [incorrect](https://allevents.in/iloilo/wordcamp-iloilo-2023/200024454111504) in this case, which could possibly be one of the reasons why we initially didn't want to display the time for WordCamp?

There are also cases where the time display is [accurate](https://montclair.wordcamp.org/2023/schedule/), so it might just have been overlooked:
image

I think we should still be able to test for the presence of a valid timestamp and display the WordCamp time + other events time.

outdoor2kode:
> So it seems fine to go with an iterative improvement for now. What do you think?

Thanks for jumping in. It makes sense to refresh the sections holistically. Regarding a micro improvement that doesn't require too much additional time, I agree that this is good to go 👍

jasmussen:
Also just to clarify, by using the term "micro improvement" I didn't mean to belittle your work 😅 — on the contrary, I appreciate the constant iteration. Thank you.

outdoor2kode:
Ah, I totally get it. Perhaps I should put it this way: "Given your feedback, I think there's no need to spend too much additional time atm, and a micro improvement would be good enough to go" Sorry for the confusion and much appreciated for the clarification 🙂

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
outdoor2kode: Updated in d6e9deb40384c12f298243596fadaded2c4d2e56.

jasmussen:
No confusion, all appreciation 🚀

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
outdoor2kode: I think I may understand how this works now; my previous understanding was slightly off. Here it's just defining the shortcode `official_wordpress_events_online`. I see that it's used on the page `events/online`. Although I still can't find a link that can link to this `events/online` page, I did notice that some posts had mentioned this page in their content. So, I guess this template still needs to be updated.

outdoor2kode: Updated in 840d20b61a2669d86b864e09559e24148016bde3

image

outdoor2kode: `Type` hasn't been added yet as I've found that adding it doesn't seem to provide much help currently, as shown below, it's all `meetups`. However, there have been instances of online WordCamp in the past few years, and from what I understand, upcoming `pilot events` are also very likely to be held online? If that's the case, then the `type` should also be added.

meetup

outdoor2kode: Note: While modifying the `events/online` page, I ran into the following issue which is the best to be resolved in another PR. I'll document it here first. There are some pictures and contexts that can aid in the explanation. If there are any thoughts, feel free to comment. If not, I'll open a ticket in Trac later for future fix.

There seems to be some discrepancy between the time displayed on the `events/online` page and the `events` page. On the `events`, it displays the local time of the event's location, sorted from earliest to latest. 

But on the `events/online`, the time is converted to the user's local time, with the timezone added afterward, such as UTC+8 that I am seeing. And in this case, it is no longer sorted from earliest to latest.

One possible simple improvement I can think of is to always display the time in the event's local time, along with that time zone. This could also retain the sort order from earliest to latest.

iandunn: > If that's the case, then the type should also be added.

Yeah, I think so. There could be some WordCamps in the future that choose to do an online event too, so IMO it's best if the template can handle it.

> On the events, it displays the local time of the event's location... But on the events/online, the time is converted to the user's local time

IIRC that was intentional, since the user's time is what matters most to online viewers. For in-person events, attendees would need to know the venue time. The sorting might be a bug though.

@ryelle , does that seem right to you?

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
iandunn: 🤔 I wonder if it was originally done that way because we didn't always publish the times for WordCamps? We started doing that a few years ago:

https://github.com/WordPress/wordcamp.org/blob/f2ae12927dd4c79aefb9a65efaeb34f66f3c2f61/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php#L402-L404

https://github.com/WordPress/wordpress.org/blob/15b2da987d3e8735fcf13ee7c3972051e628e00c/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-wordpress-events.php#L328

The inaccurate one might be a bug that we could fix in another ticket.

StevenDufresne:
@renintw I think it's worth adding the context for this PR since it's separated from its driver ([WordCamp pilot event MVP](https://github.com/WordPress/wordcamp.org/milestone/12)). 

This update is being made in order to support the [The Next Generation of WordCamps](https://make.wordpress.org/community/2023/05/08/the-next-generation-of-wordcamps/) pilot event program. More traffic will be directed to this page and clearly communicating the event type has been identified as a needed improvement.

outdoor2kode:
> @renintw I think it's worth adding the context for this PR since it's separated from its driver ([WordCamp pilot event MVP](https://github.com/WordPress/wordcamp.org/milestone/12)).
> 
> This update is being made in order to support the [The Next Generation of WordCamps](https://make.wordpress.org/community/2023/05/08/the-next-generation-of-wordcamps/) pilot event program. More traffic will be directed to this page and clearly communicating the event type has been identified as a needed improvement.

Thanks. I've added it to the PR description.

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
outdoor2kode: Updated in 8e8f067f817d095d41460228ca8fd54a4c8ab42f.

Case conversion can also be done [after this line](https://github.com/WordPress/wordpress.org/blob/trunk/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-events-online.php#L45) with something like `'type' => 'wordcamp' === $event->type ? 'WordCamp' : mb_convert_case( $event->type, MB_CASE_TITLE, "UTF-8" )`, but I assume there might be a different idea about how to use the value elsewhere someday.

ryelle: > IIRC that was intentional, since the user's time is what matters most to online viewers. For in-person events, attendees would need to know the venue time.

Correct about the timezone.

> The sorting might be a bug though.

Yeah… the sorting is a bug… `getSortedEvents` assumes the original list is presorted, because it seems like it is— but it's sorted by the `date_utc` value in the database, which as [this comment notes,](https://github.com/WordPress/wordpress.org/blob/376bdd6e4866a8600a757d89496c0fc5bc97f096/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-events-online.php#L39-L40) is actually not UTC, but the local timezone. So something at 5pm in London (16 UTC) would appear after something at 4pm in New York (20 UTC), despite the London event happening first.

Actually sorting the events in `getSortedEvents` should fix that.

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-events-online.js
```diff
@@ -12,6 +12,7 @@ function getSortedEvents( events ) {
 		// Inject a human-readable start time.
 		event.startTime = wp.date.format( 'g:i a', d );
 		event.timezone = timezone;
+		event.type = event?.type === 'wordcamp' ? 'WordCamp' : toTitleCase(event.type);
```
iandunn: What's the browser support on `?.` ? This file isn't transpiled so we'll need to make sure it meets https://make.wordpress.org/core/handbook/best-practices/browser-support/

renintw: yeah, I believe it supports the browsers required on the list.

**Filter browsers with a global usage > 1%**
image


**List created by the rule in [browser support](https://make.wordpress.org/core/handbook/best-practices/browser-support/)**
and_chr 114
android 114
chrome 114
chrome 113
chrome 112
chrome 109
edge 114
edge 113
edge 112
firefox 113
firefox 112
ios_saf 16.5
ios_saf 16.4
ios_saf 16.3
ios_saf 16.1
opera 99
opera 98
safari 16.5
safari 16.4
samsung 20

outdoor2kode: yeah, I believe it supports the browsers required on the list.

**Filter browsers with a global usage > 1%**
image

**List created by the rule from [browser support](https://make.wordpress.org/core/handbook/best-practices/browser-support/)**
and_chr 114
android 114
chrome 114
chrome 113
chrome 112
chrome 109
edge 114
edge 113
edge 112
firefox 113
firefox 112
ios_saf 16.5
ios_saf 16.4
ios_saf 16.3
ios_saf 16.1
opera 99
opera 98
safari 16.5
safari 16.4
samsung 20

iandunn: 👍🏻 thanks for checking!

pkevan:
Not fully relevant to this PR, but how is the caching supposed to work here: https://github.com/WordPress/wordpress.org/pull/152/files#diff-040d01da7ce85154ff3542bb0faaaa35dfeca9941c16446fe1701fa30fe92c26R37-R50? Am I missing something?

wordpress.org/public_html/wp-content/plugins/official-wordpress-events/template-events.php
outdoor2kode: Trac ticket: https://meta.trac.wordpress.org/ticket/7071
PR opened: https://github.com/WordPress/wordpress.org/pull/158

outdoor2kode:
> Not fully relevant to this PR, but how is the caching supposed to work here: https://github.com/WordPress/wordpress.org/pull/152/files#diff-040d01da7ce85154ff3542bb0faaaa35dfeca9941c16446fe1701fa30fe92c26R37-R50? Am I missing something?

I think there might be a misuse of the term here. I'm guessing it might have meant to indicate that it had loaded some cached stuff, but they actually come from the DB. Not so sure about the real intent here in the first place, though.
Same stuff happens [here](https://github.com/WordPress/wordpress.org/blob/515685680447c7870fde77e3d0202b64226979ab/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-wordpress-events.php#L232).

If there are no further thoughts on this guess, I'll open a ticket later to address this minor issue.

outdoor2kode:
As per the feedback [from the community](https://wordpress.slack.com/archives/C08M59V3P/p1686844308287899?thread_ts=1686794407.319169&cid=C08M59V3P), I believe it's also worth adding event type to WP Dashboard event widget. Any concerns about doing so? @StevenDufresne @naokomc

iandunn:
> I'm guessing it might have meant to indicate that it had loaded some cached stuff, but they actually come from the DB

Yeah, the database is acting as a local cache of the external APIs. It's not the canonical source, it's just a copy for performance.

IMO that's a correct use of the term, but I also agree that it might be worth a comment somewhere to clarify that it's not a _memory_ cache 👍🏻

outdoor2kode:
> the database is acting as a local cache of the external APIs. It's not the canonical source

Ah, I see now, thanks for clarifying. I also agree that it's worth leaving a comment to explain this is not a mem cache 👍

naokomc:
> it's also worth adding event type to WP Dashboard event widget

If there's room for the label, I agree it'd be good additional information for the audience.

PR #155, opened 2023-06-13
Update cookies page

jdevalk:
Reflect GA4 usage across all sites.


StevenDufresne:
Taking a look at the cookies, I see `__gsas` which is google AdSense. Is that related? Should we update it in this Pull Request? 

Otherwise this change looks right. 👍

StevenDufresne:
Merged in https://github.com/WordPress/wordpress.org/commit/caf9d36e647382c6ea83e7cc6444a0a23a98c5e7.

PR #156, opened 2023-06-14
Add settings to enable or disable ChatGPT review

trymebytes:
In this PR we add a checkbox to the [translate settings page](https://translate.wordpress.org/settings/) to allow users enable or disable chatgpt reviews.

image


akirk:
Now that we have https://github.com/GlotPress/gp-translation-helpers/pull/174, could we instead of adding an extra setting just use the localStorage detail open state to decide whether to load it?

trymebytes:
Sounds good! This suggested approach will ensure that the details element for ChatGPT review is still displayed rather than be outrightly hidden.

PR #158, opened 2023-06-15
Official WordPress Events: Sort events by UTC time

outdoor2kode:
Fixes https://meta.trac.wordpress.org/ticket/7071.

This PR proposes a code change that sorts all events according to their UTC time on [the online events list page](https://make.wordpress.org/community/events/online/), to achieve a globally unified event order.

## Screenshot

| Before | After |
| -------| ----- |
| Screen Shot 2023-06-15 at 11 07 59 PM | Screen Shot 2023-06-15 at 11 34 03 PM  |

## Testing Steps

1. Apply the changes.
2. Go to https://make.wordpress.org/community/events/online/ and see if the order is correct.

## Note

We could also fix it by adding `events.sort((a, b) => a.start_timestamp - b.start_timestamp);` to [getSortedEvents](https://github.com/WordPress/wordpress.org/blob/376bdd6e4866a8600a757d89496c0fc5bc97f096/wordpress.org/public_html/wp-content/plugins/official-wordpress-events/official-events-online.js#L7), but I assume an events array sorted by UTC would be convenient for most cases.


outdoor2kode:
Gonna deploy the change.

* The function is only used within the same file ✅ .

PR #163, opened 2023-07-03
Remove extra AJAX call and replace with AJAX prefilter

trymebytes:
In this PR we remove the extra AJAX call that saves the external translations used and replace that with AJAX prefilter which modifies the payload to include external translations used.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-plugin.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-plugin.php
@@ -195,26 +195,24 @@ class Plugin {
 	 * @param object $translation Created translation.
 	 */
 	public function update_external_translations( $translation ) {
-		if ( 'GP_Route_Translation' === GP::$current_route->class_name ) {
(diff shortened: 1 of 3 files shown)


wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-plugin.php
```diff
@@ -187,4 +188,32 @@ public function extend_translation_suggestions( $entry ) {
...
 		<?php
 	}
+
+	/**
+	 * Update the number of external translations used.
+	 *
+	 * @param object $translation Created translation.
+	 */
+	public function update_external_translations( $translation ) {
+		if ( 'GP_Route_Translation' === GP::$current_route->class_name && 'translations_post' === GP::$current_route->last_method_called ) {
```
akirk: Just an idea to keep the levels of indentation down:
```suggestion
		if ( 'GP_Route_Translation' !== GP::$current_route->class_name || 'translations_post' !== GP::$current_route->last_method_called ) return;
```

trymebytes:
Closing this in favour of #164

PR #164, opened 2023-07-12
Log the translation suggestion usage

trymebytes:
In this PR;
- we refactor the code that logs the usage of translation suggestions. We also removed the extra AJAX call in the `wporg-gp-translation-suggestions` plugin.
- We log translations whose source is translation memory. These are stored as `tm`.
- This improvements will allow easy logging of new external suggestions source that may be added in the future.

### How to test
- Go to the translations page of a any project and choose a string to translate.
1. If you enter a translation or paste a translation in the text area below, the source is logged as `frontend`
Screenshot 2023-07-17 at 22 46 17

2. If you click to copy a suggested translation e.g openai as shown below, two rows(one to log the translation source and the second to log the suggestions_used) are inserted in the `translate_meta` table, one that logs the translation source as `frontend` and the second that logs the `suggestion_used` as openai. 
This applies to clicking to copy translation suggestions from translation memory( `suggestion_used`is logged  as `tm`) and deepl ( `suggestion_used`is logged  as `deepl`).
Screenshot 2023-07-17 at 22 46 36

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/class-plugin.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/class-plugin.php
@@ -67,6 +67,7 @@ class Plugin {
 		add_action( 'gp_translation_created', array( $this, 'auto_reject_replaced_suggestions' ) );
 		add_action( 'gp_translation_created', array( $this, 'log_translation_source' ) );
 		add_action( 'gp_translation_saved', array( $this, 'log_translation_source' ) );
+		add_action( 'gp_translations_imported', array( $this, 'log_imported_translations' ) );
 
 		add_filter( 'gp_for_translation_clauses', array( $this, 'allow_searching_for_no_author_translations' ), 10, 3 );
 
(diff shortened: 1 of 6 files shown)


wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/class-plugin.php
```diff
@@ -349,7 +344,7 @@ private function save_translation_suggestion_source( $translation, $suggested_tr
 			$suggestion_used .= '_modified';
 		}
 		if ( $suggestion_used ) {
-			gp_update_meta( $translation->id, 'suggestion-used', $suggestion_used, 'translation' );
+			gp_update_meta( $translation->id, 'suggestionused', $suggestion_used, 'translation' );
```
akirk: Why that change? I think `suggestion-used` is fine. Or maybe just `suggestion`.

akirk (approved):
Tested and works well (I tested openai and openai_modified). I'd prefer to keep it `suggestion-used` or just `suggestion`. When shipping this make sure to test it and to use the latest GlotPress release where the JS is minified.

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/class-plugin.php
trymebytes: We resolved to using `_` (`suggestion_used`)because the function `gp_sanitize_meta_key` removes `-`.

akirk: Ah I didn't know!

PR #165, opened 2023-07-14
Translate: Fix initialization error in logging translation source

trymebytes:
When an import is done and we try to log the source, we get the error below;

`E_ERROR: Uncaught Error: Typed property WordPressdotorg\GlotPress\Customizations\Plugin::$imported_source must not be accessed before initialization in /home/wporg/public_html/wp-content/plugins/wporg-gp-customizations/inc/class-plugin.php:364`

In this PR we set the variable to an empty string to initalize it before it's used.


akirk (approved):
Looks good to me. The variable could be accessed uninitialized when a PO file is imported that doesn't have any (new) translations.

trymebytes:
Thank you!

PR #167, opened 2023-07-24
Log locale slug as suggestion source when suggestions from other languages are used

trymebytes:
When suggestions from other languages are clicked, the system logs them as `undefined` or `undefined_modified` if they are modified. 
Screenshot 2023-07-24 at 12 26 39

We fix this by logging the suggestion source as the slug of the locale e.g `es`, `de`, etc. When they are modified it's logged as `es_modified`, `de_modified`

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/js/translation-suggestions.js
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/js/translation-suggestions.js
@@ -478,9 +478,13 @@
 		if ( ! $row ) {
 			return;
 		}
-		externalSuggestion.suggestion_source = $row.data( 'suggestion-source' ) == 'translation' ? 'tm' : $row.data( 'suggestion-source' );
-		externalSuggestion.translation = $row.find( '.translation-suggestion__translation' ).text();
-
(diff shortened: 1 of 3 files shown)


akirk (approved):
In my testing this worked well, nice job!

amieiro (approved):
It works fine for me!

wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/js/translation-suggestions.js
```diff
@@ -478,16 +478,18 @@
...
+			externalSuggestion.translation = $row.find( '.translation-suggestion__translation' ).text();
+		}
 	}
 
 	//Prefilter ajax requests to add external translations used to the request.
 	$.ajaxPrefilter( function ( options ) {
-		const isSuggestionUsed = Object.keys( externalSuggestion ).length  > 0 ? true : false;
-
-		if ( ! externalSuggestion || ! isSuggestionUsed ) {
+		if ( ! externalSuggestion.suggestion_source || ! externalSuggestion.translation ) {
```
akirk: I think we should not remove the ` ! externalSuggestion` check.
```suggestion
		if ( ! externalSuggestion || ! externalSuggestion.suggestion_source || ! externalSuggestion.translation ) {
```

PR #175, opened 2023-08-21
Add redirect for short Gutenberg phase 3 url

adamwoodnz:
Closes https://meta.trac.wordpress.org/ticket/7223


dd32 (approved):
Looks good to me:

```
$ curl -Is https://wordpress.org/phase-3 https://make.wordpress.org/phase-3  | grep -E '^(HTTP|Location)'
HTTP/1.1 404 Not Found
HTTP/1.1 301 Moved Permanently
Location: /core/tag/phase-3/
```
(First is a 404 on the wordpress.org url, second is a redirect on the make url)

PR #177, opened 2023-10-12
Update wp-63.php for 6.3.2 credits

audrasjb:
Trac ticket: https://meta.trac.wordpress.org/ticket/7302


coffee2code:
Fixed via the Trac ticket.

PR #180, opened 2023-11-24
Add a CLI command and a cronjob to detect duplicated translations

amieiro:
## Problem

Last week, while I was developing an improvement to be able to make [bulk updates to the translations in the consistency tool](https://make.wordpress.org/polyglots/2023/05/15/bulk-updates-in-the-translation-consistency/), I made a mistake and set [some translations as "current"](https://translate.wordpress.org/projects/meta/wordcamp/gl/default/?filters[status]=either&filters[original_id]=4085022) for the same original.

image

I knew we have some translations in the same situation, maybe related with some translations created while the original strings are imported.

I made a script to detect the number of duplicates, and we have more than 42,000 duplicates in 31,600,000 translation sets (around 0.13 % of the translations).

We have different situations:

1. Same `date_added` and `date_modified`. We have rows, at least, between 2015 and in 2023:

```
SELECT * FROM `translate_translations` WHERE `original_id` = 89329 AND `translation_set_id` = 11930
```

image

```
SELECT * FROM `translate_translations` WHERE `original_id` = 14652687 AND `translation_set_id` = 4345261
```

image

2. Same `date_modified` but different `date_added`:

```
SELECT * FROM `translate_translations` WHERE `original_id` = 10345042 AND `translation_set_id` = 367588
```

image

3. Rows with multiple duplicates:

```
SELECT * FROM `translate_translations` WHERE `original_id` = 1654345 AND `translation_set_id` = 1814851
``` 

image

4. Translations created by a user and then updated by a script: user_id_last_modified = 0.

```
SELECT * FROM `translate_translations` WHERE `original_id` = 3750671 AND `translation_set_id` = 6579081
``` 

image

## Solution 

This PR adds:
- A **cronjob**, to review daily if we have some duplicated translations in the database with the `current` status. It will send a notification to a Slack channel.
- A **WP-CLI command**, to manually update the duplicates.

To **detect duplicated translations**, the code looks for translations with the same `original_id` and `translation_set_id` and the translation status set to `current`.

To **update these duplicated translations**, the code updates all the duplicates, setting:
- The `status` to `old`.
- The `user_id_last_modified` to `NULL`, reflecting a translation updated by a bot.
- The `date_modified` to the `current date time`.

Only one duplicated translation remains without been updated:
- The last modified (highest `date_modified` value).
- If two or more duplicated translations have the same `date_modified` value, the code selects the one with the highest `id` (the last introduced in the system).

### Cronjob

The cronjob checks once a day if we have some duplicated translations in the database and sends a notification to the #dotorg channel (Slack, "Making WordPress" instance). 

These are the screenshots that this cronjob will send:

image

image

### Command

The command will allow updating the duplicated translations, so we will be able to solve this problem by hand. If this problem persists, we will be able to update these duplicated translations in the cronjob (This has to be done in a different PR).

To use this command, you have to execute in the CLI:

```
wp wporg-translate duplicate-translations --url=translate.wordpress.org
```
This command has 4 flags:
- `--fix`: Updates the duplicates, setting all translations to old except the last one.
- `--notify`: Notifies the duplicates in Slack.
- `--verbose`: Shows the results in the CLI.
- `--print-sql`: Shows the SQL queries to show and to update the duplicate entries.

Some CLI examples:

```
wp wporg-translate duplicate-translations --url=translate.wordpress.org
wp wporg-translate duplicate-translations --url=translate.wordpress.org --fix
 wp wporg-translate duplicate-translations --url=translate.wordpress.org --notify
wp wporg-translate duplicate-translations --url=translate.wordpress.org --update --verbose --print-sql
```

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
@@ -250,7 +250,6 @@ class Detect_Duplicate_Translations {
 				// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 				echo $update_query . "\n";
 			}
-			break; // todo: remove this break to update all the duplicates.
 		}
 
 		if ( $this->verbose ) {
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
@@ -1,0 +1,287 @@
+<?php
+/**
+ * This class detects duplicate translations in the database in "current" status and updates them.
+ *
+ * @package WordPressdotorg\GlotPress\Customizations\CLI
+ */
+
+namespace WordPressdotorg\GlotPress\Customizations\CLI;
+
+/**
+ * Class Detect_Duplicate_Translations
+ */
+class Detect_Duplicate_Translations {
+
+	/**
+	 * The lower limit for translation_set_id.
+	 *
+	 * @var int
+	 */
+	private int $low_limit = 0;
+
+	/**
+	 * The upper limit for translation_set_id.
+	 *
+	 * Note: It must be greater than the highest translation set id available in the database.
+	 *
+	 * @var int
+	 */
+	private int $high_limit;
+
+	/**
+	 * The step size for iteration.
+	 *
+	 * With steps of 10,000 we sometimes have timeouts in the database.
+	 *
+	 * @var int
+	 */
+	private int $step = 1000;
+
+	/**
+	 * The duplicated entries.
+	 *
+	 * @var array
+	 */
+	private array $duplicates = array();
+
+	/**
+	 * If true, show the results in the CLI.
+	 *
+	 * @var bool
+	 */
+	private bool $verbose = false;
+
+	/**
+	 * If true, show the SQL queries to show and to update the duplicate entries.
+	 *
+	 * @var bool
+	 */
+	private bool $print_sql = false;
+
+	/**
+	 * Here we store the last MySQL error.
+	 *
+	 * @var string
+	 */
+	private string $last_sql_error = '';
+
+	/**
+	 * Detect_Duplicate_Translations constructor.
+	 *
+	 * @param bool $update_values If true, update the duplicate entries.
+	 * @param bool $notify        If true, notify the duplicate entries in Slack.
+	 * @param bool $verbose       If true, show the results in the CLI.
+	 * @param bool $print_sql     If true, show the SQL queries to show and to update the duplicate entries.
+	 *
+	 * @return void
+	 */
+	public function __invoke( bool $update_values = false, bool $notify = false, bool $verbose = false, bool $print_sql = false ) {
+		global $wpdb;
+		$this->verbose    = $verbose;
+		$this->print_sql  = $print_sql;
+		$this->high_limit = $this->get_high_limit();
+		$this->duplicates = $this->get_duplicates();
+		if ( $update_values ) {
+			$this->update_duplicates();
+		}
+		if ( $notify ) {
+			$this->notify_duplicates();
+		}
+	}
+
+	/**
+	 * Get the highest translation_set_id in the translations table.
+	 *
+	 * Round up the highest translation_set_id to the next multiple of $this->step.
+	 *
+	 * @return int
+	 */
+	private function get_high_limit() {
+		global $wpdb;
+		$max_translation_set_id = $wpdb->get_var( "SELECT MAX(translation_set_id) FROM {$wpdb->gp_translations}" );
+		if ( $this->verbose ) {
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			echo 'Highest translation_set_id: ' . number_format_i18n( $max_translation_set_id ) . "\n";
+		}
+
+		if ( $wpdb->last_error ) {
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			echo 'Error in the SQL query: ' . $wpdb->last_error;
+			$this->last_sql_error = $wpdb->last_error;
+		}
+
+		return ceil( $max_translation_set_id / $this->step ) * $this->step;
+	}
+
+	/**
+	 * Get the duplicate entries from the translations table.
+	 *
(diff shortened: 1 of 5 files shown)


wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
```diff
@@ -0,0 +1,286 @@
...
+			",
+				$duplicate['original_id'],
+				$duplicate['translation_set_id'],
+				$duplicate['original_id'],
+				$duplicate['translation_set_id'],
+				$duplicate['original_id'],
+				$duplicate['translation_set_id'],
+				$duplicate['original_id'],
+				$duplicate['translation_set_id']
+			);
```
akirk: Maybe I am misinterpreting what the code does but wouldn't this be much simpler?
```suggestion
foreach ( $this->duplicates as $duplicate ) {
	$id = $wpdb->get_var(
		$wpdb->prepare(
			"SELECT id
			FROM {$wpdb->gp_translations}
			WHERE original_id = %d
			AND translation_set_id = %d
			AND status = 'current'
			ORDER BY date_modified DESC, id DESC
			LIMIT 1",
			$duplicate['original_id'],
			$duplicate['translation_set_id']
		)
	);

	$update_query = $wpdb->prepare(
		"UPDATE {$wpdb->gp_translations}
			SET status = CASE
				WHEN id = %d THEN 'current'
				ELSE 'old'
			END,
			user_id_last_modified = CASE
				WHEN id = %d THEN user_id_last_modified
				ELSE NULL
			END,
			date_modified = CASE
				WHEN id = %d THEN date_modified
				ELSE NOW()
			END
			WHERE original_id = %d
			AND translation_set_id = %d
			AND status = 'current';
		",
		$id,
		$id,
		$id
		$duplicate['original_id'],
		$duplicate['translation_set_id']
	);
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
```diff
@@ -0,0 +1,286 @@
...
+			$mysql_error = "The last MySQL error was: " . $this->last_sql_error . "\n";
+		}
+
+		if ( 0 == count( $this->duplicates ) ) {
+			$message = ":white_check_mark: There are no duplicate translations in the database with *current* status.\n";
+			$message .= $mysql_error;
+		} else {
+			$message  = ":warning: There are " . count( $this->duplicates ) . " duplicate translations in the database with *current* status.\n\n";
+			$message .= "Execute *'wp wporg-translate detect-duplicate-translations --url=translate.wordpress.org --verbose'* to show the duplicate entries.\n";
+			$message .= "Execute *'wp wporg-translate detect-duplicate-translations --url=translate.wordpress.org --update-values'* to update the duplicate entries, solving the problem.\n";
```
akirk: It irks me a bit wrong that the `detect-duplicate-translations` modifies data with the right parameter. What if we renamed it to `duplicate-translations` and the `verbose` parameter to `detect` and the `update-values` to `fix`?
```suggestion
			$message .= "Execute *'wp wporg-translate duplicate-translations --url=translate.wordpress.org --detect'* to show the duplicate entries.\n";
			$message .= "Execute *'wp wporg-translate duplicate-translations --url=translate.wordpress.org --fix'* to update the duplicate entries, solving the problem.\n";
```

akirk (commented):
I like the approach, I think better naming would make it easier to comprehend

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
akirk: Or even just:
```suggestion
foreach ( $this->duplicates as $duplicate ) {
	$id = $wpdb->get_var(
		$wpdb->prepare(
			"SELECT id
			FROM {$wpdb->gp_translations}
			WHERE original_id = %d
			AND translation_set_id = %d
			AND status = 'current'
			ORDER BY date_modified DESC, id DESC
			LIMIT 1",
			$duplicate['original_id'],
			$duplicate['translation_set_id']
		)
	);

	$update_query = $wpdb->prepare(
		"UPDATE {$wpdb->gp_translations}
			SET status = 'old',
			user_id_last_modified = NULL,
			date_modified = NOW()
			WHERE original_id = %d
			AND translation_set_id = %d
			AND id != %d
			AND status = 'current';
		",
		$duplicate['original_id'],
		$duplicate['translation_set_id'],
		$id
	);
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
amieiro: I changed the command name and a parameter name:
- Command name: detect-duplicate-translations -> duplicate-translations
- Parameter name: --update -> --fix

By default, the command, without any flag, will detect the duplicated values. The "--verbose" flag will show some important values, like the number of duplicates, and the "--print-sql" flag will show the SQL queries to show and to update the duplicate entries.

wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-detect-duplicate-translations.php
amieiro: I have updated the code with this improvement!

amieiro:
@akirk I have made the changes you have suggested in the WP-CLI command interface and the improvement to the update SQL query.

akirk (approved):
Looks great!

amieiro:
I close this PR, as it has been deployed. Finally, [the notifications are sent to Matrix](https://github.com/WordPress/wordpress.org/blob/a6c5e84dd79ea16b3f77739d6d93904505042a8b/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/inc/cli/class-duplicate-translations.php#L244-L276).

NekoJonez:
@amieiro , why set them on old right away? What if the wrong one is set to old. Maybe set them fuzzy so PTE/GTE's can decide ;) 
That's safer and will avoid mistakes and bugs. Otherwise one duplicate can cause issues.

amieiro:
@NekoJonez We had a lot of translations with duplications: around 42,000, so if we put this as fuzzy, the work would be very big for the GTE.
We had a lot of old duplications, some from 2015, and most of them had the same text, so I decided to put them as old, because I considered this was not a risk.
In future executions, with small sets of duplicated translations, the script will set them as fuzzy instead of old, and the GTE will review them with low effort.

PR #187, opened 2023-12-21
Adjust 6.4 credits to avoid an extra noteworthy contributor on the last line

sabernhardt:
Removes me from the 'contributing-developers' section to make the total evenly divisible by four. However, I should still be in the list of text links further down.

**Before:** each row can have up to four names (and images) until the last row.

final two rows: one row with four images and names, the last row with one name and image


sabernhardt:
This is not very valuable anymore, so I'll close the PR. But if someone thinks this is still worth changing for the 6.4 credits page, I am okay with removing my name from the top section.

PR #188, opened 2024-01-14
Markdown Importer: Allow iframe elements

adamwoodnz:
Enables embedding of YouTube videos

The [Block Editor handbook quickstart guide](https://developer.wordpress.org/block-editor/getting-started/quick-start-guide/) has a YouTube embed. This is being stripped on import, [see Gutenberg PR for markdown change](https://github.com/WordPress/gutenberg/pull/57834).

Or perhaps there is a better way to embed this content via WPTV or another service?

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-importer.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-markdown/inc/class-importer.php
@@ -499,7 +499,7 @@ abstract class Importer {
 
 		$html = apply_filters( 'wporg_markdown_after_transform', $html, $this->get_post_type() );
 
-		add_filter( 'wp_kses_allowed_html', [ $this, 'wp_kses_allow_links' ], 10, 2 );
+		add_filter( 'wp_kses_allowed_html', [ $this, 'wp_kses_allowed_html' ], 10, 2 );
 
 		$post_data = array(
 			'ID'           => $post_id,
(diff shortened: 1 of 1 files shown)


dd32:
An iframe without validation on the `src` is a pretty big security risk.

Personally, I would not go down this route, but I would instead convert the embedded iframe back to a plain-url or an embed block instead.

dd32:
[Jetpack includes a YT embed => shortcode](https://github.com/Automattic/jetpack/blob/f4ecc331742bb69a2de4b4f2766f5a92ed8aeaac/projects/plugins/jetpack/modules/shortcodes/youtube.php#L17-L109), is that maybe doing it's thing but the result of that is being stripped out too?

adamwoodnz:
> An iframe without validation on the `src` is a pretty big security risk.

Yeah I thought as much.

> [Jetpack includes a YT embed => shortcode](https://github.com/Automattic/jetpack/blob/f4ecc331742bb69a2de4b4f2766f5a92ed8aeaac/projects/plugins/jetpack/modules/shortcodes/youtube.php#L17-L109), is that maybe doing it's thing but the result of that is being stripped out too?

This looks great, I'll give it a shot.

adamwoodnz:
> [Jetpack includes a YT embed => shortcode](https://github.com/Automattic/jetpack/blob/f4ecc331742bb69a2de4b4f2766f5a92ed8aeaac/projects/plugins/jetpack/modules/shortcodes/youtube.php#L17-L109), is that maybe doing it's thing but the result of that is being stripped out too?

So this isn't running at the moment, there's a Jetpack Writing setting which needs to be turned on for embed shortcodes. 

Screenshot 2024-01-15 at 3 37 23 PM

It appears to work locally with the markup you've added @ndiego, converting the iframe embed to a shortcode, and then back to an iframe embed on the frontend. The padding you've added to enlarge the video doesn't work due to a nested span, but we could add styles for this:

Screenshot 2024-01-15 at 3 40 42 PM

Turning on the setting seems like a step forward though, thoughts?

PR #190, opened 2024-01-18
Expand readme warnings

dd32:
Trac: https://meta.trac.wordpress.org/ticket/7412

Downstream issue: https://github.com/wordpress/plugin-check/issues/198

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
@@ -144,7 +144,7 @@ class Parser {
 	 *
 	 * @var array
 	 */
-	private $ignore_tags = array(
+	public $ignore_tags = array(
 		'plugin',
 		'wordpress',
 	);
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
@@ -225,6 +225,8 @@ class Parser {
 
 		// Handle readme's which do `=== Plugin Name ===\nMy SuperAwesomePlugin Name\n...`
 		if ( 'plugin name' == strtolower( $this->name ) ) {
+			$this->warnings['invalid_plugin_name_header'] = true;
+
 			$this->name = $line = $this->get_first_nonwhitespace( $contents );
 
 			// Ensure that the line read wasn't an actual header or description.
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
@@ -423,7 +423,7 @@ class Parser {
 		$this->short_description = $this->sanitize_text( $this->short_description );
 		$this->short_description = $this->parse_markdown( $this->short_description );
 		$this->short_description = wp_strip_all_tags( $this->short_description );
-		$short_description = $this->trim_length( $this->short_description, 150 );
+		$short_description       = $this->trim_length( $this->short_description, 150 );
 		if ( $short_description !== $this->short_description ) {
 			if ( empty( $this->warnings['no_short_description_present'] ) ) {
 				$this->warnings['trimmed_short_description'] = true;
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
@@ -425,7 +425,9 @@ class Parser {
 		$this->short_description = wp_strip_all_tags( $this->short_description );
 		$short_description = $this->trim_length( $this->short_description, 150 );
 		if ( $short_description !== $this->short_description ) {
-			$this->warnings['trimmed_short_description'] = true;
+			if ( empty( $this->warnings['no_short_description_present'] ) ) {
+				$this->warnings['trimmed_short_description'] = true;
+			}
 			$this->short_description = $short_description;
 		}
 
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
@@ -273,8 +273,14 @@ class Parser {
 			$this->tags = explode( ',', $headers['tags'] );
 			$this->tags = array_map( 'trim', $this->tags );
 			$this->tags = array_filter( $this->tags );
-			$this->tags = array_diff( $this->tags, $this->ignore_tags );
-			$this->tags = array_slice( $this->tags, 0, 5 );
+			if ( array_intersect( $this->tags, $this->ignore_tags ) ) {
+				$this->tags = array_diff( $this->tags, $this->ignore_tags );
+				$this->warnings['ignored_tags'] = true;
+			}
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
```diff
@@ -408,13 +416,18 @@ protected function parse_readme_contents( $contents ) {
 		// Use the first line of the description for the short description if not provided.
 		if ( ! $this->short_description && ! empty( $this->sections['description'] ) ) {
 			$this->short_description = array_filter( explode( "\n", $this->sections['description'] ) )[0];
+			$this->warnings['no_short_description_present'] = true;
```
ernilambar, line 419: Since we pull fallback short description from main description, warning for `no_short_description_present` is not very useful I guess.

dd32: That's true, I don't personally see it as something that should be flagged, but it's up to the client code as to whether to warn on it or not.

wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
```diff
@@ -408,13 +416,18 @@ protected function parse_readme_contents( $contents ) {
...
 		}
 
 		// Sanitize and trim the short_description to match requirements.
 		$this->short_description = $this->sanitize_text( $this->short_description );
 		$this->short_description = $this->parse_markdown( $this->short_description );
 		$this->short_description = wp_strip_all_tags( $this->short_description );
-		$this->short_description = $this->trim_length( $this->short_description, 150 );
+		$short_description = $this->trim_length( $this->short_description, 150 );
```
ernilambar: At this point `$this->short_description` already fallback to `$this->sections['description']`. So we are still not sure what is the actual length of the original short description.

dd32: Resolving that is not worth the effort.

Instead, it should be checked something like this instead then:
```php
isset( $this->warnings['trimmed_short_description'] ) && ! isset( $this->warnings['no_short_description_present'] )
```

wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
dd32: Resolving this, as this is actually useful, see: https://github.com/WordPress/wordpress.org/pull/190/files#r1457051069

wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
dd32: Alternatively, we could avoid setting it, based on if a short description was provided, see below suggestion.

wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
```diff
@@ -408,13 +416,18 @@ protected function parse_readme_contents( $contents ) {
...
+		if ( $short_description !== $this->short_description ) {
+			$this->warnings['trimmed_short_description'] = true;
```
dd32:
```suggestion
			if ( empty( $this->warnings['no_short_description_present'] ) ) {
				$this->warnings['trimmed_short_description'] = true;
			}
```

ernilambar: I tested with this suggestion and it is working as expected. If there is warning `trimmed_short_description` then we can flag it for short description exceeding 150 characters.

wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
```diff
@@ -273,8 +275,14 @@ protected function parse_readme_contents( $contents ) {
 			$this->tags = explode( ',', $headers['tags'] );
 			$this->tags = array_map( 'trim', $this->tags );
 			$this->tags = array_filter( $this->tags );
-			$this->tags = array_diff( $this->tags, $this->ignore_tags );
-			$this->tags = array_slice( $this->tags, 0, 5 );
+			if ( array_intersect( $this->tags, $this->ignore_tags ) ) {
+				$this->tags = array_diff( $this->tags, $this->ignore_tags );
+				$this->warnings['ignored_tags'] = true;
```
ernilambar, line 280: I noticed this `ignored_tags` is added here. Nice addition. Is it possible to expose the list of ignored tags so that we could show relevant message if those tags are added? Currently `$ignore_tags` is `private`.

dd32: Changing `$ignore_tags` to public shouldn't be an issue.

wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
```diff
@@ -273,8 +275,14 @@ protected function parse_readme_contents( $contents ) {
...
+			}
+			if ( count( $this->tags ) > 6 ) {
```
ernilambar: I guess it should be `5` instead of `6` here.

dd32: 6 is correct, since it's actually 0 indexed.  It's limited to 6 tags (0, 1, 2, 3, 4, 5 indexes)

ernilambar: But this will never give more than 5 tags. `array_slice( $this->tags, 0, 5 )` Do we allow 5 or 6 tags?

mukeshpanchal27: The code is designed to handle a scenario where there are more than 6 elements in the `$this->tags` array. If this condition is met, it trims the array to only include the first 5 elements and sets a warning flag `too_many_tags` to indicate that there were originally too many tags in the array. Hope that make sense now?

dd32: Yeah.. I misread the code..

We've documented it as being 12 tags everywhere else, and I'm 85% certain this was supposed to be 6 tags.. but 5 is even better.

dd32:
```suggestion
			if ( count( $this->tags ) > 5 ) {
```

wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
```diff
@@ -408,13 +416,20 @@ protected function parse_readme_contents( $contents ) {
...
 			$this->short_description = array_filter( explode( "\n", $this->sections['description'] ) )[0];
+			$this->warnings['no_short_description_present'] = true;
 		}
 
 		// Sanitize and trim the short_description to match requirements.
 		$this->short_description = $this->sanitize_text( $this->short_description );
 		$this->short_description = $this->parse_markdown( $this->short_description );
 		$this->short_description = wp_strip_all_tags( $this->short_description );
-		$this->short_description = $this->trim_length( $this->short_description, 150 );
+		$short_description = $this->trim_length( $this->short_description, 150 );
```
mukeshpanchal27:
```suggestion
		$short_description       = $this->trim_length( $this->short_description, 150 );
```

Nit

PR #197, opened 2024-02-02
Support: Tools: Update packages and build script

ryelle:
This updates the build step for the Support forum theme, as `node-sass` has been deprecated, and the version here does not work on current machines. This PR updates the packages to current versions when available.

**Build** `npm run grunt`

Update the build step order, so that `sass` runs before `postcss`, this way PostCSS runs over CSS, and the `postcss-scss` syntax is not necessary. The changes made in build

**Stylelint** `npm run grunt stylelint`

The update to stylelint & the WP rules also introduces some changes to rule names, so the stylelint config has been updated. I've also disabled a few rules in places to avoid any meaningful changes (like the selector patterns, we can't fix those without also changing the markup). The `sass/**` file changes are all from `npm run grunt stylelint -- --fix`, fixing up whitespace, adding leading `0`s, making hex values lowercase, etc.

**To test**

- Use a version of node >=16 (I was using 20)
- Run `npm install` to update packages
- Run `npm run grunt` to build CSS
- Run `npm run grunt stylelint` to lint the sass

Everything should succeed, and there should be no visual change to the Support Forums.


StevenDufresne:
This is working as expected. Thanks for the PR. I'll commit it.

StevenDufresne:
Changeset: https://meta.trac.wordpress.org/changeset/13167

PR #198, opened 2024-02-06
Have community/commercial filters apply on tags

dd32:
Trac: https://meta.trac.wordpress.org/ticket/7430


dd32:
On URLs such as this: https://wordpress.org/plugins/tags/ecommerce/?plugin_business_model=community the theme 404's.

We currently don't support empty archives, except for searches.

Screenshot 2024-02-06 at 12 22 31 pm

dd32:
The other option would be retaining the `filter` in the redirect from tags => search.

Personally I feel like the only way to make this actually "nice" is to remove the individual "archives" and instead move to a single "search archive with filters".

dd32:
After talking with @StevenDufresne we came to the conclusion that:
 - A request for /tags/something/?model=community should return an empty result-set, and not redirect to search.
 - A request for /tags/no-plugins-in-this-tag/ should redirect to search. (As it currently does)

The above effectively does this, this is a bit more complicated than I'd have liked, as due to the way WordPress handles taxonomy archives, it needed some juggling.

https://github.com/WordPress/wordpress.org/assets/767313/ccdb16ca-daac-4d00-bdee-d092c9b1ee4a

PR #201, opened 2024-02-14
Forums: Homepage redesign

adamwoodnz:
Try implementing the redesign using the current non-block support theme.

Closes https://meta.trac.wordpress.org/ticket/4007

Rearranges the forums homepage content to match the [design](https://www.figma.com/file/3eK690rwPubLvTEvJIJKRO/Forums?type=design&node-id=902-1140&mode=dev) and applies layout and typography changes to match the other 'Learn' sites ([Dev resources](https://developer.wordpress.org/), [Documentation](https://wordpress.org/documentation/)). ~I've stopped short of styling the whole page as I think there's enough work here to judge whether it's a decent approach.~

~The single forum page in the screenshots has had 0 attention, I've just included it to show the different header layout, font and page width changes. Content widths still need working out; we could do as much or as little here as we want as a first iteration.~

Mostly we can use block theme things:

- theme.json settings are output on frontend and vars and helper classes can be used for styling
- we can register block patterns and render with `do_blocks`
- local nav and search blocks work fine, I expect we can use sidebar containers and other things too

Things that need work:

- [x] Better theme.json integration. I've copied in the [wporg-parent-2021 theme.json](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/theme.json), and then made some typography modifications to match the Developer site. This is going to be a maintenance burden. Ideally we would load the parent theme.json (rather than duplicating it) and merge with local overrides like we do for other themes. Not sure how to tackle this.
- [x] Better parent theme styles and block styles integration. Loading from a remote url rather than having this theme be a dependant child isn't ideal. Not sure how to tackle this.
- [x] Remap CSS vars
- [x] Button colors
- [x] Profile restyle
- [x] Forum restyle
- [x] Topic restyle
- [x] Content widths, incl responsive
- [x] Fix 'Also viewing' bar fixed position when scrolled
- [x] Topic table headers and row link styles
- [x] Content spacing

### Testing

Needs to be tested in a sandbox.

General things to check for:
- The old WordPress blue shouldn't be used anywhere except parts of the User Account settings. We need to do work in the 2FA repo to update these
- Fonts should be Inter or EB Garamond. Font sizes have had some updates (headings, paragraphs, sidebar), but there is still lot of bbPress text which will need resizing in another iteration.
- Spacings should be generally inline with the new design, eg. content widths, edge spacing, block spacing

#### How to test

Browse the site checking the following pages and functionality:
- Forums home (check links, responsiveness, a11y)
- Forum archive, eg. https://wordpress.org/support/forum/installation/
- Topic
- All Topics
- Profile and sub pages
- Plugin forum, eg. https://wordpress.org/support/plugin/gutenberg/
- Theme forum, eg. https://wordpress.org/support/theme/twentytwentyfour/
- Check that the new Search field below the local nav behaves the same as in prod on each page.
- Moderation tools (nothing should have changed here)
- Local nav

#### Forums home

| Before | Desktop | Tablet | Mobile |
|-|-|-|-|
| wordpress org_support_forums_(3 1536x864) (3) | wordpress org_support_forums_(3 1536x864) (5) | wordpress org_support_forums_(iPad) (2) | wordpress org_support_forums_(Samsung Galaxy S20 Ultra) (2) |

#### Forum list

| Before | Desktop | Tablet | Mobile |
|-|-|-|-|
| wordpress org_support_forum_installation_(3 1536x864) (2) | wordpress org_support_forum_installation_(3 1536x864) (4) | wordpress org_support_forum_installation_(iPad) (2) | wordpress org_support_forum_installation_(Samsung Galaxy S20 Ultra) (2) |

#### Topic

| Before | Desktop | Tablet | Mobile |
|-|-|-|-|
| wordpress org_support_topic_is-htaccess-working-environment-required-for-wordpress_(3 1536x864) | wordpress org_support_topic_is-htaccess-working-environment-required-for-wordpress_(3 1536x864) (2) | wordpress org_support_topic_is-htaccess-working-environment-required-for-wordpress_(iPad) (1) | wordpress org_support_topic_is-htaccess-working-environment-required-for-wordpress_(Samsung Galaxy S20 Ultra) (1) |

#### Welcome

| Before | Desktop | Tablet | Mobile |
|-|-|-|-|
| wordpress org_support_welcome_(3 1536x864) | wordpress org_support_welcome_(3 1536x864) (1) | wordpress org_support_welcome_(iPad) | wordpress org_support_welcome_(Samsung Galaxy S20 Ultra) |

#### Profile

| Before | Desktop | Tablet | Mobile |
|-|-|-|-|
| wordpress org_support_users_adamwood_(3 1536x864) | wordpress org_support_users_adamwood_(3 1536x864) (1) | wordpress org_support_users_adamwood_(iPad) | wordpress org_support_users_adamwood_(Samsung Galaxy S20 Ultra) |

#### Account

| Before | Desktop | Tablet | Mobile |
|-|-|-|-|
| wordpress org_support_users_adamwood_edit_account_(3 1536x864) | wordpress org_support_users_adamwood_edit_account_(3 1536x864) (1) | wordpress org_support_users_adamwood_edit_account_(iPad) | wordpress org_support_users_adamwood_edit_account_(Samsung Galaxy S20 Ultra) |


wordpress.org/public_html/wp-content/themes/pub/wporg-support/functions.php
```diff
@@ -11,6 +11,36 @@
...
+	);
+}
+add_filter( 'wporg_block_navigation_menus', '\add_site_navigation_menus' );
+
+function register_patterns() {
+	require get_template_directory() . '/patterns/local-nav.php';
+	require get_template_directory() . '/patterns/forums-homepage-header.php';
+	require get_template_directory() . '/patterns/search-field.php';
+}
+add_action( 'init', 'register_patterns' );
```
adamwoodnz, line 48: We can manually register patterns, there might be a better way of doing this.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/functions.php
```diff
@@ -39,6 +69,9 @@ function wporg_support_scripts() {
 	wp_enqueue_style( 'forum-wp4-style', get_stylesheet_uri(), [ 'dashicons' ], filemtime( __DIR__ . '/style.css' ) );
 	wp_style_add_data( 'forum-wp4-style', 'rtl', 'replace' );
 
+	wp_enqueue_style( 'wporg-parent-2021-style', 'https://developer.wordpress.org/wp-content/themes/wporg-parent-2021/build/style.css', [ 'wporg-global-fonts' ] );
+	wp_enqueue_style( 'wporg-parent-2021-block-styles', 'https://developer.wordpress.org/wp-content/themes/wporg-parent-2021/build/block-styles.css' );
```
adamwoodnz: Need a better way of enqueueing these. This isn't a child theme, so it's not a dependency like in other themes.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/blocks/_group.scss
```diff
@@ -0,0 +1,14 @@
...
+
+	&.is-layout-grid.is-style-cards-grid {
+		grid-template-columns: repeat(auto-fill, minmax(min(32.3%, 100%), 1fr));
+		gap: var(--wp--preset--spacing--10);
+	}
+
+	.wp-block-wporg-link-wrapper {
+
+		#bbpress-forums & {
+			border: 1px solid var(--wp--preset--color--light-grey-1);
```
adamwoodnz, line 6: There are a few nasty specificity overrides required

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/typography/_headings.scss
```diff
@@ -1,31 +1,17 @@
...
+h6 {
 	clear: both;
-
-	// Override wp4.css.
-	font-family: inherit;
-	line-height: $font__line-height-body;
-	margin: 2rem 0 1rem;
-}
-
-h1 {
```
adamwoodnz, line 13: These heading styles come from the parent block theme now

wordpress.org/public_html/wp-content/themes/pub/wporg-support/archive-forum.php
```diff
@@ -16,6 +16,23 @@
 
 	</main>
 
+	<div class="forum-home-footer has-text-color has-white-color has-background has-charcoal-2-background-color has-small-font-size">
```
adamwoodnz: Needs more block markup for padding and columns

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/blocks/_group.scss
```diff
@@ -0,0 +1,14 @@
+.wp-block-group {
+
+	&.is-layout-grid.is-style-cards-grid {
+		grid-template-columns: repeat(auto-fill, minmax(min(32.3%, 100%), 1fr));
```
adamwoodnz: Should be able to use block markup for this actually, see https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/patterns/front-page-content.php#L17

StevenDufresne:
Looking at this PR, I'm actually pretty happy with how it looks and that we should be able to use the block themes. I trust your instinct on this one. @ryelle has also converted existing themes to block themes and could advise.

If we do think this is worth it. It may be best to simply ship what you have here (polished) instead of trying to get all the forums and such styled which would introduce more risk into this PR. 

If we can flip it to a block theme but not have any content changes that's probably even best case. We could then work on the individual pages or flows and ship them when ready.

I didn't apply this patch, so make sure to also check:
- Plugin/Theme forums -> https://wordpress.org/support/plugin/gutenberg/
- Forum profiles -> https://wordpress.org/support/users/{username}/
- That the forums still function properly.
  - Adding a new topic
  - Moderating functions

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/blocks/_group.scss
StevenDufresne: Override existing theme styles, bbpress styles or global legacy styles from `wp4.css`? At any rate, a non-issue for me.

ndiego:
This is looking great! The only thing I am noticing is that the search is missing on the homepage. I know that work is not done, but just wanted to note that.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/blocks/_group.scss
adamwoodnz: In this case bbpress, but there are wp4 things to work around too

wordpress.org/public_html/wp-content/themes/pub/wporg-support/theme.json
adamwoodnz, line 1: This is copied from wporg-parent-2021 and then modified with mostly the same changes as we have in the Dev Resources theme.json. I wonder if we could have a build task that fetches the original parent `theme.json`, merges it with a `theme-support.json` and saves the output as this `theme.json` file.

ryelle: Do you mean something that replicates the current child theme process (where we have a parent theme.json merged with a child theme.json)? Does it need to be a build task, or do you think there could be some code (like in the parent theme) to basically require `../../wporg-parent-2021/theme.json` & treat it like the parent theme?

adamwoodnz: I thought that behaviour of the child theme.json being merged with the parent was a default WP parent/child thing, which we wouldn't get if we didn't have a child theme. If we already have something built that does this that's perfect, I'll take a look.

adamwoodnz: https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/functions.php#L144

ryelle: yep! ^ The parent/child merge is default core behavior, but not everything is merged as expected (see https://github.com/WordPress/gutenberg/issues/40557), so that function handpicks a few values to override it. If you use something similar, you'll probably want to merge everything.

You could also look at how [the rosetta overrides work](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/inc/rosetta-styles.php), that one fakes a "user style" (as if it were set in the Site Editor) which overrides the current theme's theme.json by core. It might not help, but it's another example of this kind of manipulation.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/variables-site/_colors.scss
adamwoodnz, line 1: Should just replace the usages of these vars, the mapping is an unnecessary abstraction

wordpress.org/public_html/wp-content/themes/pub/wporg-support/functions.php
ryelle: Hm… I thought patterns should be loaded from the `patterns` directory in classic themes just like they are in block themes. I tried it (I'm sure you did too), but it's not working here, I have no idea why. Core should be doing exactly what you're doing here.

Even when I get around the cached pattern issue, the pattern appears to register, but does not show up where it should.

For dev-ex, I wonder if we should load all patterns in a single file (for example, [how twentynineteen does it](https://core.trac.wordpress.org/browser/trunk/src/wp-content/themes/twentynineteen/inc/block-patterns.php)), since the patterns dir isn't working as expected? (though ideally we could fix whatever's wrong there…)

wordpress.org/public_html/wp-content/themes/pub/wporg-support/functions.php
```diff
@@ -39,6 +70,9 @@ function wporg_support_scripts() {
 	wp_enqueue_style( 'forum-wp4-style', get_stylesheet_uri(), [ 'dashicons' ], filemtime( __DIR__ . '/style.css' ) );
 	wp_style_add_data( 'forum-wp4-style', 'rtl', 'replace' );
 
+	wp_enqueue_style( 'wporg-parent-2021-style', 'https://developer.wordpress.org/wp-content/themes/wporg-parent-2021/build/style.css', [ 'wporg-global-fonts' ] );
```
ryelle:
```suggestion
	wp_enqueue_style( 'wporg-parent-2021-style', get_theme_root_uri() . '/wporg-parent-2021/build/style.css', [ 'wporg-global-fonts' ] );
```

This would at least not rely on a separate site URL.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/blocks/_group.scss
ryelle: I tried dequeueing wp4 (`wp_dequeue_style( 'wp4-styles' );`), it doesn't seem to change much (though it does fix the local nav alignment), so that's also an option. Probably can't dequeue bbpress, I'm sure we need some of that :)

| With `wp4.css` | Without `wp4.css` |
|---|---|
| home-wp4 | home-no-wp4 |
| forum-wp4 | forum-no-wp4 |

ryelle:
This is looking a lot better than I expected for hacking the styles into a classic theme 🙂  Looking good on Rosetta forums too: https://es.wordpress.org/support/forums/

I think this is a good path for this site, especially if you can get the parent theme.json loaded without copying it.

adamwoodnz:
> This is looking a lot better than I expected for hacking the styles into a classic theme 🙂 Looking good on Rosetta forums too: https://es.wordpress.org/support/forums/
> 
> I think this is a good path for this site, especially if you can get the parent theme.json loaded without copying it.

Great! Thanks for the feedback, very helpful 👍

wordpress.org/public_html/wp-content/themes/pub/wporg-support/functions.php
adamwoodnz: Works great, thanks

wordpress.org/public_html/wp-content/themes/pub/wporg-support/theme.json
adamwoodnz: Now getting the parent theme.json and merging with the support theme.json, see [532a03d](https://github.com/WordPress/wordpress.org/pull/201/commits/532a03d33bcb5256daa74a7560b275822ff3487e)

adamwoodnz: I missed the subtlety in the typography fontSizes data shape and why `_merge_by_slug` was important. My recursive merge function needs more work to account for this.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/functions.php
adamwoodnz: Added a [DirectoryIterator](https://github.com/WordPress/wordpress.org/pull/201/commits/a385929554dee7ccecda8a2befa5ed95df958019#diff-2d167406fe0e1ea03098dc0c90a59782d8c8d97096d5173788c1052ac82d85ecR41) to require all the files in `/patterns/`, see above

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/style-original.scss
```diff
@@ -119,7 +119,7 @@ ul#views {
 .bbp-reply-header .bbp-meta .bbp-reply-permalink {
 	float: none;
 	margin-left: 0;
-	color: $color__link;
+	color: #0073aa;
```
adamwoodnz, line 122: Sass color vars have been removed, so reverting this to what I believe would be the original value

wordpress.org/public_html/wp-content/themes/pub/wporg-support/sass/blocks/_group.scss
adamwoodnz: I've now dequeued wp4, works nicely thanks

adamwoodnz:
There is currently a layout issue with the local nav on mobile.



I haven't got to the bottom of it yet, but it might be a conflict with the mu-plugin block and the older version of Gutenberg (we're pinned to 16.8 due to issues with Blocks Everywhere).

adamwoodnz:
@WordPress/meta-design please review the screenshots. This is obviously just a first iteration so there are many things left to address. But let me know if there is anything glaring that you'd like to avoid shipping 🙏

wordpress.org/public_html/wp-content/themes/pub/wporg-support/theme.json
adamwoodnz, line 1: This is based on the [theme.json from Developer Resources](https://github.com/WordPress/wporg-developer/blob/trunk/source/wp-content/themes/wporg-developer-2023/theme.json), with a few removals and additions, for example existing colors.

ndiego:
I'll leave the full review to Design, but this is looking great! I did notice one thing. Currently, the user avatar is centered on the edge of the container, but it's a bit off in the new design. 

image

ryelle:
> I haven't got to the bottom of it yet, but it might be a conflict with the mu-plugin block and the older version of Gutenberg (we're pinned to 16.8 due to issues with Blocks Everywhere).

This is probably it— the dropdown in the local nav has a good amount of customization of the GB modal (to make it more dropdown-y instead of full overlay modal).

wordpress.org/public_html/wp-content/themes/pub/wporg-support/header.php
```diff
@@ -41,36 +41,52 @@
...
-				<?php endif; ?>
-			</div><!-- .site-branding -->
-		</header><!-- #masthead -->
-		<div id="lang-guess-wrap"></div>
+						<!-- /wp:group -->
+
+					</div>
+					<!-- /wp:group -->
+
+					<div id="lang-guess-wrap"></div>
```
ryelle:
```suggestion
					<!-- wp:wporg/language-suggest {"align":"full"} -->
					<div class="wp-block-wporg-language-suggest alignfull"></div>
					<!-- /wp:wporg/language-suggest -->
```

If you swap this, you can remove the locale banner code from wporg-support. Including `_legacy_notice.scss`, which overrides the styles if you just drop this in.

Though, separately, `lang-guess.php` (on dotorg.trac) should be updated to understand `?uri=/support/forums/` as [right now it returns](https://wordpress.org/lang-guess/lang-guess-ajax.php?uri=/support/forums/):

> WordPress is also available in [한국어](https://ko.wordpress.org/), [日本語](https://ja.wordpress.org/), 그리고 [Español](https://es.wordpress.org/).

When it should be 

> WordPress support forums are also available in [한국어](https://ko.wordpress.org/support/), [日本語](https://ja.wordpress.org/support/), 그리고 [Español](https://es.wordpress.org/support/).

(see https://meta.trac.wordpress.org/ticket/6599, https://meta.trac.wordpress.org/ticket/6600)

wordpress.org/public_html/wp-content/plugins/wporg-bbp-also-viewing/wporg-bbp-also-viewing.js
```diff
@@ -11,6 +11,8 @@
 		banner = false,
```
ryelle, line 11: Not a blocker, just a suggestion— Could you create a block for this? maybe with the Interactivity API? (the declarative structure of the interactivity api might not work nicely with the setInterval calls). Rather than checking for `$localNavBar` to split the behavior, the block theme would use the block version, and the jQuery version could be deprecated.

That also might be more trouble than it's worth, and can be revisited if/whenever we fully blockify this :)

wordpress.org/public_html/wp-content/plugins/wporg-bbp-also-viewing/wporg-bbp-also-viewing.js
```diff
@@ -105,17 +121,28 @@
...
-		}
-
-		if ( jQuery(window).scrollTop() > bannerOffset ) {
-			banner.css( 'position', 'fixed' );
-			banner.css( 'top', offset );
+		if ( $localNavBar.length ) {
+			// Local nav bar position is set to relative on small screens to stop it being sticky
+			if ( $localNavBar.hasClass( 'is-sticking' ) && $localNavBar.css( 'position' ) !== 'relative' ) {
+				banner.css( 'position', 'sticky' );
+				banner.css( 'top', 'calc(var(--wp-admin--admin-bar--height, 32px) + var(--wp--custom--local-navigation-bar--spacing--height, 60px))' );
```
ryelle: If the admin bar exists, `--wp-admin--admin-bar--height` should be set. If it's not set, there is no admin bar, so the fallback would be `0px`.

Functionally I think the admin bar always exists on Forums, but if something changes, this would cause misalignment.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/bbpress/loop-forums-homepage.php
```diff
@@ -1,30 +1,44 @@
...
-		<?php endwhile; ?>
-
-
-</div><!-- .forums-directory -->
-
-<div class="themes-plugins">
-	
-	<h3><?php _e( 'Themes and Plugins', 'wporg-forums' ); ?></h3>
+<section class="forums-homepage-list">
+	<h2 class="has-heading-5-font-size">Forums</h2>
```
ryelle: Missing i18n?

wordpress.org/public_html/wp-content/themes/pub/wporg-support/functions.php
```diff
@@ -228,23 +394,22 @@ function wporg_support_get_views() {
 		'all-topics',
 		'no-replies',
 		'support-forum-no',
-		'taggedmodlook',
```
ryelle, line 231: Why is this removed?

wordpress.org/public_html/wp-content/themes/pub/wporg-support/theme.json
```diff
@@ -0,0 +1,193 @@
...
+								"lineHeight": 1.2
+							}
+						}
+					}
+				}
+			},
+			"z-index": {
+				"iso-editor": 100,
+				"local-nav": 101
+			}
```
ryelle, line 109: Great idea to set these here 💯

ryelle:
I clicked through a few of the sidebar links, and some of them show full posts, so the design is a bit awkward. None of these links are public though so I don't want to share any screenshots.

https://wordpress.org/support/view/spam/
https://wordpress.org/support/view/pending/
https://wordpress.org/support/view/archived/
https://wordpress.org/support/view/all-replies/

I wasn't able to get the locale banner to appear on any pages, and the `locale-banner.js` script doesn't seem to be loaded. I left a suggestion about using the language-suggest block instead.

Minor design note: quoted text looks a bit too large, and while I'll defer to design's opinion, I don't think this should be EB Garamond.

Screenshot 2024-03-01 at 3 44 57 PM

wordpress.org/public_html/wp-content/themes/pub/wporg-support/functions.php
adamwoodnz: It isn't in the design, but I guess that's a logged out view. Reverted in [2e5f3f8](https://github.com/WordPress/wordpress.org/pull/201/commits/2e5f3f8f3b14d4b90be8a32acd7d7645d1fb3efb)

wordpress.org/public_html/wp-content/themes/pub/wporg-support/bbpress/loop-forums-homepage.php
adamwoodnz: Nice catch, fixed in [6eb6355](https://github.com/WordPress/wordpress.org/pull/201/commits/6eb63553975b5ac0a2c4a3a0138e7faeb29d6e58)

wordpress.org/public_html/wp-content/plugins/wporg-bbp-also-viewing/wporg-bbp-also-viewing.js
adamwoodnz: Fixed in [b2bbdc2](https://github.com/WordPress/wordpress.org/pull/201/commits/b2bbdc27e19dde5846d642f09f1d7e1528830c74)

wordpress.org/public_html/wp-content/plugins/wporg-bbp-also-viewing/wporg-bbp-also-viewing.js
adamwoodnz: Yeah great idea. Ticket created https://meta.trac.wordpress.org/ticket/7503

adamwoodnz:
> I'll leave the full review to Design, but this is looking great! I did notice one thing. Currently, the user avatar is centered on the edge of the container, but it's a bit off in the new design.
> 
> image

Fixed in [6440483](https://github.com/WordPress/wordpress.org/pull/201/commits/644048322d30a19c0f62a62214b1f15648e4a70f)

wordpress org_support_topic_is-htaccess-working-environment-required-for-wordpress_(3 1536x864) (3)

adamwoodnz:
> I wasn't able to get the locale banner to appear on any pages

Nice catch @ryelle. Turns out [`lang-guess.php`](https://dotorg.trac.wordpress.org/browser/wordpress/website/wp-content/mu-plugins/main-network/lang-guess.php#L12) will only add the inline script for this on these URIs `'wordpress.org/', 'wordpress.org/download/', 'wordpress.org/support/', so it's not working in prod either. As you say, the language doesn't make sense as is, so simply updating `wordpress.org/support/` to `wordpress.org/support/forums/` doesn't solve the whole problem. I'll create a separate ticket to address this.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/header.php
adamwoodnz: Had a look into why this banner isn't working and raised a [ticket](https://dotorg.trac.wordpress.org/ticket/32).

I also discovered that https://de.wordpress.org/support/ is still live, whereas if I try to visit https://wordpress.org/support/ I end up with a web search. I expected a 301, (or worst case a 404), so this behaviour is pretty odd. I haven't covered restyling this page in this PR as I thought it was redirected, based on https://meta.trac.wordpress.org/ticket/6841. Where do you end up @ryelle?

Seems like we need to setup redirects for .org and all subdomains if the intention is for `/support/forums/` to be the homepage from now on.

dd32: @adamwoodnz Can you document that on a meta.trac ticket instead, since the one you linked is private?

StevenDufresne:
Looking good!

Here are some things I found:

## Tables:
Can we take this a bit farther and closer to designs at least visually? Looking at the current table:
- it's too blue. 
  - Not all links need to be blue. 
- The header color is really distracting.
- Checkmarks are underlined

| Underlined checkmarks |
|--------|
| Screenshot 2024-03-04 at 2 37 13 PM | 

| Too blue |
|--------|
| Screenshot 2024-03-04 at 2 36 04 PM | 

## Forum Sidebar links.
The sidebar links shouldn't be underlined? They are a bit distracting when looking at a theme/plugin forum.

Screenshot 2024-03-04 at 2 46 10 PM

## Reviews:
We need to improve the review section. 
- Remove the top border on the reviews on the right
- Increase "Average rating" title size?
Screenshot 2024-03-04 at 2 47 30 PM

## Breadcrumb
The last part is pretty lightly colored. Can we make it darker for better contrast?
Screenshot 2024-03-04 at 2 54 45 PM

The welcome page also has the breadcrumbs and the title directly beside each other. Do we need it here?
Screenshot 2024-03-04 at 2 55 56 PM


## Mobile Menu missing padding:
| No Padding |
|--------|
| Screenshot 2024-03-04 at 2 57 51 PM |

## Logout banner text alignment.

When you log out from a forum, you get redirected to the homepage. The text is not vertically aligned:

Screenshot 2024-03-04 at 2 32 18 PM

## Homepage text alignment.
The spacing between the text makes the content feel separated. I think the second paragraph should come directly after the first. They don't need to be on new lines..

Screenshot 2024-03-04 at 3 00 57 PM

wordpress.org/public_html/wp-content/themes/pub/wporg-support/header.php
adamwoodnz: Copied the details over to https://meta.trac.wordpress.org/ticket/7504

adamwoodnz:
Thanks for testing @StevenDufresne 

> Can we take this a bit farther and closer to designs at least visually? Looking at the current table:
> 
> * it's too blue.
>   
>   * Not all links need to be blue.
> * The header color is really distracting.
> * Checkmarks are underlined

I just went for a straight color swap here, with the intention of doing more in the next iteration. I think you're right that we could change the header color and remove the underlines. I'd like to call out that the underlines are our default so that's why they're showing, and I believe within content areas this is for accessibility.

> The sidebar links shouldn't be underlined? They are a bit distracting when looking at a theme/plugin forum.

Again this is our default style so I haven't tried to deviate from that in our first iteration. But I tend to agree, and I see they aren't underlined in the design.

> The last part is pretty lightly colored. Can we make it darker for better contrast?

> The welcome page also has the breadcrumbs and the title directly beside each other. Do we need it here?

I'm going to leave breadcrumbs til a next cut. This PR is already too large.

I'll fix the last 3 issues (mobile local nav, logout banner and homepage text alignment), nice catches 🙌

StevenDufresne:
> I just went for a straight color swap here, with the intention of doing more in the next iteration. I think you're right that we could change the header color and remove the underlines. I'd like to call out that the underlines are our default so that's why they're showing, and I believe within content areas this is for accessibility.

Keep in mind that these are defaults for the default theme that is applied elsewhere on the site in different contexts. Users of the forums are not used to everything being underlined. Suggesting that they are underlined now because other parts of the sites show links underlined is not an upgrade for everyday users and likely introduces more harm than good.

adamwoodnz:
> Keep in mind that these are defaults for the default theme that is applied elsewhere on the site in different contexts. Users of the forums are not used to everything being underlined. Suggesting that they are underlined now because other parts of the sites show links underlined is not an upgrade for everyday users and likely introduces more harm than good.

Yep understood, and I think this is a good call.

adamwoodnz:
Updated screenshot for forums topics table, following @StevenDufresne's review cc @WordPress/meta-design. Blue header background changed to white, and underlines on table content and sidebar links removed:

jasmussen:
This looks like a fantastic step forward. There are also a lot of details to refine, but I wonder if those aren't best addressed in followups? E.g. just to name a few, these checkboxes could be modernized across:

Screenshot 2024-03-04 at 10 40 41

And this submit button is pretty small:

Screenshot 2024-03-04 at 10 40 45

But it's hard to know where to draw the line. This is so vastly better, more legible, accessible, than what exists, that it's tempting to move forward and follow up with QA fixes. Alternately, is there a way to test this with a query string?

wordpress.org/public_html/wp-content/themes/pub/wporg-support/header.php
ryelle: The Rosetta sites _do_ still have their own `/support` homepages because the `/documentation` site is English only (for now, but it is), and some Rosetta sites have translated documentation.

Unfortunately that means there are a few extra templates that need styling 😩 

- There are two support homepage variations, depending on if the site has set "front page" widgets: [the Japanese site has widgets](https://ja.wordpress.org/support/) and looks more like the old w.org/support, while [the German site doesn't](https://de.wordpress.org/support/) and looks like /support/forums (it shows the same content at de.wordpress.org/support/ & de.wordpress.org/support/forums/).
- Articles: [without a sidebar](https://it.wordpress.org/support/article/come-installare-wordpress/), [with a sidebar](https://es.wordpress.org/support/article/before-you-install/), [second example with sidebar](https://ja.wordpress.org/support/article/new_to_wordpress_-_where_to_start/) — TBH this looks okay-ish as-is, could just use some box-shadow cleanup & fix the sidebar-less alignment.
- Category archives: [one with many items](https://ja.wordpress.org/support/category/getting-started/), and [one with just two](https://es.wordpress.org/support/category/basic-administration/).

> whereas if I try to visit https://wordpress.org/support/ I end up with a web search.

That's strange, it _should_ be 301-ing to /documentation.

adamwoodnz:
> But it's hard to know where to draw the line.

Yeah 100% agree!

> This is so vastly better, more legible, accessible, than what exists, that it's tempting to move forward and follow up with QA fixes.

This is the plan at present. Ship then iterate. I imagine next will be the topic list page, then single topic. These will address the things you've pointed out, but there is a lot more obviously. All the form elements basically need updates.

> Alternately, is there a way to test this with a query string?

I don't think so. The way we normally do this is with a theme switch, but in this case we are modifying the existing theme.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/header.php
adamwoodnz: The Russian version has another template again https://ru.wordpress.org/support/

> Seems like we need to setup redirects for .org and all subdomains if the intention is for `/support/forums/` to be the homepage from now on.

@WordPress/meta-design what is the intention here; do we want a consistent homepage experience for `en` and Rosetta sites like the design shows, or is a homepage at `/support/` still something that should be supported?

ryelle: > The Russian version has another template again 

Looks like this is the same template as the Japanese version, just with different widgets.

> is a homepage at /support/ still something that should be supported?

I thought that `/support/` was going to come back to the main site as well (with https://meta.trac.wordpress.org/ticket/6841)? 

If `/support/` is removed on rosetta, how would locale site visitors get to that documentation? Eventually `/documentation` will include localized content, but there's no proposal or timeline from #docs, so IMO we need to keep that info available here.

adamwoodnz: Makes sense to bring it back to me. I think what is adding to the confusion is that the designs have dropped almost all mention of 'Support', and the link from the global nav is also 'Forums', so I'm not quite sure what a Support landing page should look like. Maybe we need to pause and allow some design time around this problem?

Screenshot 2024-03-05 at 8 55 24 AM

adamwoodnz: I'm really not sure what to do about the widget style homepages, as the content can't be handled with our existing card styles, there are so many links.



As a first step I think I'll convert the header to the new style, and leave the widgets as is. I'd appreciate your thoughts please @WordPress/meta-design

StevenDufresne:
@adamwoodnz You could create a new theme instead of merging directly into wporg-support and setup theme switching. 

We are doing that for plugins as well: 
https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024.

That bbPress .js file may not work when theme-switched but that's not tremendously important.

adamwoodnz:
> Alternately, is there a way to test this with a query string?

@jasmussen I've created a new theme now and we'll have a theme switcher setup tomorrow.

adamwoodnz:
Closing the PR as it has done it's job. The changes have been committed to [a new theme in Trac](https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024) so we can deploy it for preview without activating it.

wordpress.org/public_html/wp-content/themes/pub/wporg-support/header.php
jasmussen: Indeed, we need the new header style rather than the tall blue one. 

I think the content of the page can relatively easily be converted to one of the new patterns we have. I assume it's this page? https://ja.wordpress.org/support/

My first instinct here would be to remove the icons. Secondly, to explore converting that list of links into one of the existing patterns such as here, from [Learn](https://www.figma.com/file/ZKaf6u8mIHkAanluWafXAp/Learn?type=design&node-id=2149%3A5036&mode=design&t=sy7CaBe9i3xQgNph-1):

Screenshot 2024-03-05 at 09 41 42

Whether that's feasible or not, just left aligning the heading and removing the icon could be a start.

adamwoodnz: OK thanks, sounds good. I'll look at expanding our card grid styles to handle cases where the whole card isn't clickable, with nested links.

StevenDufresne:
The subnav obscures the author image and title. [Example](https://wordpress.org/support/topic/change-reset-password-enabled-without-an-old-password/#post-17476937).

Screenshot 2024-03-08 at 11 14 35 AM

ndiego:
This needs some design feedback, but the Themes & Plugin text at the bottom of the page feels out of place and it's hard to find if looking for theme and plugin support. What if we moved it to the main grid, something like this? Maybe it could have a different background color as well?

| Current | Proposed |
|-|-|
|image|image|

@WordPress/meta-design alternative solutions welcomed 🙏

jasmussen:
Looks good insofar as it avoids what can almost be described as a layout-based widow (like a typographic widow). Happy to try this.

adamwoodnz:
> What if we moved it to the main grid, something like this?

I've just deployed this. Note that the number of forums is dynamic, so I've made that card fill the last row:

### Screenshots

| English Forums home | German Support home | Spanish Support home |
|-|-|-|
| Screenshot 2024-03-14 at 4 31 36 PM | Screenshot 2024-03-14 at 4 31 23 PM | Screenshot 2024-03-14 at 4 38 14 PM |

dd32:
> The subnav obscures the author image and title. [Example](https://wordpress.org/support/topic/change-reset-password-enabled-without-an-old-password/#post-17476937).

https://github.com/WordPress/wporg-mu-plugins/issues/586

PR #225, opened 2024-04-06
Trac: Add visually hidden or alternative text for attachment links

sabernhardt:
* "Download" and the file name are added as visually hidden text for raw attachment download links, at least within the comments area. (The "Download" title attribute is removed.)
* The filename is inside its own container for use in the preview links. (The class could be applied to the first link instead, if preferable.)
* The preview images could use the filename as alternative text in all cases, but I favored the user-added description—when available—in this patch.

https://meta.trac.wordpress.org/ticket/6319


obenland:
@sabernhardt Could I ask you to refresh the patch?

joedolson (approved):
This makes sense to me.

wordpress.org/public_html/style/trac/wp-trac.js
```diff
@@ -301,10 +301,18 @@ var wpTrac, coreKeywordList, gardenerKeywordList, hideFromNewTickets, reservedTe
 				if ( href.match(/\.(jpg|jpeg|png|gif|svg|webp)$/i) ) {
 					image = new Image();
 					image.src = href;
+					alt = appendTo.find( '.comment' );
+					if ( alt.length > 0 ) {
+						alt = alt.text();
+					} else {
+						// Use attachment filename if it has no description.
+						alt = el.find( '.trac-attachment-name' ).text();
+					}
```
sabernhardt, line 310: I changed this to check whether the `.comment` finds a match before using `.text()` on it. I also reused the `appendTo` variable.

It was:
```
				alt = li.parent().parent().find( '.comment' ).text();
				if ( alt.length < 1 ) {
					// Use attachment filename if it has no description.
					alt = el.find( '.trac-attachment-name' ).text();
				}
```

wordpress.org/public_html/style/trac/wp-trac.js
```diff
@@ -301,10 +301,18 @@ var wpTrac, coreKeywordList, gardenerKeywordList, hideFromNewTickets, reservedTe
 				if ( href.match(/\.(jpg|jpeg|png|gif|svg|webp)$/i) ) {
 					image = new Image();
 					image.src = href;
+					alt = appendTo.find( '.comment' );
+					if ( alt.length > 0 ) {
+						alt = alt.text();
+					} else {
+						// Use attachment filename if it has no description.
+						alt = el.find( '.trac-attachment-name' ).text();
```
obenland, line 309:
```suggestion
						alt = li.find( 'a[href^="/attachment/"]' ).text();
```

Not sure if there's still a `.trac-attachment-name` element after the Trac update to 1.6. Could you check if this works?

sabernhardt: This PR adds the `.trac-attachment-name` class. (It did not exist before the Trac update either.)

That said, `li.find( 'a[href^="/attachment/"]' ).text()` seems to work as well.

PR #226, opened 2024-04-07
Requirements page: Bump recommended versions of MySQL and MariaDB

t-hamano:
Ticket: https://meta.trac.wordpress.org/ticket/7236

It is planned to update the Rosetta site to the new version in the following issue, and the new version is already enabled in some locales.

https://github.com/WordPress/wporg-main-2022/issues/266

My understanding is that if the new version is enabled, the Requirements page will load the following pattern:

https://github.com/WordPress/wporg-main-2022/blob/trunk/source/wp-content/themes/wporg-main-2022/patterns/about-requirements.php

The recommended version of this **pattern** is output via a shortcode and has been updated by the PRs below.

- https://github.com/WordPress/wporg-main-2022/pull/396
- https://github.com/WordPress/wporg-main-2022/pull/273
- https://github.com/WordPress/wporg-main-2022/pull/364

However, if a Rosetta site is not yet a new version, the page **template** below is loaded.

https://github.com/WordPress/wordpress.org/blob/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-main/page-about-requirements.php

Therefore, it is necessary to update the recommended version of MySQL/MariaDB even in locales where the new version has not yet been applied.

Below is [the current Requirements page for the Japanese locale](https://ja.wordpress.org/about/requirements/), which is not yet a new version.

ja wordpress org_about_requirements

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-main/page-about-requirements.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-main/page-about-requirements.php
@@ -51,7 +51,7 @@ the_post();
 						<li>
 							<?php
 							/* translators: 1: URL to MySQL; 2: MySQL Version; 3: URL to MariaDB; 4: MariaDB Version */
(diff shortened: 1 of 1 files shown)


dd32:
My understanding is that this requirements page in the older theme is no longer in use anywhere, and this pr can be closed.

PR #244, opened 2024-04-11
wporg-support-2024: Include locale specific styles

### LINKED ISSUE #240: Support forums: Apply language-specific customization
We used to have issues with unexpected line break in the Japanese Rosetta site but managed to handle it by adding language specific customization. cf https://github.com/WordPress/wporg-parent-2021/pull/123
 
However this customization is not applied in the redesigned support forum.

|Current|Expected|
|:--|:--|
|Screenshot 2024-04-10 at 5 14 49 PM|Screenshot 2024-04-10 at 5 15 10 PM|

adamwoodnz:
Include `wporg-parent-2021/inc/rosetta-styles.php` [in a similar manner to that theme](https://github.com/WordPress/wporg-parent-2021/blob/trunk/source/wp-content/themes/wporg-parent-2021/functions.php#L11), but in this case we need to navigate theme directories.

Fixes https://github.com/WordPress/wordpress.org/issues/240

## Testing
1. Visit https://ja.wordpress.org/support/?new-theme=1
2. Inspect a heading
3. `--wp--custom--heading--typography--text-wrap: unset` should be set (default is `balance`)

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/functions.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/functions.php
@@ -6,6 +6,11 @@
  */
 
 /**
+ * Include locale specific styles.
+ */
+require_once get_theme_root() . '/wporg-parent-2021/inc/rosetta-styles.php';
+
+/**
  * Use the ‘Lead Topic’ uses the single topic part
  * allowing styling the lead topic separately from the main reply loop.
  */


wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/functions.php
```diff
@@ -5,6 +5,11 @@
  * @package WPBBP
  */
 
+/**
+ * Include locale specific styles.
+ */
+require_once dirname( dirname( get_template_directory() ) ) . '/wporg-parent-2021/inc/rosetta-styles.php';
```
adamwoodnz: Is there a better way to reference this file?

We do something similar for the [theme.json](https://github.com/WordPress/wordpress.org/blob/trunk/wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/functions.php#L76)

ryelle: Maybe [get_theme_root](https://developer.wordpress.org/reference/functions/get_theme_root/)? I haven't tried it, but `get_theme_root( 'wporg-parent-2021' )` should work.

adamwoodnz: Thanks, `require_once get_theme_root() . '/wporg-parent-2021/inc/rosetta-styles.php';` works 🎉

PR #250, opened 2024-04-15
Support Forums: Allow rosetta configuration of local nav menu items

### LINKED ISSUE #236: Support forums: Pages in local navigation don't exist on Rosetta sites
See https://github.com/WordPress/wporg-main-2022/issues/430 — someone was looking for the German "Guidelines" page, but the link 404s. The local nav has local links to "Guidelines" and "Welcome to Support", but these don't exist on Rosetta sites.

Screenshot 2024-04-09 at 1 26 23 PM

The English pages are just pages, so it's not translatable content at the moment. Maybe these pages should also have a sync-to-patterns process?

adamwoodnz:
Allows rosetta support sites to configure a `navigation` menu with items to replace the `en_US` items.

If a menu is not found with that name, the fallback is a menu with only the Get Involved link.

Fixes #236 

### Screenshots

| Admin menu | With configured menu (es_ES) | Without configured menu (de_DE) |
|-|-|-|
| Screenshot 2024-04-15 at 12 15 42 PM | Screenshot 2024-04-15 at 12 17 41 PM | Screenshot 2024-04-15 at 12 40 46 PM |

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/functions.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/functions.php
@@ -17,10 +17,22 @@ require_once get_theme_root( 'wporg-parent-2021' ) . '/wporg-parent-2021/inc/ros
 add_filter( 'bbp_show_lead_topic', '__return_true' );
 
 /**
+ * Get the local navigation menu object if it exists.
+ */
+function get_local_nav_menu_object() {
+	$local_nav_menu_locations = get_nav_menu_locations();
+	$local_nav_menu_object = isset( $local_nav_menu_locations['local-navigation'] )
+		? wp_get_nav_menu_object( $local_nav_menu_locations['local-navigation'] )
+		: false;
+
+	return $local_nav_menu_object;
+}
+
+/**
  * Register a local nav menu for non-English forums, if it doesn't already exist.
  */
 function register_local_nav_menu() {
-	if ( substr( get_locale(), 0, 2 ) === 'en' || wp_get_nav_menu_object( 'local-navigation' ) ) {
+	if ( substr( get_locale(), 0, 2 ) === 'en' || get_local_nav_menu_object() ) {
 		return;
 	}
 
@@ -54,7 +66,7 @@ function add_site_navigation_menus( $menus ) {
 			),
 		);
 	} else {
-		$local_nav_menu_object = wp_get_nav_menu_object( 'local-navigation' );
+		$local_nav_menu_object = get_local_nav_menu_object();
 		$menu_items_fallback = array(
 			'forums' => array(
 				 array(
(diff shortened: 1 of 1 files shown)


ndiego:
Cool idea. What do you think @javiercasares?

javiercasares:
> Cool idea. What do you think @javiercasares

I think it looks great! and It will be very helpful for locale sites ;)

ryelle:
This requires the site owners to create a new menu, and call it `navigation` exactly, not navigation in their own language, not "Navigation", right? Would it add extra complexity to the code to use (register_nav_menu)[https://developer.wordpress.org/reference/functions/register_nav_menu/] to create a menu location instead, so the menu name doesn't matter?

javiercasares:
In the past, there were some things like that (naming things in some ways) so, it shouldn't be a problem if it's documented… Maybe the way is to create a "menu zone" like "Forums menu" and put whatever we want there.

adamwoodnz:
> Would it add extra complexity to the code to use [register_nav_menu](https://developer.wordpress.org/reference/functions/register_nav_menu/) to create a menu location instead, so the menu name doesn't matter?

I thought there might be a smart way to configure this, thanks @ryelle 🙂 

AFAIK it's not possible to change the slug in the menu UI so I think creating it programmatically is a good call.

wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/functions.php
```diff
@@ -16,26 +16,77 @@
  */
 add_filter( 'bbp_show_lead_topic', '__return_true' );
 
+/**
+ * Register a local nav menu for non-English forums, if it doesn't already exist.
+ */
+function register_local_nav_menu() {
```
adamwoodnz, line 34: Added this @ryelle but it doesn't seem to work, any ideas? When I debug it seems to be called correctly but is never added to the menu locations array.

wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/functions.php
```diff
@@ -16,26 +16,77 @@
...
+					'label' => __( 'Get Involved', 'wporg-forums' ),
+					'url' => 'https://make.wordpress.org/support/handbook/contributing-to-the-wordpress-forums/',
+				)
 			),
-			array(
-				'label' => __( 'Guidelines', 'wporg-forums' ),
-				'url' => '/guidelines/',
+		);
+	} else {
+		$local_nav_menu_object = wp_get_nav_menu_object( 'local-navigation' );
```
ryelle: The `wp_get_nav_menu_object` function doesn't understand locations, so you need to get the menu at this location first, then you can use it. I'm guessing this will fix the issue [you mentioned here](https://github.com/WordPress/wordpress.org/pull/250/files#r1566580291).

```suggestion
		$local_nav_menu_locations = get_nav_menu_locations();
		$local_nav_menu_object = isset( $local_nav_menu_locations['local-navigation'] ) ? wp_get_nav_menu_object( $local_nav_menu_locations['local-navigation'] ) : false;
```

ryelle: I ended up creating a test menu on es.w.org/support to test this out, since the Location wouldn't even appear in wp-admin until after a menu existed.

PR #254, opened 2024-04-16
Plugins 2024 Theme: SVG alignment with the text

marcarmengou:
Increased the height of the SVG to align it with the text

Before: `.plugin-card footer span svg{height:24px;` - After: `.plugin-card footer span svg{height:29px;`


obenland:
The SVGs are meant to be vertically centered. That may look a little odd in the case of the bar chart icon, but can be observed in the other elements.

Closing as this PR has conflicts and existing code works as expected.

PR #269, opened 2024-04-18
Support Forums: Make urls editable for localization

mkismy:
In the footer area I found 2 links that was hard coded to English sites, , `https://wordpress.org/documentation/` and `https://make.wordpress.org/support/handbook/`.
These links should be different for each locale.

image


StevenDufresne:
Thanks for the Pull Request. 

I implemented it in this commit https://meta.trac.wordpress.org/changeset/13861.

I understand the current need to update these URLs, however, we should try to converge on the same resources as opposed to managing different ones.

obenland:
Fixed in [meta.trac.wordpress.org/changeset/13861](https://meta.trac.wordpress.org/changeset/13861).

PR #320, opened 2024-05-17
Translate: Add additional custom errors

markoheijnen:
I was talking with @Zodiac1978 and he created https://gist.github.com/Zodiac1978/1d1bd363d5ba7d1ea68e3ffe352008f1. Based on that I updated the plugin with additional checks.

I also updated `is_core_project()` so it works for the main project as well.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-custom-errors/wporg-gp-custom-errors.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-custom-errors/wporg-gp-custom-errors.php
@@ -16,6 +16,20 @@ class WPorg_GP_Custom_Translation_Errors {
 	const WORPRESS_CORE_PROJECT_ID = 1;
 
 	/**
+	 * The parent WordPress core project ID.
+	 *
+	 * @var int
+	 */
+	const CONTEXT_ENUM_CHECK_LIST = [
+		'text direction' => [ 'ltr', 'rtl' ],
+		'Open Sans font: on or off' => [ 'on', 'off' ],
+		'Open Sans font: add new subset (greek, cyrillic, vietnamese)' => [ 'no-subset', 'greek', 'cyrillic', 'vietnamese' ],
+		'Word count type. Do not translate!' => [ 'characters_excluding_spaces', 'characters_including_spaces', 'words' ],
+		'decline months names: on or off' => [ 'on', 'off' ],
+		'Comment number declension: on or off' => [ 'on', 'off' ],
+	];
+
+	/**
 	 * Registers all methods starting with error_ with GlotPress.
 	 */
 	public function __construct() {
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/plugins/wporg-gp-custom-errors/wporg-gp-custom-errors.php
```diff
@@ -186,7 +203,81 @@ public function error_timezone_date_format( $original, $translation, $gp_origina
...
+			return true;
+		}
+		if ( in_array( $translation, self::CONTEXT_ENUM_CHECK_LIST[$gp_original->context]) ) {
+			return true;
+		}
+
+		return esc_html__(
+			'Must be ' . $this->array_to_string( self::CONTEXT_ENUM_CHECK_LIST[$gp_original->context] ),
+			'glotpress'
+		);
```
dd32, line 235: This should be a literal string, given it's being translated:

```suggestion
		return sptintnt(
			esc_html__(
				'Must be one of %s',
				'glotpress'
			),
			 $this->array_to_string( self::CONTEXT_ENUM_CHECK_LIST[$gp_original->context] )
		 );
```

wordpress.org/public_html/wp-content/plugins/wporg-gp-custom-errors/wporg-gp-custom-errors.php
```diff
@@ -169,10 +186,10 @@ public function error_timezone_date_format( $original, $translation, $gp_origina
 		if ( ! $this->is_core_project( $gp_original ) ) {
 			return true;
 		}
-		if ( is_null( $gp_original->context ) ) {
+		if ( is_null( $gp_original->comment ) ) {
 			return true;
 		}
-		if ( ! str_contains( $gp_original->context, 'timezone date format' ) ) {
+		if ( strpos( $gp_original->comment, 'https://www.php.net/manual/datetime.format.php' ) === false ) {
```
dd32, line 192: This can keep using `str_contains()` as WordPress has a polyfill for it. Unless you want to do a case-insensitive check then this should be a `stripos`.

wordpress.org/public_html/wp-content/plugins/wporg-gp-custom-errors/wporg-gp-custom-errors.php
```diff
@@ -15,6 +15,20 @@ class WPorg_GP_Custom_Translation_Errors {
...
+	 * @var int
+	 */
+	const CONTEXT_ENUM_CHECK_LIST = [
+		'text direction' => [ 'ltr', 'rtl' ],
+		'Open Sans font: on or off' => [ 'on', 'off' ],
+		'Open Sans font: add new subset (greek, cyrillic, vietnamese)' => [ 'no-subset', 'greek', 'cyrillic', 'vietnamese' ],
+		'Word count type. Do not translate!' => [ 'characters_excluding_spaces', 'characters_including_spaces', 'words' ],
+		'decline months names: on or off' => [ 'on', 'off' ],
+		'Comment number declension: on or off' => [ 'on', 'off' ],
+	];
```
dd32, line 30: These in general follow a specific format, I wonder if we could add something more generic.

 - `Word: ... (option1, option2)` = must be option1 or option2
 - `Word: option1 or option2` = Must be option1 or option2
 - `Do not translate` in the comment, and neither of the above = `Triple check please`?

Wouldn't cover 100% of cases, but would help with such strings in the future.

PR #321, opened 2024-05-17
Plugin Directory: Use a rounded number for the plugins count

ryelle:
See https://github.com/WordPress/wporg-theme-directory/issues/68#issuecomment-2115897450. The Theme Directory uses a rounded number ("Over 12,000 free themes to customize your WordPress site"), which is an easier number to read. This PR updates the plugin directory header to round down to the nearest thousand, and adds "over" — which will give us the tagline:

> Extend your WordPress experience! Browse over 59,000 free plugins.

Screen Shot 2024-05-17 at 14 45 05

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/patterns/front-page-header.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/patterns/front-page-header.php
@@ -5,6 +5,18 @@
  * Inserter: no
  */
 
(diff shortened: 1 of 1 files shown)


StevenDufresne:
Codewise, this looks fine. 👍  

I can't speak to the reason why it's an exact number and whether that was purposeful. @dd32 Any instinct about rounding this number?

dd32 (approved):
I have no relevant historical knowledge here, but I agree that rounding it would be best.

PR #322, opened 2024-05-21
Page Limiter: Update hook for setting max values so it applies to cached queries.

ryelle:
See some discussion on https://github.com/WordPress/wporg-theme-directory/issues/69, https://github.com/WordPress/wporg-theme-directory/issues/68. 

To recap, on the [new theme directory preview](https://wordpress.org/themes/?new-theme=1), the count of total themes & pages for a given archive would sometimes use the "49 page limit" set in `WPORG_Page_Limiter`, regardless of logged-in status. This seems to be due to query caching, the `found_posts` filter only runs before caching, not again after returning the cached results. The `found_posts` value is cached along with the results, so it's saved when the filter runs for a logged out user, and then returned for a logged in user. 

The opposite was also true, logged-out users might see the full 100+ pages available (though attempting to view page 50+ would correctly return a 404).

This PR switches the filter for updating `found_posts` to run on `posts_results` and set the query properties directly. This ensures the filter is run for all queries.

I've also set a custom property on `$query` for the original found_posts value, so that it can be passed through to the Query Total block. Like so: https://github.com/WordPress/wporg-mu-plugins/compare/try/page-limiter-hook — This is a general issue with the page-limit code + Query Total block, we just haven't run into it until now.

**To test**

- View the theme directory
- Logged in views should show all pages
- Logged out views should show a max of 49 pages
- Ensure this works on other sites: News, Patterns, Plugins, etc

### DIFF
--- a/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-page-limiter.php
+++ b/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-page-limiter.php
@@ -30,8 +30,7 @@ class WPORG_Page_Limiter {
 		// 404 on high pages instead of performing a DB query.
 		add_action( 'parse_query', [ $this, 'parse_query' ], 100 );
 
-		// Limit WP_Query::max_num_pages to self::MAX_PAGES
-		add_filter( 'found_posts', [ $this, 'found_posts' ], 100, 2 );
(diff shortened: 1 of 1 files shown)


ryelle:
I'm going to commit this now since it's a pretty visible bug on the theme directory, and I want to open that to wider feedback today. But feel free to leave any feedback here and I can iterate on it  🙂

PR #323, opened 2024-05-23
Add rejection reason "not-exceptional"

benniledl:
The photo directory is for exceptional photos, and if a photo can be taken by anybody, then it is not exceptional. It needs to stand out for some reason. That reason can be that it is beautiful, or that it is in some other way great. Keep the quality high, not only the resolution.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/photo-directory/inc/rejection.php
+++ b/wordpress.org/public_html/wp-content/plugins/photo-directory/inc/rejection.php
@@ -72,6 +72,10 @@ class Rejection {
 				'label' => __( 'Not a photo: Not a photo (e.g. screenshot, digital art)', 'wporg-photos' ),
 				'email' => __( 'The image did not appear to be a photograph. We do not accept screenshots, digital art, or other non-photographic images.', 'wporg-photos' ),
 			],
+			'not_exceptional'    => [
+				'label' => __( 'Not exceptional: Nothing special', 'wporg-photos' ),
+				'email' => __( 'The photo did not stand out as being exceptional in quality or subject matter.', 'wporg-photos' ),
+			],
 			'other'         => [
 				'label' => __( 'See below: Reason specified below', 'wporg-photos' ),
 				'email' => '',


coffee2code:
I agree with the intent here. I've debated if we needed a more explicit way to decline photos that, due to the subject matter, are simply not good enough. There's a great deal of subjectivity to making that assessment. And more than that, great care should be taken when conveying this assessment.

There is an existing rejection reason for "quality". The label for mods is:
> Quality: Insufficient image quality (e.g. blurriness, composition, lighting, lens issues)

The snippet included in rejection emails is:
> The photo had an issue regarding image quality. Submissions should be of high quality composition, lighting, focus, and color. The image should be free of blur (for the primary subject), noise, lens flare, glare, and spots due to water or dirt on the lens.

It doesn't focus on subject matter as much as on some objective issues of the photo (e.g. too blurry, dust on lens, focus, etc). But originally (and currently) this rejection reason  was intended to cover the not-exceptional reason as well. But there can be merit in making it a separate and explicit rejection reason.

You proposed `"The photo did not stand out as being exceptional in quality or subject matter."` as the text included in rejection emails. "Quality" is rather vague and, as noted, already an existing rejection reason. Maybe it should be replaced with something like "composition"?

Do others agree on the need for this new rejection reason? Do you feel this is a sufficient explanation without seeming harsh? 

One related idea I've considered: if a photo would be rejected as not-exceptional, what if it had to also get a second opinion from another moderator? The UI would show mods that a photo was deemed not-exceptional by a given moderator, and if another moderator also feels the same way, then the photo can be declined.

werkform:
How can you resolve the tension between objective quality and subjective perception of quality? **Not at all.** But you can generally improve the quality of the photo directory by 
a) giving the moderators more options (so yes: add the requested option), 
b) reporting about it as transparently as possible (learn from wikipedia) and 
c) an option for Contributors can also improve the quality themselves through meta information (rating, additional description, ...)

benniledl:
I agree that the proposed text for the email is not clear enough. Let's go somewhere like 

> The photo did not stand out as being exceptional in subject matter, photos that do not stand out and can be made easily by anyone are not the scope of or do not benefit a public directory .

I do not support the need for a second moderator's opinion for these rejections. This would additional work, and I think the photos that would be covered with the "not-exceptional" rule are likely to be clearly below standard, making unanimous agreement likely, 

However, conducting a **test run** to see how often a second opinion aligns with the initial "not-exceptional" judgment could be interesting.

roytanck:
I agree that the existing image quality option is primarily about the technical quality of the image. It would be helpful to have an option to reject based on (a lack of) "artistic/creative quality". I'd prefer that term over "outstanding" or "exceptional". 

Not all good photo's are exceptional (although we obviously would love those). But we do want images where the photographer took the effort to make it as nice as possible. Essentially, we're looking for a way to weed out "snapshots", right?

I feel that "composition" en "lighting" should be moved from the existing quality option to the new one. The existing one would then be just about technical shortcomings.

coffee2code:
I overlooked it previously, and maybe everyone else has as well, but there is already this rejection option:

> Label: Subject matter: Image subject matter
> Text used in email: The photo included subject matter of insufficient quality.

The array key for that reason is "image_subject".

This rejection reason would be the one used for the types of rejections this ticket targets.

This rejection reason represents 1% of rejections, whereas image_quality is 56%, so it clearly isn't being used very often. Perhaps rather than a new 'not-exceptional' reason, we adjust the wording of the 'image_subject' reason to make it clearer? Coupled with an announcement to moderators to make them more aware of its presence and purpose.

obenland:
> Perhaps rather than a new 'not-exceptional' reason, we adjust the wording of the 'image_subject' reason to make it clearer? Coupled with an announcement to moderators to make them more aware of its presence and purpose.

Closing the PR in favor of this resolution.

Issue #327, opened 2024-06-10
Plugin Directory: The searchbar on single page is not bound to its container.


StevenDufresne:
This is likely due to a Gutenberg update.

## Screenshots

<img width="1728" alt="Screenshot 2024-06-10 at 9 44 24 AM" src="https://github.com/WordPress/wordpress.org/assets/1657336/8102d7dc-369a-4455-bad3-2ee358c6bfa6">

PR #328, opened 2024-06-10
Plugin Directory: Move spacing to parent element for search bar

ryelle:
This is an alternate solution to #327 — rather than creating a duplicate container for the search bar, the left/right spacing can be added to the `alignfull` element directly. This adds back the expected left/right spacing now that nested alignfull elements have negative margins.

There were other changes in [the commit](https://github.com/WordPress/wordpress.org/commit/a05e8bbb772f8446adc5d34a8f7baeb8b1ffb7fd) that I left alone, except for one— I moved the search bar back into the `main` tag. This does not change anything visually.


StevenDufresne:
If it works, ship it!

ryelle:
Committed in https://meta.trac.wordpress.org/changeset/13855

PR #329, opened 2024-06-11
Themes: Try loading a playground instance for theme previews.

dd32:
This is a draft exploring whether using a playground instance for Theme Previews is viable.

This..
 - Is based on the current theme previewer, NOT the new theme directory previewer, a PR for that is linked below.
 - Includes support for a theme to have a custom `preview-blueprint.json`
 - Includes a custom plugin to load starter content into the fresh playground instance (bin/theme-preview-load-starter-content.php)
 - Loads the existing `style-variations` and `pattern-preview` plugins from wp-themes.com (Temporarily hosted in a github repo)
   - The `pattern-preview` needs a page to exist, so that's also forced in, this works for most themes.

Observations:
 - On a fast connection, I experience about an 8s initial load time (when playground resources are locally cached).
 - Playground offers a better preview experience IMHO
 - Some starter content imports are not 100% the same as on wp-themes.com, unsure if this is a wp-themes starter content bug the themes have worked around, or if this is a bug in the import.


dd32:
Merged.

PR #336, opened 2024-06-30
Fixed : Issue related to "Ensures links have discernible text" on learn.wordpress.org

narenin:
Trac ticket: https://meta.trac.wordpress.org/ticket/7688


adamwoodnz (commented):
Thanks for the PR @narenin, however this theme is being superseded and will no longer be in use later this month, so I'm going to close the meta ticket as won't fix.

PR #337, opened 2024-06-30
Fixed : Color Contrast Issue on Hosting Page of WordPress.org

narenin:
Trac ticket: https://meta.trac.wordpress.org/ticket/7689


obenland:
The subtitle was removed on March 5, 2026 when the page was switched away from using the page template.

I'll close the PR for now as this is no longer applicable. Please feel free to reopen if there are other pages that are effected by this.

PR #338, opened 2024-06-30
fix Color Contrast Issue on Five for the Future Page of WordPress.org

deepakrohillas:
With background color: #1e8cbe andforeground color: #ffffff, The block Element has insufficient color contrast of 3.78. and Expected contrast ratio of 4.5.
since color:#fff (white) is fine but background color #1e8cbe is causing issue,
Using background color: #1a7eab with original foreground color: #ffffff , contrast ratio 4.5 has meet.

Trac ticket: https://meta.trac.wordpress.org/ticket/7690


obenland:
This was fixed with the switch to the `wporg-5ftf-2024` theme on Sep 13, 2024.

PR #339, opened 2024-06-30
Strict comparison missign with post get_post_type() check

deepakrohillas:
Here multiple places strict check is missing and alignment issue is found

Trac ticket: https://meta.trac.wordpress.org/ticket/7691


obenland:
There are a vast amount of code quality fixes that can be done across the meta repository. I'm not sure selected patches like this are the best way solve them. These should rather be updates as part of regular work or in a coordinated effort like it was done in Core.

PR #340, opened 2024-06-30
Missing Escaping, Strict Check, and PHPCS Issues in Multiple Files

deepakrohillas:
Trac ticket: https://meta.trac.wordpress.org/ticket/7692


obenland:
There are a vast amount of code quality fixes that can be done across the meta repository. I'm not sure selected patches like this are the best way solve them. These should rather be updates as part of regular work or in a coordinated effort like it was done in Core.

PR #342, opened 2024-07-01
Support forums: Remove border in content on rosetta homepage

ryelle:
Some rosetta sites have an unexpected border above the last section of card-links. This seems to be left over from the old theme, so I've simply removed it. If the border is necessary, it can be brought back, but we'll want to tweak the spacing around it.

| Before | After |
|---|---|
|  |  |


adamwoodnz (approved):
I agree it does look leftover, and I can't think of anywhere it would be appropriate in the new theme. Thanks for fixing.

ryelle:
Committed: https://meta.trac.wordpress.org/changeset/13895

PR #344, opened 2024-07-08
Plugin Directory: Release Confirmation: Require 2FA instead of email verification

dd32:
When a user has 2FA setup, we should rely upon that instead of email.

This is more secure, and for those who are using Keys, likely more streamlined.

This is a work in progress, and requires a few more steps.

See https://meta.trac.wordpress.org/ticket/7704

This is reliant upon https://github.com/WordPress/wporg-two-factor/pull/283


dd32:
Rather than having a prompt to 2FA, this would be better implemented if the buttons were just shown as clickable, but upon click the JS handler for revalidation prompted it at that point.

PR #350, opened 2024-07-17
Plugin Directory: Use the shared favorite button for favoriting plugins

### LINKED ISSUE #265: Accessibility: Plugin favorites heart color contrast failure
On the plugin single page ([Gutenberg plugin test page](https://test.wordpress.org/plugins/gutenberg/)), the plugin favorites heart in its unselected state is #cbcdce on a white background which is 1:59:1.

Screenshot 2024-04-16 at 5 45 36 PM

Here's the[ contrast on WebAIM's checker.](https://webaim.org/resources/contrastchecker/?fcolor=CBCDCE&bcolor=FFFFFF) This fails for UI components and many people with low-vision will not even know that is there.

Because we don't want to fail Use of Color, it would not be sufficient to just darken the color of the heart. I'd recommend adding a border to it so it has a border and light or no color if unselected and a solid color if hovered or selected.

ryelle:
This updates the single plugin page to use the Favorite block, and so that the buttons lined up correctly I've also switched the plugin action buttons (download, preview) to use button blocks as well. Download is now primary, with preview as an outline button, to prevent double primary buttons.

The focus state was also a little rough on the favorite block, so I've copied in the default focus styles from themes & patterns, which does update the style in the content as well (videos at the end to show before/after).

| | Before | After |
|---|---|---|
| Favorited |  |  |
| Unfavorited |  |  |

Focus states, before

https://github.com/user-attachments/assets/0e70b899-eae4-4bca-bcc5-1af0f3355621

Focus states, after

https://github.com/user-attachments/assets/b6665ad6-87c2-478c-8d26-a23a80be3d9e

**To test**

- Use a sandbox, be logged in
- Visit any plugin page
- The favorite button should appear next to the download button
- Click the button, it will trigger an API request but not reload the page
- The button should update to reflect the new favorite status
- Tab through the site, the focus state should be visible and match the text color
    - No visible focus on the screenshots or FAQs, but that's true on production too
- Download and Preview buttons should still work

If you're not logged in, the button is not visible.


StevenDufresne:
I tested this on my sandbox. It works as expected. It's a bit slow to change states and doesn't have an indicator. Do you think we need a loading state?

Screen Capture on 2024-07-18 at 10-47-35

ryelle:
> I tested this on my sandbox. It works as expected. It's a bit slow to change states and doesn't have an indicator. Do you think we need a loading state?

I _think_ that's just the sandbox. Do you think it needs a loading state on the other sites? It's already in place on themes & patterns.

ryelle:
I tried it out on themes again and it's pretty quick, but not quite instant. And it could probably use something for slower connections. What do you think of this fade animation?

https://github.com/user-attachments/assets/052851c6-ba61-437e-814b-22a9698241c8

That's using a delay to mimic a 2s load time. A faster example just barely has the fade, but it also adds a nice touch IMO:

https://github.com/user-attachments/assets/85728f80-e99f-4d37-b74f-0279b95ef5cd

StevenDufresne:
That's a neat animation. 👍 

I say we go for it!

ryelle:
Committed in [r13913](https://meta.trac.wordpress.org/changeset/13913).

PR #351, opened 2024-07-18
Add a 2FA nag on login for users who should have 2FA but don't.

dd32:
In conjunction with https://github.com/WordPress/wporg-two-factor/pull/288 this adds a nag upon login that the user should really setup 2FA.

I imagine instead of landing on the final account screen that's shown in the below video, it'd land on the wizard from https://github.com/WordPress/wporg-two-factor/pull/286 being triggered immediately.

https://github.com/user-attachments/assets/545668db-4bed-4d75-b479-ef0056352ea5


dd32:
TODO:
 - Replace documentation link with WordPress.org 2FA doc

PR #357, opened 2024-08-09
Delete a suggestion from TM

amieiro:
This PR adds a button, only for the GTE, in the suggestions from the translation memory, to be able to delete a suggestion from the translation memory. This button is added only if the similarity between the current original and the suggestion from the translation memory is 100%.

Button visible only for a Galician GTE in a [Galician string](https://translate.wordpress.org/projects/wp-plugins/zero-bs-crm/dev/gl/default/?filters%5Boriginal_id%5D=16467169).

image

If the user is not a GTE, she won't have this button. Image for a user who isn't a Spanish GTE in a [Spanish string](https://translate.wordpress.org/projects/wp-plugins/zero-bs-crm/dev/es/default/?filters%5Boriginal_id%5D=16467169).

image

This PR needs to implement the `delete` method for the `Translation_Memory_Client` class, but first we need a new method in the https://translate.wordpress.com/api/tm/ API.

The AJAX request implements the `POST` method, instead the `DELETE` method, because we need to send some data in the request, and the `DELETE` method doesn't allow it.


wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/js/translation-suggestions.js
```diff
@@ -387,27 +396,89 @@
...
 	}
+
+	/**
+	 * Removes a suggestion from the TM.
+	 *
+	 * @param {object} event
+	 *
+	 * @return {void}
+	 */
+	function deleteSuggestionFromTM( event ) {
```
akirk, line 412: I think we should add a `confirm()` here since after you delete it, it's not easy to get it back.

wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-translation-memory-client.php
```diff
@@ -132,22 +138,35 @@ public static function query( string $text, string $target_locale ) {
...
+	/**
+	 * Deletes a translation from translation memory.
+	 *
+	 * @param string $original_string    Original string.
+	 * @param string $translation_string Translation string.
+	 * @param string $locale_slug        Locale slug.
+	 * @param string $set_slug           Translation set slug.
+	 */
+	public static function delete( string $original_string, string $translation_string, string $locale_slug, string $set_slug ):bool {
+		// @todo Implement.
```
akirk: Please implement this like this. It will require changing the function signature.
```suggestion
		$body = wp_json_encode(
			array(
				'token'    => WPCOM_TM_TOKEN,
				'source'       => array(
					'singular' => $original->singular,
					'plural' => $original->plural,
					'context' => $original->context,
				),
				'translation' => array(
					'singular' => $translation->translation_0,
					'plural' => $translation->translation_1,
					'locale'   => $locale,
				)
			)
		);

		$request = wp_remote_post(
			self::API_BULK_ENDPOINT,
			array(
				'method'    => 'DELETE',
				'timeout'    => 10,
				'user-agent' => 'WordPress.org Translate',
				'body'       => $body,
			)
		);
```

akirk (changes requested):
I'm implementing the TM API changes on WordPress.com and added the required API call properties.

wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/js/translation-suggestions.js
amieiro: Done [here](https://github.com/WordPress/wordpress.org/pull/357/commits/a269763d1e07d5803a8332511a5f131ebe6660ed)!

wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-suggestions/inc/class-translation-memory-client.php
```diff
@@ -132,22 +138,35 @@ public static function query( string $text, string $target_locale ) {
...
+			return array();
 		}
 
-		$suggestions = [];
+		$suggestions = array();
 		foreach ( $result['matches'] as $match ) {
-			$suggestions[] = [
+			$suggestions[] = array(
 				'similarity_score' => $match['score'],
 				'source'           => $match['source'],
```
akirk:
```suggestion
				'source'           => $match['source'],
				'source_context'   => $match['source_context'],
```

PR #358, opened 2024-08-16
Show a notice on login when the user has no recovery codes recorded, is running low, or has logged in with a recovery code.

dd32:
See https://github.com/WordPress/wporg-two-factor/issues/279

This PR shows the following notice when logging in with a recovery code, or is running low on recovery codes.

| Login | Running low | Have no recovery codes |
| --- | --- | --- |
| Screenshot 2024-08-16 at 1 11 30 PM | Screenshot 2024-08-16 at 1 13 30 PM | Screenshot 2024-08-16 at 1 14 23 PM |

In the first case, the `I'll do this later` only shows if the user has at least one more recovery code on the account. I'm not against removing the bypass link entirely though.

I realise that the 2nd and 3rd have a bad orphan, I need some text inspiration :)

### DIFF
--- a/common/includes/wporg-sso/wp-plugin.php
+++ b/common/includes/wporg-sso/wp-plugin.php
@@ -893,10 +893,10 @@ if ( class_exists( 'WPOrg_SSO' ) && ! class_exists( 'WP_WPOrg_SSO' ) ) {
 			}
 
 			// If the user logged in with a recovery code..
-			$session_token        = wp_get_session_token() ?: ( $this->last_auth_cookie['token'] ?? '' );
-			$session              = WP_Session_Tokens::get_instance( $user->ID )->get( $session_token );
-			$used_recovery_code   = str_contains( $session['two-factor-provider'] ?? '', 'Backup_Codes' );
-			$codes_available      = Two_Factor_Backup_Codes::codes_remaining_for_user( $user );
+			$session_token      = wp_get_session_token() ?: ( $this->last_auth_cookie['token'] ?? '' );
+			$session            = WP_Session_Tokens::get_instance( $user->ID )->get( $session_token );
+			$used_recovery_code = str_contains( $session['two-factor-provider'] ?? '', 'Backup_Codes' );
+			$codes_available    = Two_Factor_Backup_Codes::codes_remaining_for_user( $user );
 
 			if (
 				// If they didn't use a recovery code,
--- a/common/includes/wporg-sso/wp-plugin.php
+++ b/common/includes/wporg-sso/wp-plugin.php
@@ -29,7 +29,7 @@ if ( class_exists( 'WPOrg_SSO' ) && ! class_exists( 'WP_WPOrg_SSO' ) ) {
 			// Primarily for logged in users.
 			'updated-tos'     => '/updated-policies',
 			'enable-2fa'      => '/enable-2fa',
-			'recovery-codes'  => '/recovery-codes',
+			'backup-codes'    => '/backup-codes',
 			'logout'          => '/logout',
 
 			// Primarily for logged out users.
@@ -110,7 +110,7 @@ if ( class_exists( 'WPOrg_SSO' ) && ! class_exists( 'WP_WPOrg_SSO' ) ) {
 					add_action( 'set_auth_cookie', [ $this, 'record_last_auth_cookie' ], 10, 6 );
 
 					// Maybe nag about 2FA
-					add_filter( 'login_redirect', [ $this, 'maybe_redirect_to_recovery_codes' ], 500, 3 );
+					add_filter( 'login_redirect', [ $this, 'maybe_redirect_to_backup_codes' ], 500, 3 );
 					add_filter( 'login_redirect', [ $this, 'maybe_redirect_to_enable_2fa' ], 1100, 3 );
 				}
 			}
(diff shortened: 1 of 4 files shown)


common/includes/wporg-sso/wp-plugin.php
```diff
@@ -851,6 +872,53 @@ public function maybe_redirect_to_enable_2fa( $redirect, $orig_redirect, $user )
...
+			$session              = WP_Session_Tokens::get_instance( $user->ID )->get( $session_token );
+			$used_recovery_code   = str_contains( $session['two-factor-provider'] ?? '', 'Backup_Codes' );
+			$codes_available      = Two_Factor_Backup_Codes::codes_remaining_for_user( $user );
+
+			if (
+				// If they didn't use a recovery code,
+				! $used_recovery_code &&
+				(
+					// They have ample codes available..
+					$codes_available > 3 ||
```
dd32, line 906:
```suggestion
					$codes_available > 5 ||
```

This should be 5 to match the 2FA interface.

https://github.com/WordPress/wporg-two-factor/blob/4a154cc6a7e37a5acc28d467cff327cd80712e88/settings/src/components/backup-codes.js#L230

common/includes/wporg-sso/wp-plugin.php
```diff
@@ -98,12 +106,25 @@ public function __construct() {
...
 					// Maybe nag about 2FA
-					add_filter( 'login_redirect', [ $this, 'maybe_redirect_to_enable_2fa' ], 1000, 3 );
+					add_filter( 'login_redirect', [ $this, 'maybe_redirect_to_recovery_codes' ], 500, 3 );
+					add_filter( 'login_redirect', [ $this, 'maybe_redirect_to_enable_2fa' ], 1100, 3 );
 				}
 			}
 		}
 
+		/**
+		 * Records the last set cookies, because WordPress.
```
StevenDufresne, line 120: 🤣  I understand this. 

The documentation might be pretty confusing for someone skimming in the future.

wordpress.org/public_html/wp-content/themes/pub/wporg-login/recovery-codes.php
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		_e( 'Account Recovery Codes', 'wporg-login' );
+	}
+?></h2>
+
+<p>&nbsp;</p>
+
+<p><?php
+	if ( $used_recovery_code ) {
+		_e( "You've logged in with a Recovery Code.<br>These codes are intended to be used when you lose access to your authentication device.<br>Please take a moment to review your Account Settings and ensure your Two-Factor settings are updated.", 'wporg-login' );
```
StevenDufresne:
```suggestion
		_e( "You've logged in with a Recovery Code.<br>These codes are intended to be used when you lose access to your authentication device.<br>Please take a moment to review your account settings and ensure your Two-Factor settings are up-to-date.", 'wporg-login' );
```
I think account settings can be lowercase since we don't have a matching string of "Account Settings" in any of our interfaces.

StevenDufresne:
This looks good. Our interfaces uses "Backup Codes" as opposed to "Recovery codes". We should remain consistent.

common/includes/wporg-sso/wp-plugin.php
dd32: Totally agree :) 

```suggestion
		 * Records the last set cookies, because WordPress.
		 *
		 * During the WordPress login process, the authentication cookies are not yet available,
		 * but we need to know the user token (contained in those cookies) to retrieve their session.
		 * To work around this, we store the set authentication cookies here for later usage.
```

dd32:
> Our interfaces uses "Backup Codes" as opposed to "Recovery codes". We should remain consistent.

I was thinking the same thing. We should probably standardise on Recovery Codes, as that's what we show on the Login page after the upstream rename https://github.com/WordPress/two-factor/pull/521

common/includes/wporg-sso/wp-plugin.php
dd32: I'm tempted to leave this at 3 despite the settings UI switching to a warning at 5, it makes sense that we'd nag on login later than we'd nag at them just casually browsing the settings UI.

Mind you, having used more than 50% of your codes is probably a sign you use them fairly often..

wordpress.org/public_html/wp-content/themes/pub/wporg-login/recovery-codes.php
```diff
@@ -0,0 +1,88 @@
...
+
+<p>&nbsp;</p>
+
+<p><?php
+	_e( 'If you run out of Recovery Codes and no longer have access to your Authentication device, you are at risk of being locked out of your WordPress.org account if we are unable to verify account ownership.', 'wporg-login' );
+?></p>
+
+<p>&nbsp;</p>
+
+<p><a href="<?php echo esc_url( $account_settings_url ); ?>"><button class="button-primary"><?php _e( 'Take me to Account Settings', 'wporg-login' ); ?></button></a></p>
```
StevenDufresne:
```suggestion
<p><a href="<?php echo esc_url( $account_settings_url ); ?>"><button class="button-primary"><?php _e( 'View my account settings', 'wporg-login' ); ?></button></a></p>
```
I [think we've agreed](https://github.com/WordPress/wporg-two-factor/pull/286#issuecomment-2283658180) on sentence case for everything but proper nouns.

common/includes/wporg-sso/wp-plugin.php
```diff
@@ -851,6 +876,53 @@ public function maybe_redirect_to_enable_2fa( $redirect, $orig_redirect, $user )
...
+			) {
+				// Then we don't need to redirect to the enable 2FA page.
+				return $redirect;
+			}
+
+			// If the user logged in with a recovery code..
+			$session_token        = wp_get_session_token() ?: ( $this->last_auth_cookie['token'] ?? '' );
+			$session              = WP_Session_Tokens::get_instance( $user->ID )->get( $session_token );
+			$used_recovery_code   = str_contains( $session['two-factor-provider'] ?? '', 'Backup_Codes' );
+			$codes_available      = Two_Factor_Backup_Codes::codes_remaining_for_user( $user );
```
dd32:
```suggestion
			// If the user logged in with a recovery code..
			$session_token      = wp_get_session_token() ?: ( $this->last_auth_cookie['token'] ?? '' );
			$session            = WP_Session_Tokens::get_instance( $user->ID )->get( $session_token );
			$used_recovery_code = str_contains( $session['two-factor-provider'] ?? '', 'Backup_Codes' );
			$codes_available    = Two_Factor_Backup_Codes::codes_remaining_for_user( $user );
```

PR #360, opened 2024-09-03
Improve discoverability of Plugin Previews

akirk:
This modifies a few bits in the plugin directory to make it easier to discover plugin previews and how they work:

### Add a "Preview Not Available"
To indicate that there could be a preview button. The eye indicates that this is only visible to the plugin admin.

Before:
Screenshot 2024-09-03 at 10 48 37

After:
Screenshot 2024-09-03 at 12 11 42

### Link the documentation on the advanced page
Before:
Screenshot 2024-09-03 at 10 11 26

After:
Screenshot 2024-09-03 at 11 27 23

### Add an eye to the "Test Preview"
Indicate that the "Test Preview" doesn't mean the preview is enabled:
Before:
Screenshot 2024-09-03 at 10 41 15

After:
Screenshot 2024-09-03 at 11 30 04

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/template-tags.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/template-tags.php
@@ -590,12 +590,12 @@ function the_plugin_self_toggle_preview_button() {
 		return;
 	}
 
-	echo '<h4>' . esc_html__( 'Toggle Live Preview', 'wporg-plugins' ) . '</h4>';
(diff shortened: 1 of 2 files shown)


StevenDufresne:
> **Add a "Preview Not Available**

I don't know either way whether that would work, but we can always run the experiment.

> **Link the documentation on the advanced page**

- We should change that logic so an error banner is not shown unless the blueprint is broken. 
- I would also consider updating the language to do a bit of selling: "Live Previews allow users to experience your plugin in real-time. Read more about Live Previews."

PR #362, opened 2024-09-04
Add svn password language.

StevenDufresne:
The approved plugin email should be updated to direct users to use svn passwords. This updates the approval email to use markdown to allow words to be linked. This has some minor effects on the layout, indenting lists.

| Before | After |
|--------|--------|
| Screenshot 2024-09-04 at 2 12 08 PM | Screenshot 2024-09-04 at 2 13 32 PM |

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
@@ -15,7 +15,7 @@ class Plugin_Approved extends Markdown_Base {
 	}
 
 	function markdown() {
-		/* translators: 1: plugin name, 2: plugin author's username, 3: plugin slug */
+		/* translators: 1: plugin name, 2: plugin author's username, 3: plugin slug, 4: link to plugin authors profile */
 		$email_text = __(
 'Congratulations, the plugin hosting request for %1$s has been approved.
 
@@ -64,7 +64,6 @@ If you have issues or questions, please reply to this email and let us know.
 
 Enjoy!', 'wporg-plugins' );
 
-		/* translators: 1: plugin name, 2: plugin author's username, 3: plugin slug, 4: link to plugin authors profile */
 		return sprintf(
 			$email_text,
 			$this->plugin->post_title,
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
@@ -3,7 +3,7 @@ namespace WordPressdotorg\Plugin_Directory\Email;
 
 use WordPressdotorg\Plugin_Directory\Tools;
 
-class Plugin_Approved extends Base {
+class Plugin_Approved extends Markdown_Base {
 	protected $required_args = [];
 
 	function subject() {
@@ -14,7 +14,7 @@ class Plugin_Approved extends Base {
 		);
 	}
 
-	function body() {
+	function markdown() {
 		/* translators: 1: plugin name, 2: plugin author's username, 3: plugin slug */
 		$email_text = __(
 'Congratulations, the plugin hosting request for %1$s has been approved.
@@ -29,35 +29,35 @@ Once your account access has been activated, you may upload your code using a SV
 To answer some common questions:
 
 * You must use SVN to upload your code -- we are unable to do that for you
-* Your SVN username is %2$s and is case sensitive.
-* Generate your SVN password in your <a href="https://profiles.wordpress.org/%2$s/profile/edit/group/3/?screen=svn-password">WordPress.org profile</a>.
+* Your SVN username is `%2$s` and is case sensitive.
+* Generate your SVN password in your [WordPress.org profile](https://profiles.wordpress.org/%2$s/profile/edit/group/3/?screen=svn-password).
 * SVN will not accept your email address as a username
 * Due to the size of the directory, it may take 72 hours before all search results are properly updated
 
 To help you get started, here are some links:
 
-Using Subversion with the WordPress Plugin Directory:
+Using Subversion with the WordPress Plugin Directory:<br>
 https://developer.wordpress.org/plugins/wordpress-org/how-to-use-subversion/
 
-Generating your SVN Password:
+Generating your SVN Password:<br>
 https://make.wordpress.org/meta/handbook/tutorials-guides/svn-access/
 
-FAQ about the WordPress Plugin Directory:
+FAQ about the WordPress Plugin Directory:<br>
 https://developer.wordpress.org/plugins/wordpress-org/plugin-developer-faq/
 
-WordPress Plugin Directory readme.txt standard:
+WordPress Plugin Directory readme.txt standard:<br>
 https://wordpress.org/plugins/developers/#readme
 
-A readme.txt validator:
+A readme.txt validator:<br>
 https://wordpress.org/plugins/developers/readme-validator/
 
-Plugin Assets (header images, etc):
+Plugin Assets (header images, etc):<br>
 https://developer.wordpress.org/plugins/wordpress-org/plugin-assets/
 
-WordPress Plugin Directory Guidelines:
+WordPress Plugin Directory Guidelines:<br>
 https://developer.wordpress.org/plugins/wordpress-org/detailed-plugin-guidelines/
 
-Block Specific Plugin Guidelines:
+Block Specific Plugin Guidelines:<br>
 https://developer.wordpress.org/plugins/wordpress-org/block-specific-plugin-guidelines/
 
 If you have issues or questions, please reply to this email and let us know.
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
```diff
@@ -29,8 +29,8 @@ function body() {
 To answer some common questions:
 
 * You must use SVN to upload your code -- we are unable to do that for you
-* Your SVN username is %2$s and your password is the same as you use to log in to WordPress.org
-* Your username is case sensitive
+* Your SVN username is %2$s and is case sensitive.
+* Generate your SVN password in your <a href="https://profiles.wordpress.org/%2$s/profile/edit/group/3/?screen=svn-password">WordPress.org profile</a>.
```
dd32: It appears that this might be a plain-text email, and not a HTML one.

This can probably be converted to a Markdown-based email
 - Change it to extend `Markdown_Base` instead of `Base`
 - Rename `function body()` to `function markdown()`

Test.

To test the email, you can do something like this:
```
( new WordPressdotorg\Plugin_Directory\Email\Plugin_Approved( get_post( 12345 ), get_user_by( 'slug', 'steve...' ) ) )->send();

or
$email = new WordPressdotorg\Plugin_Directory\Email\Plugin_Approved( get_post( 12345 ), get_user_by( 'slug', 'steve...' ) );

echo $email->html();

```

dd32 (changes requested):

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
StevenDufresne: Thanks for that. That does seem to work.

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
```diff
@@ -29,32 +29,35 @@ function body() {
 To answer some common questions:
 
 * You must use SVN to upload your code -- we are unable to do that for you
-* Your SVN username is %2$s and your password is the same as you use to log in to WordPress.org
-* Your username is case sensitive
+* Your SVN username is `%2$s` and is case sensitive.
+* Generate your SVN password in your [WordPress.org profile](https://profiles.wordpress.org/%2$s/profile/edit/group/3/?screen=svn-password).
```
dd32: SVN username/user_login is not the same as the the slug/nicename.

`%4$s` will need to be added, that references `$user->user_nicename`

StevenDufresne: Done in [13db8ab](https://github.com/WordPress/wordpress.org/pull/362/commits/13db8ab5d545874775970016f3a324d5c91c58e1).

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
```diff
@@ -29,32 +29,35 @@ function body() {
 To answer some common questions:
 
 * You must use SVN to upload your code -- we are unable to do that for you
-* Your SVN username is %2$s and your password is the same as you use to log in to WordPress.org
-* Your username is case sensitive
+* Your SVN username is `%2$s` and is case sensitive.
+* Generate your SVN password in your [WordPress.org profile](https://profiles.wordpress.org/%4$s/profile/edit/group/3/?screen=svn-password).
```
dd32: Apologies, one more...

Since this is a URL that might change, and this is a translatable string, perhaps this should be moved outside of the text.

ie. `WordPress.org profile](%4$s)`
&
`"https://profiles.wordpress.org/{$this->user->user_nicename}/profile/edit/group/3/?screen=svn-password "`
&
`/* translators: 1: plugin name, 2: plugin author's username, 3: plugin slug, 4: link to plugin authors profile */`

StevenDufresne: Is this what you had in mind? [7a0672d](https://github.com/WordPress/wordpress.org/pull/362/commits/7a0672dcab9e4190b5359c97b66d3b4808345350)

dd32: Yes, but the `translators` comment needs to be before the `__()` string, not where it's used

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
```diff
@@ -29,43 +29,48 @@ function body() {
 To answer some common questions:
 
 * You must use SVN to upload your code -- we are unable to do that for you
-* Your SVN username is %2$s and your password is the same as you use to log in to WordPress.org
-* Your username is case sensitive
+* Your SVN username is `%2$s` and is case sensitive.
+* Generate your SVN password in your [WordPress.org profile](%4$s).
```
frantorres: I'd suggest to make the change more noticiable as this would be the first place some plugin authors will see this, and they need to realize that it now works in a different way.

Maybe something like this:

```
* You can <strong>set up your SVN credentials</strong> (if you haven't already) in the “Account & Security” section of your [WordPress.org profile](%4$s).
* <i>Note: Previously, the credentials to access SVN were your WordPress login and password, this has changed. </i>
```

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
```diff
@@ -29,43 +29,48 @@ function body() {
...
-A readme.txt validator:
+A readme.txt validator:<br>
 https://wordpress.org/plugins/developers/readme-validator/
 
-Plugin Assets (header images, etc):
+Plugin Assets (header images, etc):<br>
 https://developer.wordpress.org/plugins/wordpress-org/plugin-assets/
 
-WordPress Plugin Directory Guidelines:
+WordPress Plugin Directory Guidelines:<br>
```
davidperezgar, line 57: I'd add Plugin Check as a tool for Developers to remind to keep using it.

```
Plugin Check: <br>
https://wordpress.org/plugins/plugin-check/
```

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
StevenDufresne: Good idea. I'm thinking overall this email needs an larger update that strongly suggests users use all _new/best-practice_ features (release from tags, svn passwords, release confirmations), I'll open a meta ticket for that.

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
StevenDufresne: Ha. What am I doin'... 😏 😭

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
dd32: Personally I would avoid language like `Previously...` and focus on new users.

Either they're an existing committer, and they should've been informed / learnt about the change, OR they're a new author who doesn't need to know about the historical aspects of it.

In other words; we should avoid having too much knowledge debt in automated emails (and predefs where possible)

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
StevenDufresne: I don't think it hurts to add it, but I don't believe that a developer would consider it useful though. Once they get through the review, they are thinking "release" over anything else.

davidperezgar: Yes, but It's the tool for developers and it's practical for new releases to keep the plugin up to date with security measures.

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
frantorres: I can agree with that, I just think there will be people who need more information to change what they're used to, but I also agree that we don't have to focus on them. 
In that case, just with the first phrase that's fine.
`* You can <strong>set up your SVN credentials</strong> (if you haven't already) in the “Account & Security” section of your [WordPress.org profile](%4$s).
`

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-plugin-approved.php
StevenDufresne: Fair enough. Can we make that change in a separate PR?

StevenDufresne:
Alright, I've made the relevant changes. @dd32 Am I good to deploy?

dd32 (approved):

StevenDufresne:
Deployed in https://meta.trac.wordpress.org/changeset/14073.

PR #364, opened 2024-09-05
Redirect users to profiles if they are viewing account.

StevenDufresne:
We moved account settings from the forums to the profiles. This redirect makes sure users with saved urls are moved there otherwise they will see an error.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-hooks.php
+++ b/wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-hooks.php
@@ -362,6 +362,15 @@ class Hooks {
 	public function redirect_legacy_urls() {
 		global $wp_query, $wp;
 
+		// Account information is no longer set in the support forums.
+		// We don't use `wp_get_current_user()` because super admins often reset data for other users.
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-hooks.php
```diff
@@ -362,6 +362,15 @@ public function redirect_update_php_page() {
 	public function redirect_legacy_urls() {
 		global $wp_query, $wp;
 
+		// Account information is no longer set in the support forums.
+		// We don't use `wp_get_current_user()` because super admins often reset data for other users.
+		if ( preg_match( '!^users/(?P<username>[^/]+)/edit/account!i', $wp->request, $matches ) ) {
+			$username = urlencode( $matches['username'] );
+			$url = "//profiles.wordpress.org/{$username}/profile/edit/group/3";
```
dd32, line 369:
```suggestion
			$url = "https://profiles.wordpress.org/{$username}/profile/edit/group/3";
```

I personally see no need for protocol-relative links anymore.

Hmm.. I wonder if this should utilise `get_edit_account_url()`? I'd like to think we'll make this `/group/3` stuff more pretty at some point..

```suggestion
			$url = get_edit_account_url( get_user_by( 'slug', $username ) );
```
(plus the function import into the file)

PR #369, opened 2024-09-15
Show submitted date in Plugin Controls

ernilambar:
Meta ticket: https://meta.trac.wordpress.org/ticket/7761

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/admin/metabox/class-controls.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/admin/metabox/class-controls.php
@@ -193,6 +193,11 @@ class Controls {
 			</tr>
 
 			<tr>
+				<td><?php _e( 'Submitted:', 'wporg-plugins' ); ?></td>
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/wp-content/plugins/plugin-directory/admin/metabox/class-controls.php
```diff
@@ -192,6 +192,11 @@ protected static function display_meta() {
 				<td><strong><?php printf( '<span title="%s">%s ago</span>', esc_attr( $post->last_updated ), human_time_diff( strtotime( $post->last_updated ) ) ); ?></strong></td>
 			</tr>
 
+			<tr>
+				<td><?php _e( 'Submitted:', 'wporg-plugins' ); ?></td>
+				<td><strong><?php printf( '<span title="%s">%s ago</span> (%s)', esc_attr( $post->_submitted_date ), human_time_diff( strtotime( $post->_submitted_date ) ), esc_html( gmdate( 'Y/m/d g:i a', $post->_submitted_date ?? 0 ) ) ); ?></strong></td>
```
tellyworth, line 197: I think the `strtotime()` is incorrect here; `_submitted_date` seems to be stored as a unix timestamp rather than a string like `last_updated`.

ernilambar: I was not able to setup proper WP local dev setup for "wordpress.org". So I prepared PR just by guessing variables.

ernilambar: Appreciate if you could edit the PR and update as required.

PR #380, opened 2024-09-30
Update plugins reviews component to shared blocks.

StevenDufresne:
This PR updates the ratings code for plugins to make sure of the [shared blocks](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks).

# Considerations

In this PR I decided to update the widget and not move it into the theme to maintain consistency with the other Widgets. It does result in an extra ratings call as both cases (theme filter and plugin) need to check for ratings.

 I'm not opposed to updating this PR and moving the code though.

#Screenshots

| Before | After |
|--------|--------|
| Screenshot 2024-09-30 at 1 49 43 PM | Screenshot 2024-09-30 at 1 49 11 PM |


ryelle:
> In this PR I decided to update the widget and not move it into the theme to maintain consistency with the other Widgets.

I think it would be better to use this as a way to start blockifying the sidebar properly. I actually had started some of this myself a few months ago (I stalled out trying to do the full sidebar, the logic is a little confusing). But if I scale it back and pull in your theme update, this is the result: https://github.com/ryelle/wordpress.org/commit/073276fa7f1a40f9ff7ea3e008c025129e7aea22

Screenshot 2024-10-01 at 4 54 32 PM

In that code, I've started a "sidebar" pattern that can be used to hold the rest of sidebar as the widgets are converted — "Meta" would be a good one to do next, since it can use the same basic `meta-list` structure that the themes block uses. Or the widgets can be blocks, ex `<!-- wp:legacy-widget {"idBase":"plugin_meta","instance":{}} /-->`

Feel free to grab my code, or I can make a PR with that branch I linked.

StevenDufresne:
Sure. I can get it started in this PR.

StevenDufresne:
The only thing missing from that code is the ability to show/hide the links under the reviews depending on whether reviews exists or the user is logged in.

That problem seems to exist in the theme directory as well.

Where would be the best place to add that logic you think?

## Theme with no ratings
Screenshot 2024-10-02 at 12 13 28 PM

ryelle:
> Where would be the best place to add that logic you think?

Perhaps `ratings-bars` should output a "no ratings yet" string, and the `ratings-link` block binding could check for reviews first.

StevenDufresne:
Deployed in https://meta.trac.wordpress.org/changeset/14140.

PR #393, opened 2024-10-03
Patterns API: Fix endpoint tests

ryelle:
We recently had an issue causing errors on client WP sites because the Patterns API returned malformed category data — see https://github.com/WordPress/pattern-directory/pull/711. If these tests were working and easy to run, it would have been caught earlier.

These tests function as e2e tests running against production data, checking that the output behaves as expected. This PR fixes the tests by updating the test cases to better match the real content, and to check the expected type of all patterns in a response.

Note: I'll also add a schema test to the pattern directory plugin, which could run as part of the CI, and catch this before it even merged.

**To test**

Requires a sandbox (I think?)

1. Get composer on your sandbox if you don't have it already (I just downloaded the `composer.phar` to my home directory, there are also brew instructions in the dotorg repo at `/wordpress/api/README.md`)
2. Install deps: `composer install`
3. Run the tests: `composer run test -- --group patterns`
4. The tests take about a minute, but all should pass except for one incomplete

The events tests are also failing, but that can be fixed in another PR.

### DIFF
--- a/api.wordpress.org/public_html/patterns/1.0/tests/test-index.php
+++ b/api.wordpress.org/public_html/patterns/1.0/tests/test-index.php
@@ -81,12 +81,12 @@ class Test_Patterns extends TestCase {
 	 * @group e2e
 	 */
 	public function test_browse_all_patterns() : void {
-		$response = send_request( '/patterns/1.0/?per_page=100' );
+		$response = send_request( '/patterns/1.0/?per_page=20' );
 		$this->assertResponseHasPattern( $response );
 
-		// When all locales and keywords are included, there should be at least 100 patterns.
+		// When all locales and keywords are included, there should be at least 20 patterns.
 		$patterns = json_decode( $response->body );
-		$this->assertSame( 100, count( $patterns ) );
+		$this->assertSame( 20, count( $patterns ) );
 
 		/*
 		 * The exact number of unique categories will vary based on which cohort of pattens happen to be returned,
@@ -108,7 +108,7 @@ class Test_Patterns extends TestCase {
 		 * This can't include a `pattern-keyword` param because of the workaround in
 		 * `WordPressdotorg\Pattern_Directory\Pattern_Post_Type\register_rest_fields()`.
 		 */
-		$response = send_request( '/patterns/1.0/?pattern-categories=' . $button_term_id . '&locale=en_US' );
+		$response = send_request( '/patterns/1.0/?per_page=20&pattern-categories=' . $button_term_id . '&locale=en_US' );
 		$this->assertResponseHasPattern( $response );
 
 		$patterns   = json_decode( $response->body );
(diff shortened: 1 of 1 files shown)


api.wordpress.org/public_html/patterns/1.0/tests/test-index.php
```diff
@@ -81,19 +83,12 @@ public function get_term_slugs( $patterns ) {
 	 * @group e2e
 	 */
 	public function test_browse_all_patterns() : void {
-		$response = send_request( '/patterns/1.0/?per_page=100' );
+		$response = send_request( '/patterns/1.0/?per_page=20' );
```
ryelle, line 86: Decreased this to 20 to prevent the API response from timing out. The number doesn't matter as much as getting a set number. 20 is also arbitrary.

api.wordpress.org/public_html/patterns/1.0/tests/test-index.php
```diff
@@ -297,7 +292,7 @@ public function data_search_patterns() {
 	 * @group e2e
 	 */
 	public function test_search_title_match_boosted_above_description_match() : void {
-		$search_term = 'image';
+		$search_term = 'heading';
```
ryelle, line 295: "image" no longer returns patterns without the term in the title, so the array is arbitrarily resorted (all sort weights are 0). Switching to "heading" returns some items without the term in the title, so the sorting works.

api.wordpress.org/public_html/patterns/1.0/tests/test-index.php
```diff
@@ -260,7 +255,7 @@ public function data_search_patterns() {
 
 			// The Core keyword (11) is hardcoded in `test_search_patterns()`, so don't need to specify it here.
 			"only match Core posts" => array(
-				'search_term'       => 'two buttons', // Post ID 727.
+				'search_term'       => 'Two images with text and buttons', // Post ID 727.
```
ryelle, line 258: "two buttons" matches a core pattern now.

api.wordpress.org/public_html/patterns/1.0/tests/test-index.php
pkevan: in the wpcut tests that were added some known patterns were used with structure that did break requests - I wonder if we should either do similar here, or have a mix of known and random patterns?

api.wordpress.org/public_html/patterns/1.0/tests/test-index.php
pkevan: does this cover: https://github.com/WordPress/wordpress.org/pull/393/files#r1792990280?

api.wordpress.org/public_html/patterns/1.0/tests/test-index.php
ryelle: > some known patterns were used with structure that did break requests

Which known patterns are breaking requests? do you mean the category_slugs thing? This particular test would not catch that, but the other tests (that run `assertResponseHasPattern`) will check individual pattern return types (like `test_results_limited_to_requested_locale` &  `test_browse_patterns_by_category`).

pkevan: it was pattern id `199` which caused the errors - unsure if we want to do it statically or not like: https://api.wordpress.org/patterns/1.0/?include=199,229080,446939,482686,278112

PR #395, opened 2024-10-13
Use sha256 hashing algorithm for gravatar

sy-records:


dd32:
Gravatar does now suggest using SHA256 rather than md5. https://docs.gravatar.com/api/avatars/hash/

sy-records:
@dd32 Are there any plans to merge?

dd32:
I was just thinking of that when I saw your last merge commit :) 
_Sidenote: We can't merge PRs here, so merging trunk doesn't speed it up, just reminds me this is still outstanding_

I think I'll wait for https://core.trac.wordpress.org/ticket/60638 to land such that the gravatar we use here will be the same as used elsewhere within WordPress.org/WordPress.

sy-records:
Thank you for reply, I got it!

PR #400, opened 2024-10-21
Allow some URL for the events in the robots.txt

### LINKED ISSUE #356: Social Sharing is broken because of robots.txt
We need Social networks to access these URLs despite everything else on translate.wordpress.org not being indexable:
See https://translate.wordpress.org/robots.txt

Screenshot 2024-09-17 at 17 30 22

Thanks @alexdeborba for reporting!

amieiro:
The https://translate.wordpress.org/robots.txt file is blocking some URL for the events, as you can see at https://github.com/WordPress/wporg-gp-translation-events/issues/356. This functionality blocks the posibility to correctly share the individual events in the social networks.

This PR enables 3 patterns in the robots.txt for https://translate.wordpress.org/ (https://translate.wordpress.org/robots.txt):
- `/events/rss` for the RSS feed.
- `/events/feed` for the RSS feed.
- `/events/*/*` for 3-level URLs like https://translate.wordpress.org/events/2024/wordpress-translation-week-2024-online/

Fixes https://github.com/WordPress/wporg-gp-translation-events/issues/356

### DIFF
--- a/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-robots.php
+++ b/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-robots.php
@@ -18,7 +18,10 @@ function wporg_robots_txt( $robots ) {
 		           "Allow: /locale/$\n" .
 		           "Allow: /locale/*/glossary/$\n" .
 		           "Allow: /locale/*/stats/plugins/$\n" .
-		           "Allow: /locale/*/stats/themes/$\n";
+		           "Allow: /locale/*/stats/themes/$\n" .
+		           "Allow: /events/rss\n" .
+		           "Allow: /events/feed\n" .
+		           "Allow: /events/*/*\n";
 
 	} elseif ( 'wordpress.org' === $blog_details->domain ) {
 		// WordPress.org/search/ should not be indexed.


wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-robots.php
```diff
@@ -18,7 +18,10 @@ function wporg_robots_txt( $robots ) {
 		           "Allow: /locale/$\n" .
 		           "Allow: /locale/*/glossary/$\n" .
 		           "Allow: /locale/*/stats/plugins/$\n" .
-		           "Allow: /locale/*/stats/themes/$\n";
+		           "Allow: /locale/*/stats/themes/$\n" .
+		           "Allow: /events/rss\n" .
+		           "Allow: /events/feed\n" .
+		           "Allow: /events/*/*\n";
```
akirk, line 24: Would this also work? Since we'll want to allow everything anyway:
```suggestion
		           "Allow: /events/\n";
```

amieiro: The idea with these 3 lines is to enable the RSS feed and the events, but keeping disabled the other URLs, such as edit, delete,...

dd32: Since Edit/Delete shouldn't be accessible to robots (which are logged out) and shouldn't be returning `200` responses, it's probably safe to just allow `/events/`

amieiro:
Committed [here](https://meta.trac.wordpress.org/changeset/14138).

PR #407, opened 2024-10-28
Add a release page

StevenDufresne:
Related https://meta.trac.wordpress.org/ticket/7783.

This PR adds the "Releases" tab behind a feature flag `show_release_beta` so we can work together on the site. 

**TODO:**
Everything seems to be working properly, however, there seems to be a race condition where the first view doesn't reliably display. You can reproduce it sometimes by loading the release page often. The content will not be displayed.


tellyworth:
@StevenDufresne I think the CPT code is kinda-sorta functional, thought not really hooked up to anything yet.

I've added a bin script you can use to (re)generate CPT entries for a specific plugin for testing purposes:

```
php bin/update-release-cpt.php --plugin=debug-bar
```

Next step is to properly hook this up to the main Plugin Directory class, and add some internal and REST APIs.

tellyworth:
With the last commit [f5cc81a](https://github.com/WordPress/wordpress.org/pull/407/commits/f5cc81acb3d27296637e9ef6b70eafc4e7c40520) I was thinking about how to backfill in production, and how to eventually deal with migrating and replacing the `releases` postmeta.

Assuming we want to replace the postmeta with CPTs, it would be fairly straightforward to replace functions like `Plugin_Directory::get_releases()` with wrappers for `Plugin_Release::get_releases()` etc. And adapt  `Plugin_Directory::prefill_releases_meta()` to backfill CPTs from svn data. Then update anything that depends on the postmeta (like Release Confirmations) to work with a post object rather than the simple postmeta array format.

The main difficulty with that is migrating in such a way that everything continues to work smoothly - ie plugins that haven't been backfilled smoothly can still have releases confirmed.

I think the best way to manage that would be:

1. Have the import code backfill CPTs on the first trunk commit for each plugin. That way we'll gradually backfill actively developed plugins first. (Already done in f5cc81a)
2. When we're happy that the backfill data is good, run a one-time script to backfill other plugins. Could be done from a sandbox.
3. Then, replace the code to swap in CPTs for postmeta as outlined above.

I guess I could write and test a gated version of (3) so that it's ready to go and we just have to flip a switch.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-release.php
```diff
@@ -0,0 +1,420 @@
...
+			} else {
+				// If we don't have revision numbers, use dates. Maybe this should be removed.
+				if ( strtotime( $release['date'] ) <= strtotime( $last_release->release_date ) ) {
+					return false; // Not an error, just skip.
+				}
+			}
+		}
+
+		// Store the commit log in postmeta. We'll only do this for drafts.
+		$last_release_revision = max( $last_release->release_revision );
```
StevenDufresne: @tellyworth I think this block needs to be inside of the `if ( $last_release )`. I just ran this with a new plugin and this errored out.

StevenDufresne: My assumption is incorrect. First time drafts won't ever receive commits.

StevenDufresne: Okay, so it looks like we could pass `SVN::log( $trunk_url, [ 1, 'HEAD' ] )` to get all the commits for new plugins without releases. 

However, what happens when we roll this out for old plugins that used trunk. We would pull in the whole plugin commit log. Do we care if we show something like that denotes 100 commits:

Screenshot 2024-12-11 at 3 55 31 PM

To those users those commits are not new, however, it could be argued that they are new to the tag/release.

tellyworth: Yeah [1, 'HEAD'] will probably work for the initial version or where no release exists.

How about if I limit the commit log meta to the most recent 5 or 10 items for each release?

StevenDufresne: I think that would be less expected and maybe more confusing...? I think we maybe just list them all and see what users say?

mahmudurrahmanlabib (approved):

PR #408, opened 2024-10-28
Replace support review component with shared one.

StevenDufresne:
This PR updates the support forum reviews stars & bars to use the shared component.

Related: https://github.com/WordPress/wporg-mu-plugins/pull/633, https://github.com/WordPress/wordpress.org/pull/380

## Important

The forums display more precise rating numbers, while the new components round ratings to the nearest 0.5. Since we already use the new components in multiple places, if we want to keep the current precision, we should do it everywhere at once in the new components.

## Considerations
The `ratings-stars` and `ratings-bars` components [expect](https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/blocks/ratings-stars/render.php#L3) the context to contain `postId`. However, based on how bbPress works and how it's been modified, the `postId` is not set. 

As a result, I've made use of the `render_block_context` block to set it. we still don't use that in `wporg_set_rating_data`, as we need to call `wporg_support_get_compat_object` to get all the relevant data. It feels a bit silly, but feels better than the alternative. The alternative would be to change the blocks to not return if the post_id is not set. That feels less predictable.

I've also done some minor styling updates.

## Screenshots

| Before | After |
|--------|--------|
| Screenshot 2024-10-25 at 12 08 11 PM | Screenshot 2024-10-25 at 11 52 50 AM |
| Screenshot 2024-10-25 at 11 49 34 AM | Screenshot 2024-10-25 at 11 53 09 AM  |

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-ratings-compat.php
+++ b/wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-ratings-compat.php
@@ -212,18 +212,7 @@ class Ratings_Compat {
 		add_filter( 'bbp_get_topic_last_topic_title', array( $this, 'undo_topic_title' ), 10, 1 );
 ?>
 <div class="review-ratings">
-	<div class="col-3">
-		<div class="reviews-about" style="display:none;"><?php echo esc_html( $this->object->post_title ); ?></div>
-		<div class="reviews-total-count"><?php
-			printf(
-				/* translators: %s: number of reviews */
-				_n( '%s review', '%s reviews', $this->reviews_count, 'wporg-forums' ),
-				'<span>' . number_format_i18n( $this->reviews_count ) . '</span>'
-			);
-		?></div>
-		<?php echo do_blocks( '<!-- wp:wporg/ratings-bars /-->' ); ?>
-	</div>
-	<div class="col-5">
+	<div>
 		<div style="font-weight:bold;"><?php _e( 'Average Rating', 'wporg-forums' ); ?></div>
 		<?php echo do_blocks( '<!-- wp:wporg/ratings-stars /-->' ); ?>
 		<div class="reviews-submit-link">
@@ -251,6 +240,17 @@ class Ratings_Compat {
 			}
 		?>
 		</div>
+	</div>
+	<div>
+		<div class="reviews-about" style="display:none;"><?php echo esc_html( $this->object->post_title ); ?></div>
+		<div class="reviews-total-count"><?php
+			printf(
+				/* translators: %s: number of reviews */
+				_n( '%s review', '%s reviews', $this->reviews_count, 'wporg-forums' ),
+				'<span>' . number_format_i18n( $this->reviews_count ) . '</span>'
+			);
+		?></div>
+		<?php echo do_blocks( '<!-- wp:wporg/ratings-bars /-->' ); ?>
 	</div>
 </div>
 		<?php
--- a/wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-ratings-compat.php
+++ b/wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-ratings-compat.php
@@ -221,56 +221,11 @@ class Ratings_Compat {
 				'<span>' . number_format_i18n( $this->reviews_count ) . '</span>'
 			);
 		?></div>
-		<?php
-			foreach ( array( 5, 4, 3, 2, 1 ) as $rating ) {
-				$ratings_count = isset( $this->ratings_counts[ $rating ] ) ? $this->ratings_counts[ $rating ] : 0;
-				$ratings_count_total = isset( $this->ratings_counts ) ? array_sum( $this->ratings_counts) : 0;
-				$stars_title = sprintf(
-					/* translators: %s: number of stars */
-					_n(
-						'Click to see reviews that provided a rating of %d star',
-						'Click to see reviews that provided a rating of %d stars',
-						$rating,
-						'wporg-forums'
-					),
-					$rating
-				);
-				/* translators: %d: number of stars */
-				$stars_text = sprintf(
-					/* translators: %d: number of stars */
-					_n( '%d star', '%d stars', $rating, 'wporg-forums' ),
-					$rating
-				);
-				$width = 0;
-				if ( $ratings_count && $ratings_count_total ) {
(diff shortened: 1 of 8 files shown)


wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/inc/block-config.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $compat_object = wporg_support_get_compat_object();
+
+        if ( $compat_object ) {
+            $context = array_merge( $context, [ 'postId' => $compat_object->ID ] );
+        }
+    }
+
+    return $context;
+}
```
ryelle: Super minor, but my editor flagged it— this section is using spaces, not tabs.

wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/inc/block-config.php
```diff
@@ -0,0 +1,68 @@
+<?php
+/**
+ * Set up configuration for dynamic blocks.
+ */
+
```
ryelle, line 5: Should this file be namespaced? If not, maybe the function prefixes should be `wporg_support_`?

wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-ratings-compat.php
```diff
@@ -221,56 +221,11 @@ public function do_view_header() {
...
-		<?php
-			echo \WPORG_Ratings::get_dashicons_stars( $this->avg_rating );
-			printf(
-				/* translators: 1: number of stars in rating, 2: total number of stars (5) */
-				__( '%1$s out of %2$s stars', 'wporg-forums' ),
-				round( isset( $this->avg_rating ) ? $this->avg_rating : 0, 1 ),
-				'<span>5</span>'
-			);
-		?>
+		<?php echo do_blocks( '<!-- wp:wporg/ratings-stars /-->' ); ?>
```
ryelle, line 217: Can these two blocks be switched in the markup, rather than reversed in CSS? Currently that's causing awkward tab order, where you tab through the ratings rows, then jump back to the "Add your own…" button.

ryelle:
> The forums display more precise rating numbers, while the new components round ratings to the nearest 0.5. Since we already use the new components in multiple places, if we want to keep the current precision, we should do it everywhere at once in the new components.

I remembered [this discussion](https://github.com/WordPress/wporg-theme-directory/issues/147) — it was already like this in the old theme, so this is just fixing that inconsistency. I have no strong preference on which precision is better, but if we want to change it, we can just transfer that linked issue to wporg-mu-plugins and update the `wporg/ratings-stars` block.

> As a result, I've made use of the render_block_context block to set it. we still don't use that in wporg_set_rating_data, as we need to call wporg_support_get_compat_object to get all the relevant data. It feels a bit silly, but feels better than the alternative. The alternative would be to change the blocks to not return if the post_id is not set. That feels less predictable.

Technically post_id is used in the filter, so it could be used, though it isn't used _here_. You could just return something truthy from `wporg_render_block_context` if there's any concern about the performance of calling `wporg_support_get_compat_object` twice.

wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/inc/block-config.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @return array
+ */
+function wporg_set_rating_data( $data, $post_id ) {
+	$post = wporg_support_get_compat_object();
+
+	if ( class_exists( '\WPORG_Ratings' ) ) {
+		$rating  = \WPORG_Ratings::get_avg_rating( $post->type, $post->post_name ) ?: 0;
+		$ratings = \WPORG_Ratings::get_rating_counts( $post->type, $post->post_name ) ?: array();
+	}
```
ryelle, line 27: What happens if the class isn't found?

ryelle (commented):
Looks good, just a few little notes.

wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/inc/block-config.php
StevenDufresne: This theme isn't generally using namespacing, but there are a few references around. I'll use it.

wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/inc/block-config.php
StevenDufresne: Oof. It goes boom. 😮‍💨

wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/inc/block-config.php
StevenDufresne: Hmm, back-and-forth pasting isn't a good workflow.

wordpress.org/public_html/wp-content/plugins/support-forums/inc/class-ratings-compat.php
StevenDufresne: I'll take a look.

wordpress.org/public_html/wp-content/themes/pub/wporg-support-2024/inc/block-config.php
```diff
@@ -0,0 +1,72 @@
...
+ * @param array $data    Rating data.
+ * @param int   $post_id Current post.
+ *
+ * @return array
+ */
+function wporg_set_rating_data( $data, $post_id ) {
+	$post = wporg_support_get_compat_object();
+
+	if ( ! class_exists( '\WPORG_Ratings' ) ) {
+		return '';
```
ryelle: This function should return an array — the empty string does work since the filter value is run through `wp_parse_args`, but to match the function docs it should be an array.

```suggestion
		return [];
```

ryelle: Actually, this could also return $data, but that might be less clear?

```suggestion
		return $data;
```

ryelle (approved):
Looks good! One more minor comment about return types, feel free to grab either code suggestion 🙂

PR #471, opened 2025-02-10
Move forum column titles to wporg-forums text domain

### LINKED ISSUE #470: Move forum column titles to wporg-forums text domain
The column titles are used form bbPress. But Participants is not used there.

https://de.wordpress.org/support/view/all-topics/
Image

timse201:
fixes: #470 
All strings are already used in loop-reply-topics.php


timse201:
https://meta.trac.wordpress.org/ticket/7911

PR #483, opened 2025-04-09
Restrict self-removal for plugin committers.

sebastienserre:
Updated logic to prevent committers from removing themselves. This ensures better user management.

Should fix https://meta.trac.wordpress.org/ticket/7947


dd32:
I don't think this is the optimal route, since there are times when you want to remove yourself from a plugin.

I'd probably only restrict it when you're the owner of the plugin, or you're the only committer of the plugin.

Perhaps this would be better implemented through a modal warning that they're about to remove themselves.

PR #485, opened 2025-04-11
Forums - some more space for reply column

timse201:
Adds some more space to the reply column to fit translations.

Trac ticket: [#7951](https://meta.trac.wordpress.org/ticket/7951)


bramburn (approved):
Looks good

Issue #481, opened 2025-04-03
wporg-plugins-2024: Link text to view all reviews is difficult to localize


t-hamano:
There is a link to view all reviews in the sidebar of the theme/plugin page:

![Image](https://github.com/user-attachments/assets/b0451519-6cda-407f-aa9a-6ea333428ea6)

This link is generated here:

https://github.com/WordPress/wordpress.org/blob/82d3f79f21b7a776073d1f36a26c98b4851ebd35/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/block-bindings.php#L43-L47

The generated HTML is as follows:

```html
<a href="">See all<span class="screen-reader-text"> reviews</span></a>
```

This sentence consists of the following words:

- `See`: verb
- `All`: adjective
- `Reviews`: noun

The word `reviews` is invisible because of the CSS class for screen readers, so the visible sentence is `See all`, which consists of `verb + adjective`. In English, a sentence may be natural without a noun, but in other languages, this may not be the case.

In some languages, the sentence that represents `verb + adjective` may change if the noun is removed. Therefore, the current implementation makes it hard to provide a correct translated text for both the visible sentence and the sentence for screen readers. For this reason, I recommend not to hide a word from a sentence.

I recommend providing a consistent sentence "See all reviews", but if localization may cause conflicts with the "Add my review" text on the left, it may be good to add `flex-wrap: wrap` to the parent Group block.

P.S. We may be tempted to use the `title` attribute to solve this issue like the following example, but note that this is not recommended from an a11y perspective:

```html
<a href="" title="See all reviews">See all</a>
```

Ref: https://make.wordpress.org/accessibility/handbook/best-practices/quick-start-guide/#links

badembafr:
Hello,

If I understood correctly, the idea is to display the full sentence, like "See all reviews", without using screen-reader-text.

Also, we need to allow line breaks with `flex-wrap: wrap` to prevent the text from overflowing or overlapping.

Thanks in advance!

badembafr:
Hello,

I have opened a pull request to address this issue: [#489](https://github.com/WordPress/wordpress.org/pull/489)
It improves the localization of the “See all reviews” link.
Let me know if any adjustments are needed!

Thanks in advance 🙂

mathetos:
This also seems to be ready to be closed based on the conversation in the PR #489 

But in looking into this it surfaced another translation bug I reported in #792 

[...]

PR #489, opened 2025-05-24
Fix: Make 'See all reviews' link fully visible for better localization

badembafr:
**Fix localization of "See all reviews" link by removing hidden screen-reader text**

Hello all,

This pull  request addresses the issue described here:
[https://github.com/WordPress/wordpress.org/issues/481](https://github.com/WordPress/wordpress.org/issues/481)

**What I changed :**
I replaced the original translation string:

```php
__( 'See all<span class="screen-reader-text"> reviews</span>', 'wporg-themes' )
```

with:

```php
__( 'See all reviews', 'wporg-themes' )
```

**Why this change is necessary :**
This improves localization and  accessibility.
In French for example,  the previous string would result in something like:

> "Voir tous les"

Which is incomplete and incorrect, because "reviews" was hidden using the `screen-reader-text` class.
After this change, the sentence becomes:

> "Voir tous les avis"

Which is grammatically correct and understandable for all users.

**Note:**
I'm not sure whether you would also recommend adding `flex-wrap: wrap` on the parent container to prevent layout issues on smaller screens. Let me know if you'd like me to add that as well.

Thanks in advance !

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/block-bindings.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/block-bindings.php
@@ -43,7 +43,7 @@ function get_meta_block_value( $args, $block ) {
 			return sprintf(
 				'<a href="%s">%s</a>',
 				esc_url( 'https://wordpress.org/support/plugin/' . $plugin_post->post_name . '/reviews/' ),
-				__( 'See all<span class="screen-reader-text"> reviews</span>', 'wporg-themes' )
+				__( 'See all reviews', 'wporg-themes' )
 			);
 		case 'submit-review-link':
 			return sprintf(


dd32:
This PR changes the link display for all users, where as I believe the string is in it's current state because it was requested to be shorter for design purposes (while the `reviews` additional was added for accessibility purposes).

I'm in support of changing this though, perhaps to a screen reader vs visual representation through something like:
```
<span aria-hidden="true">See all</span>
<span class="screen-reader-text">See all reviews</span>
```
(Both internal strings would be translated without the <span>)
That could be done through two replacements:
```
sprintf(
'<a href="%s">%s%s</a>',
...,
'<span aria-hidden="true">' . __()....,
'<span class="screen-reader-text">' . __()....,
)
```

badembafr:
Ok I will try to fix this problem.

Thanks !

t-hamano:
Thanks for the PR.

> perhaps to a screen reader vs visual representation through something like

This certainly seems like the best approach.

mahmudurrahmanlabib (approved):

timse201:
I don't like the idea.

image

For German it is: Alle <span class="screen-reader-text">Rezensionen</span> anzeigen
The double space is **auto corrected**.

This way we force every language to translate it again and force most of us to respect the line break. It is not necessary. If a single language has problems - just remove that tag. We don't need to force every language and the design team will not like line breaks.

t-hamano:
I looked into this issue again and found that, at least for the Japanese locale, I was able to translate it in a way that was both visually appropriate and accessibility-appropriate by adjusting the translation method.

That is, I changed the text inside the span tag itself to match the language.

Before:
```
すべての<span class="screen-reader-text">レビュー</span>を見る
```

After:
```
すべてのレビュー<span class="screen-reader-text">を見る</span>
```

I'm not sure if this approach will work for all other languages, but if there are no issues in other languages, I think it's okay to close this.

obenland:
Closing as "worksforme".

PR #498, opened 2025-07-09
Phased rollout of plugins

dd32:
See https://meta.trac.wordpress.org/ticket/8009

NOTE: This branch/PR is being used as a staging ground, and being committed in pieces to WordPress.org, before being merged into this branch from trunk.

The changes present in the PR diff are changes that are not yet deployed to production.


wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/template-tags.php
```diff
@@ -772,3 +775,64 @@ function the_author_notice( $post = null ) {
 		);
 	}
 }
+
+/**
+ * Displays the "phased rollout" settings for a plugin.
+ */
+function the_phased_rollout_settings() {
```
swissspidy: Could this actually be tied to a release confirmation so you can configure this _per release_?

Major updates will have different requirements than minor or patch updates.

dd32: Yes :) This is actually suggested in further changes like 99404d64c072f7383d6f063fbc400cd1772ef607 and https://github.com/WordPress/wordpress.org/pull/498/files#diff-e129d0aa1e6f1607ca89d58a3fddff3eca7109aa2419e3d498e3ddd57fba8e9b which makes  the per-plugin just the fallback

I've considered making this a release-confirmation-only feature too, but I don't think there's much of a benefit to doing that, due to the lack of interest in using it by many.

PR #506, opened 2025-09-01
Theme Directory: upgrade blueprint

dd32:
See https://meta.trac.wordpress.org/ticket/8077

This updates the preview endpoint to handle theme previews for themes.trac, and adds a few options.

This is still in need of a cleanup I suspect, so PR'ing it for now.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/theme-directory/rest-api/class-theme-preview.php
+++ b/wordpress.org/public_html/wp-content/plugins/theme-directory/rest-api/class-theme-preview.php
@@ -57,8 +57,8 @@ class Theme_Preview {
 				),
 			),
 			'permission_callback' => function( $request ) {
-				$theme_data    = wporg_themes_theme_information( $request['slug'] ?? '' );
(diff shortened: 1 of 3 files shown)


wordpress.org/public_html/wp-content/plugins/theme-directory/rest-api/class-theme-preview.php
```diff
@@ -7,18 +7,69 @@
...
+					'sanitize_callback' => 'rest_sanitize_boolean',
+				),
+				// Whether to install the Theme Previewer plugins.
+				'theme_previewer' => array(
+					'type'              => 'boolean',
+					'required'          => false,
+					'default'           => true,
+					'sanitize_callback' => 'rest_sanitize_boolean',
+				),
+			),
```
dd32: I think these should be removed, and instead a new endpoint of `/theme-review-preview/SLUG/VERSION` be added instead. That can build off the rest of this PR changes but would give future stability for older trac ticket links.

PR #521, opened 2025-10-24
streamline forums home labels

### LINKED ISSUE #520: streamline forums navigation home labels
We have 2 labels for the forum's main page. Home and Forums. Home is used for breadcrumbs. We should move Home to Forums for better accessibility and to avoid incorrect/misleading translations. So users know directly that they will lead to the forum's home URL.

Image

timse201:
fixes #520


dd32:
Thanks @timse201! This change seems reasonable to me.

PR #531, opened 2025-11-14
Events API: Add click tracking to generate stats of meetup interest

dd32:
This PR adds anonymous click-tracking to the events api output.

The URLs are passed through an intermediatory URL as such
> `https://api.wordpress.org/events/redirect/?meetup=311953935&url=https://www.meetup.com/wordpress-brisbane/events/311953935/&ref=core`

which stores the number of redirect-requests against the event in the database.

Additionally, some basic stats (type and country) is stored in another anonymous dataset.

This allows us to see how much event reach we're getting through our dashboard widget, and potentially feedback on which topics are more interesting to local communities.

To allow for edge-cases, the URL is included in the tracked url, which if the hostname matches a safe-list will allow a redirect to that location even if our DBs are offline or the table is empty. _(This would allow redirection of non-chapter meetup groups as a result, but I don't see a meetup.com redirect as unsafe)_

The main downside to this is that the URL seen in the WordPress dashboard will be the click-tracked url, and it's not a pretty-url :) I don't personally see this as a problem, but some might.

### DIFF
--- a/api.wordpress.org/public_html/events/1.0/index.php
+++ b/api.wordpress.org/public_html/events/1.0/index.php
@@ -908,7 +908,7 @@ function get_events( $args = array() ) {
 
 	$raw_events = $wpdb->get_results( $wpdb->prepare(
 		"SELECT
-			`type`, `title`, `url`,
+			`type`, `source_id`, `title`, `url`,
 			`meetup`, `meetup_url`,
 			`date_utc`, `date_utc_offset`, `end_date`,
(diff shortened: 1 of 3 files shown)


dd32:
One benefit of adding such tracking is that it will hide the WordPress site as the referer in traffic to Meetup.com, setting api.wordpress.org as the referer. Whether that's a benefit to everyone is debatable :)

PR #532, opened 2025-12-10
Plugin Directory: Use secondary style for "Discard" release button in confirmation UI

manhphuc:
This change styles the destructive “Discard” release button as a secondary button, so it no longer appears the same as the primary “Confirm” action.

It keeps the existing behaviour intact and only adjusts the visual hierarchy of the actions.

Fixes: https://meta.trac.wordpress.org/ticket/8142


dd32 (changes requested):
PR does nothing.

Screenshot 2025-12-10 at 3 53 53 pm

manhphuc:
Thanks for the feedback!

You were absolutely right — my initial change only updated the class and didn't adjust the visual styling.

I've updated the PR to:

- Add a dedicated `.discard-release` style to the `wporg-plugins-2024` theme so the “Discard” button correctly appears as a secondary/outline action.
- Keep the “Confirm” button as the primary action.

The “Discard” button now reflects the intended secondary hierarchy in the release confirmation dialog.  
Happy to refine the styling or selectors further if needed!

PR #534, opened 2025-12-13
Fix: Support link padding

Takshil-Kunadia:
### Description
Fixes the Support link/button having incorrect padding on plugin pages.

**Trac Ticket**: https://meta.trac.wordpress.org/ticket/8147

### Problem
The "Support" link on plugin pages (e.g., https://wordpress.org/plugins/sync-content-from-notion) had asymmetric padding, with the right padding missing. This caused the active state to appear visually unbalanced.

| Before | After |
|--------|--------|
| support-button-defect | Screenshot 2025-12-13 at 1 57 15 PM |

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/client/components/plugin/style.scss
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/client/components/plugin/style.scss
@@ -284,7 +284,7 @@
 
 		a {
 			display: inline-block;
-			padding: 0.64rem 0 0.64rem 1.25rem;
+			padding: 0.64rem 1.25rem 0.64rem 1.25rem;
 			font-size: var(--wp--preset--font-size--normal);
 		}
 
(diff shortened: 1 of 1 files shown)


manhphuc:
Tested on a plugin page and confirmed the Support link padding looks consistent in the active state now.

Minor suggestion: the padding could be simplified to `padding: 0.64rem 1.25rem;` for readability (same computed values).

Takshil-Kunadia:
Thanks @manhphuc for looking into this. Made the change. ✅

Takshil-Kunadia:
Hi @dd32, just checking in on this one, happy to rebase or make any changes if needed! 🙂

dd32:
As this dates back to http://meta.trac.wordpress.org/changeset/5399 I feel like this is intentional.

While the screenshot above clearly shows the padding would be better equal, it's not in the context of the greater page, the purple box added clear shows that the text is right aligned; except the visual download button which is aligned with the text indented like a button should be.  The support **link** is a link, and not a button.

Screenshot 2026-03-02 at 2 12 45 pm

Takshil-Kunadia:
Thanks for the context, @dd32 ! I see your point about the link vs. button distinction. Happy to close this if the current behavior is by design, or adjust the approach if there's a middle ground as I think the "onclick" behaviour looks a bit out of place.

obenland:
Closing, as this is by design.

PR #551, opened 2026-02-11
Add application password authorization flow to wp-login.php

obenland:
## Summary

Adds an `authorize_application` action to `wp-login.php` that lets users create application passwords through the login flow. Requests are validated against a UUID-based allowlist. After approval, the MCP client configuration is rendered with a copy button for easy setup.

Not sure if this UUID-gating is strictly needed but I thought I'd be a bit conservative to start out with.
I suppose at some point we'd want an application password management UI in profiles.wordpress.org.

### Login prompt
When visiting the authorization URL while logged out, a contextual message is shown:

Screenshot 2026-02-11 at 2 04 26 PM

### Authorization form
After logging in, the user can approve or reject the connection:

Screenshot 2026-02-11 at 2 04 48 PM

### MCP configuration
On approval, the ready-to-use MCP client config is displayed with per-client setup instructions:

Screenshot 2026-02-11 at 2 04 55 PM

## TODO

- [x] Gate to `login.wordpress.org`.
- [x] Only show MCP config when creating app password for the MCP app.

## Test plan

- [ ] Visit `wp-login.php?action=authorize_application&app_id=c4c73a54-96d7-47b9-9bdc-1a66b9b04505` while logged out and verify redirect to login form with contextual message.
- [ ] Log in and verify the authorization form renders correctly.
- [ ] Approve and verify the MCP configuration is displayed with the correct credentials.
- [ ] Copy the config and verify it works with an MCP client.
- [ ] Reject and verify redirect to wp-admin.
- [ ] Verify that an invalid or missing `app_id` is rejected.
- [ ] Verify that providing `success_url` or `reject_url` for the MCP app (which has empty hosts) is rejected.
- [ ] Re-approve and verify the previous application password is revoked before creating a new one.


obenland:
@dd32 When you get a chance, I'd appreciate any feedback you have on this. Naming, file location, stuff I missed, anything. Thank you!

wordpress.org/public_html/wp-content/mu-plugins/pub/authorize-application-login.php
```diff
@@ -0,0 +1,588 @@
...
+					$error = $created;
+				} else {
+					list( $new_password ) = $created;
+
+					if ( $success_url ) {
+						$redirect = add_query_arg(
+							array(
+								'site_url'   => urlencode( 'https://wordpress.org' ),
+								'user_login' => urlencode( wp_get_current_user()->user_login ),
+								'password'   => urlencode( $new_password ),
```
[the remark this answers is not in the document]
obenland: dc8ccdb

wordpress.org/public_html/wp-content/mu-plugins/pub/authorize-application-login.php
```diff
@@ -0,0 +1,588 @@
...
+			} else {
+				foreach ( array( 'success_url', 'reject_url' ) as $key ) {
+					if ( empty( $request[ $key ] ) ) {
+						continue;
+					}
+
+					$host = wp_parse_url( $request[ $key ], PHP_URL_HOST );
+					if ( ! $host || ! in_array( $host, $allowed_hosts, true ) ) {
+						$error->add( 'unauthorized_redirect', __( 'The callback URL does not point to an allowed domain.' ) );
+					}
```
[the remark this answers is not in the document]
obenland: dc8ccdb

wordpress.org/public_html/wp-content/mu-plugins/pub/login-application-passwords.php
```diff
@@ -0,0 +1,597 @@
...
+
+		$app_id      = sanitize_text_field( wp_unslash( $_POST['app_id'] ) );
+		$success_url = sanitize_url( wp_unslash( $_POST['success_url'] ) );
+		$reject_url  = sanitize_url( wp_unslash( $_POST['reject_url'] ) );
+		$redirect    = '';
+
+		// Re-validate POST values (don't trust hidden form fields).
+		$user         = wp_get_current_user();
+		$allowed_apps = wporg_get_allowed_apps();
+		$app_name     = isset( $allowed_apps[ $app_id ] ) ? $allowed_apps[ $app_id ]['name'] : '';
```
dd32: Let's use null-coalesce when possible.
```suggestion
		$app_name     = $allowed_apps[ $app_id ]['name'] ?? '';
```

wordpress.org/public_html/wp-content/plugins/login-application-passwords/login-application-passwords.php
```diff
@@ -0,0 +1,597 @@
...
+					}
+				}
+			}
+		}
+
+		if ( $redirect ) {
+			// Explicitly not using wp_safe_redirect b/c sends to arbitrary domain.
+			wp_redirect( $redirect );
+			exit;
+		}
```
dd32, line 192: Is this how the Core WordPress endpoint works? Passing the app password via a GET value in a redirect? 

If so, I guess it's fine.. Not ideal, but fine.

This should really be done via a POST to the success url at a minimum, or preferably a temporary token that the client then uses to fetch a password... hmm.. I guess due to the hashed passwords that would be more complicated 🤷 

The OAuth Implicit flow I think uses `#....` values for this kinda thing, that still isn't recommended anymore, but for non-server-to-server communications is probably better than this..

Again, not a blocker.

wordpress.org/public_html/wp-content/mu-plugins/pub/login-application-passwords.php
```diff
@@ -0,0 +1,597 @@
...
+		$reject_url = sanitize_url( wp_unslash( $_REQUEST['reject_url'] ) );
+	} elseif ( $success_url ) {
+		$reject_url = add_query_arg( 'success', 'false', $success_url );
+	} else {
+		$reject_url = null;
+	}
+
+	$user         = wp_get_current_user();
+	$allowed_apps = wporg_get_allowed_apps();
+	$app_name     = isset( $allowed_apps[ $app_id ] ) ? $allowed_apps[ $app_id ]['name'] : '';
```
dd32:
```suggestion
	$app_name     = $allowed_apps[ $app_id ]['name'] ?? '';
```

wordpress.org/public_html/wp-content/plugins/login-application-passwords/login-application-passwords.php
```diff
@@ -0,0 +1,597 @@
...
+				} else {
+					list( $new_password ) = $created;
+
+					if ( $success_url ) {
+						$redirect = add_query_arg(
+							array(
+								'site_url'   => 'https://wordpress.org',
+								'user_login' => wp_get_current_user()->user_login,
+								'password'   => $new_password,
+							),
```
dd32, line 173: Pretty sure this is supposed to have `urlencode()` before being passed to `add_query_arg()`. Not 100% on that, and It's probably not needed with the values that would be present either.

wordpress.org/public_html/wp-content/mu-plugins/pub/login-application-passwords.php
```diff
@@ -0,0 +1,597 @@
+<?php
+/**
+ * Plugin Name: Application Password Authorization on Login
+ * Description: Provides the application password authorization flow on wp-login.php.
+ */
+
+declare( strict_types = 1 );
+
+// Only load on login.wordpress.org.
+if ( get_current_blog_id() !== 350 ) {
```
dd32:
```suggestion
// Limit to the login site on WordPress.org. Intentionally available to all sites in local environments.
if ( defined( 'WPORG_LOGIN_REGISTER_BLOGID' ) && get_current_blog_id() !== WPORG_LOGIN_REGISTER_BLOGID ) {
```

dd32 (commented):
As for code location.. I don't really mind where this lives. This seems like something that could be upstreamed to WordPress core perhaps, as a front-end UI for the existing wp-admin authorize application flow.  It seems this has a few things in it that are very WordPress.org specific, and that's OK.

wporg-sso plugin has a bunch of logic related to login in it. I don't really like that.
wporg-login theme has a bunch of logic in it. I also don't really like that.

I assume this is intentionally open to accepting literally any redirect location at present, and that we might look at changing it to a more strict host set before deploy? Otherwise, what's the point of an application allow list.. the UUIDs would be public knowledge..

wordpress.org/public_html/wp-content/plugins/login-application-passwords/login-application-passwords.php
obenland: > Is this how the Core WordPress endpoint works? 

[It is, yeah](https://github.com/WordPress/wordpress-develop/blob/6.9.1/src/wp-admin/authorize-application.php#L58-L62). It's pretty permissive. I guess the onus is on users _really_ understanding who they're sharing their access with.

obenland:
> I assume this is intentionally open to accepting literally any redirect location at present

Each app in the allowlist has a hosts array that controls which domains `success_url`/`reject_url` may point to. The MCP app has  an empty hosts array, which means no callback URLs are accepted at all (the validation rejects any URL when hosts is empty). A future app that needs redirects would specify its allowed domains there. Knowing the UUID shouldn't let you redirect anywhere.

wordpress.org/public_html/wp-content/plugins/login-application-passwords/login-application-passwords.php
obenland: Fixed in [202296c](https://github.com/WordPress/wordpress.org/pull/551/commits/202296cc2d14a43b8c2b79ef3e48b779d89c9834)

dd32:
> The MCP app has an empty hosts array, which means no callback URLs are accepted at all (the validation rejects any URL when hosts is empty).

@obenland OH! I didn't realise it just then prompted, I was thinking it just allowed an open-redirect with no hosts 🙃  which as you can imagine... concerned me :)

obenland:
Yeah, I think that's how the current Core version works. The form shows the URL that it'll redirect to, but it mostly seems to rely on users knowing what they're doing

wordpress.org/public_html/wp-content/mu-plugins/pub/login-application-passwords.php
```diff
@@ -0,0 +1,597 @@
...
+ *
+ * @return array<string, array{name: string, hosts: string[]}> Map of app_id UUID => app config.
+ */
+function wporg_get_allowed_apps(): array {
+	return array(
+		'c4c73a54-96d7-47b9-9bdc-1a66b9b04505' => array(
+			'name'  => 'WordPress.org MCP',
+			'hosts' => array(),
+		),
+	);
```
dd32: We should probably wrap this in a filter I guess.

dd32 (approved):
I'm not seeing any issues with this PR, code-wise, maybe a mu-plugin isn't ideal, we could just use a regular plugin that's activated on login.w.org, but I don't really have a big opinion on that :) 

I think long-term this should probably be in core, as the wp-admin authorize endpoint is IMHO less ideal than it could be.

I guess the question stands on whether most WordPress users would expect that they could be prompted by an arbitrary application :)

obenland:
Merged in https://meta.trac.wordpress.org/changeset/14682.

PR #555, opened 2026-02-23
Plugin Directory: Add wp-env local development environment

dd32:
## Summary

Local development environment for the Plugin Directory using `wp-env` (Docker, PHP 8.4).

**Environment setup:**
```bash
cd environments && npm install
npm run plugins:env start
```

On first start, ~30 plugins are auto-imported from wordpress.org (featured, popular, beta sections).

### What works
- Plugin browsing, single plugin pages, and search
- Plugin submissions and the review/approval process
- Readme validator, block plugin checker, release management pages
- Dev login button on wp-login for quick admin access
- `npm run plugins:import <slug>` to import individual plugins from SVN
- `npm run plugins:refresh` to re-import all plugin metadata from the API
- Admin notice reminding to import approved plugins via CLI

### What does not work
- Trigger SVN Sync button
- Custom crons tightly coupled to Cavalcade
- Commits metabox (no `trac_plugins` table)
- HelpScout metabox (no HelpScout access)
- Reviews (could be mocked in future)

### Environment (`environments/`)
- `plugin-directory/.wp-env.json` — wp-env config with PHP 8.4, mu-plugin mocks, theme mappings, lifecycle scripts
- `plugin-directory/bin/after-start.sh` — Installs CLI tools (svn, unzip, zip), sets up permalinks, creates pages, imports stub DB tables, creates browse terms, imports plugins
- `plugin-directory/bin/import-plugins.php` — Bulk imports plugin data from the wordpress.org REST API (metadata, users, icons, banners, screenshots, tags, contributors, committer access)
- `plugin-directory/bin/database-tables.sql` — Stub `wp_svn_access` and `wp_update_source` tables
- `mocks/mu-plugins-loader.php` — Loads mu-plugins from `pub/` and `wporg-mu-plugins/`
- `mocks/wporg-query-filter.php` — Intercepts queries to production-only tables
- `mocks/wporg-dev-login.php` — One-click admin login on wp-login screen
- `mocks/wporg-plugins-local-reminders.php` — Admin notice for approved plugins needing import
- `.nvmrc` — Node 20 requirement

### Plugin Directory code fixes
- Gate SVN credential usage on `defined()` checks, bail on writes without credentials
- Fix `SVN::ls()` fatal when XML parsing fails (no SVN server)
- Fix undefined `$plugins_prefix` variable in `custom_redirect()`
- Fix `/readme.txt` redirect path for non-`/plugins/` environments
- Gate HelpScout stats on `HELPSCOUT_PLUGINS_MAILBOXID` constant
- Gate `Two_Factor` and `is_email_address_unsafe()` on existence checks

### CI / Testing
- Flatten test directory structure (`tests/phpunit/tests/` to `tests/`)
- Update bootstrap for wp-env test framework compatibility
- Add Plugin Directory to GitHub Actions using environments wp-env with `cli` container
- Remote API tests excluded from CI suite

## Test plan
- [ ] `cd environments && npm install && npm run plugins:env start`
- [ ] Verify plugins are imported with icons, banners, and contributors
- [ ] Browse featured/popular/beta sections
- [ ] View a single plugin page without errors
- [ ] Submit a new plugin via /developers/add/
- [ ] Approve a plugin and verify the import notice appears
- [ ] Run `npm run plugins:import <slug>` for an approved plugin
- [ ] Run `npm run plugins:refresh` to re-import metadata
- [ ] Verify CI passes


obenland:
I had a bit of trouble getting it set up, but it might have just been that I had wp.org sandboxed and that messed some things up.

Things I noticed while taking it for a spin:

* It would be good for `./environments/node_modules` to be svn:ignored in the final version, if that's not already on your list.
* The FAQ accordion in single views seems to be broken.
* The advanced view is broken on first load. It needs a refresh to look normal again.
* Committers can't be added from advanced view, but that's probably to be expected without the db table.
* `./readme.txt` as the example readme doesn't seem to exist.
* The Readme Validator at http://localhost:8888/developers/readme-validator/ is empty.
* Uploading a plugin for review fails with "Error: The plugin has no name. Add a Plugin Name: line to your main plugin file and upload the plugin again.". Looks like `unzip` is missing. Maybe that could be installed in afterSetup or afterStart?
* Author Card meta box is pretty broken with PHP warnings.

I think it would be great if we could get plugin uploads to work. Pretty sweet beyond that, kudos!

obenland:
I ran into the same setup problems while trying to get `wp-env` working in [wporg-two-factor](https://github.com/WordPress/wporg-two-factor). Claude solved it by mapping mu-plugins in `.wp-env.json` to a mu-plugins folder, instead of individual files. Will have to check whether changing that here would solve that for me, too.

dd32:
> It would be good for ./environments/node_modules to be svn:ignored in the final version, if that's not already on your list.

Yeah I hadn't got that far yet :) 

> The FAQ accordion in single views seems to be broken.

Hmm... I hadn't even tested that yet, It looks like the custom import script (entirely claude built, and limited testing by me) doesn't import FAQ items properly.

(It's not handling screenshots either)

> Committers can't be added from advanced view, but that's probably to be expected without the db table.

The lack of the committers table is indeed the fault here, although I was originally intending not to have that functionality enabled, I ran into some other aspects that kinda needed it too, so it might be better to just stub it out.

> ./readme.txt as the example readme doesn't seem to exist.

A lot of stuff in the plugin directory still assumes that Plugins is always at `/plugins/`, right now `Plugin_Directory::custom_redirects()` isn't handling it correctly.  I knew the redirects didn't work, didn't notice the readme.txt :) 
bde611cff

> The Readme Validator at http://localhost:8888/developers/readme-validator/ is empty.

That's fixed in 12c6128b2 - I was going to have this pull the  pages & content directly from w.org, but the pages rest endpoint is broken :)  


> Author Card meta box is pretty broken with PHP warnings.
I had fixed that in trunk, b00d5b1ab5e7364d3236c57de68f46e9ad820fb8 wasn't merged into this branch yet though.

> The advanced view is broken on first load. It needs a refresh to look normal again.

Huh.. I thought.. OH... yeah that's because the `svn` executable doesn't exist... 

> Uploading a plugin for review fails with "Error: The plugin has no name. Add a Plugin Name: line to your main plugin file and upload the plugin again.". Looks like unzip is missing. Maybe that could be installed in afterSetup or afterStart?

This is caused by the single biggest reason I don't like `wp-env`; Most CLI tools are unavailable in the containers, which the plugin directory relies upon, ie. `unzip` and `svn`. This causes a lot of broken things when you start really trying to use it.

I've legitimately considered using other products like `ddev` instead, but they bring other problems.

I was planning on looking at wp-env to see if I could either a) Update the Docker containers it uses to have common CLI tools, or b) Add a configuration parameter to wp-env to allow executing additional steps into the DockerFiles.

I've also looked at using the playground runtime here instead, thinking it would use the host machines CLI tools, but I can't get `--config` and `--runtime` to play nicely together, it just crashes constantly.

> but it might have just been that I had wp.org sandboxed and that messed some things up.

The afterStart script does rely upon w.org being accessible, and if you have w.org sandboxed without a system-wide proxy, it'll fail to connect. 

> I ran into the same setup problems while trying to get wp-env working in [wporg-two-factor](https://github.com/WordPress/wporg-two-factor). Claude solved it by mapping mu-plugins in .wp-env.json to a mu-plugins folder, instead of individual files. Will have to check whether changing that here would solve that for me, too.

Claude has a lot of trouble mapping things into mu-plugins, as it highly depends upon the order they're added to the json.. I've equally had a lot of trouble with mapping mu-plugins, and this is the solution I've come up with to make it work reliably.. for me at least.

obenland:
Could you see if mapping `mu-plugins` first, then overlaying subdirectories works for you? It fixed all the errors I ran into.

It would map the whole `mu-plugins` directory to `./mocks/mu-plugins/` and have `wporg-query-filter.php`, `wporg-dev-login.php`, and `wporg-mu-plugins-loader.php` live in that folder. The directory overlay for `pub/` and `wporg-mu-plugins/` should still work on top.

```diff
	"mappings": {
+		"wp-content/mu-plugins": "./mocks/mu-plugins",
		"wp-content/mu-plugins/pub": "../wordpress.org/public_html/wp-content/mu-plugins/pub",
		"wp-content/mu-plugins/wporg-mu-plugins": "WordPress/wporg-mu-plugins#build",
-		"wp-content/mu-plugins/wporg-mu-plugins-loader.php": "./mocks/wporg-mu-plugins-loader.php",
-		"wp-content/mu-plugins/wporg-query-filter.php": "./mocks/wporg-query-filter.php",
-		"wp-content/mu-plugins/wporg-dev-login.php": "./mocks/wporg-dev-login.php",
		"wp-content/env-bin": "./plugin-directory/bin"
	},
```

dd32:
I've done something similar in 822bc20e1ce1f49f5cb40ab2433102e00afa09f5

I didn't map the mock directory as a whole, as I'd like to add other environments, and I'm not sure that all of these will apply to all environments yet.

We already had a mu-plugins/pub mapped, so just mapping the parent of that achieves the same thing effectively.

obenland:
Hm, interesting. Even with https://github.com/WordPress/wordpress.org/commit/822bc20e1ce1f49f5cb40ab2433102e00afa09f5 there's something about mapping individual files that seems to fail on my end.

```
Error response from daemon: 
failed to create task for container: 
failed to create shim task: 
OCI runtime create failed: runc create failed: 
unable to start container process: 
error during container init: 
error mounting "/host_mnt/Users/obenland/Documents/code/wordpress.org/environments/mocks/wporg-dev-login.php" to rootfs at "/var/www/html/wp-content/mu-plugins/wporg-dev-login.php": 
create mountpoint for /var/www/html/wp-content/mu-plugins/wporg-dev-login.php mount: 
mountpoint "/run/host_virtiofs/Users/obenland/Documents/code/wordpress.org/wordpress.org/public_html/wp-content/mu-plugins/wporg-dev-login.php" is outside of rootfs "/var/lib/docker/overlay2/98ecbe19536ab6686fb92e449405fc4a7c5f39a341d8ccf4519f2a70c392a72d/merged"
```

dd32:
@obenland I suggest... running a `destroy`, I've also updated the packages just incase that was related..

For some reason, even when using `--config` the mapping paths are relative to where you're running the command from, not relative to the wp-env config, which I think could be part of the problem, along with a invalid docker mount config from a previous iteration of the environment..

Claude was incorrect in the assumption behind 39d164b...

dd32:
At this point, I think it's as about complete as it's going to be for now.

Plugin submissions work, as does the review process.
In [c87b0b2](https://github.com/WordPress/wordpress.org/pull/555/commits/c87b0b2c5fb34fe44700a0d960589ca3b0425f2a) I've added a reminder that once the plugin is approved, it needs to be imported (I'm assuming that people submit  real plugins to test it...)

Some things don't work
 - Trigger SVN Sync button
 - Custom Crons that are tightly coupled to Cavalcade
 - Commits metabox (no trac_plugins table)
 - Helpscout metabox (no helpscout access)
 - No reviews (Could mock it)

It would be "helpful" if the Global WordPress Header/Footer detected running locally and semi-disabled itself, but that's outside of this issues scope.

✅  It would also be helpful if we had an API on w.org that dumps out `get_post()` + `array_intersect( $public_safe_fields, get_metadata() )`, which would allow a faster provisioning for plugins... It would avoid the current Metadata -> API output -> API Parsing -> Local metadata; Resolved in 7822382

dd32:
Merged in bc3041fa9f162145c37f12c49444b5cbeaa4dc25

PR #556, opened 2026-03-03
Fix: Tickets menu on Changeset pages redirects to make core/reports/ page instead of opening tickets menu

hbhalodia:
## Issue

Meta ticket -- https://meta.trac.wordpress.org/ticket/8183

- There was a JS error on the page accessing the value from the null.
- Added a safe check before accessing.

### DIFF
--- a/wordpress.org/public_html/style/trac/wp-trac.js
+++ b/wordpress.org/public_html/style/trac/wp-trac.js
@@ -1752,7 +1752,10 @@ var wpTrac, coreKeywordList, gardenerKeywordList, hideFromNewTickets, reservedTe
 				// This seems to be the easiest place to find the current Ticket ID..
 				var canonical = $( 'link[rel="canonical"]' ).prop( 'href' );
 				if ( canonical ) {
-					ticket = canonical.match( /\/ticket\/(\d+)$/ )[1];
+					let match = canonical.match( /\/ticket\/(\d+)$/ );
+					if ( match && match[1] ) {
+						ticket = match[1];
+					}
 				}
 
 				if ( ! ticket ) {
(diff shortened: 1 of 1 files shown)


wordpress.org/public_html/style/trac/wp-trac.js
```diff
@@ -1752,7 +1752,10 @@ var wpTrac, coreKeywordList, gardenerKeywordList, hideFromNewTickets, reservedTe
 				// This seems to be the easiest place to find the current Ticket ID..
 				var canonical = $( 'link[rel="canonical"]' ).prop( 'href' );
 				if ( canonical ) {
-					ticket = canonical.match( /\/ticket\/(\d+)$/ )[1];
+					let match = canonical.match( /\/ticket\/(\d+)$/ );
+					if ( match && match[1] ) {
+						ticket = match[1];
+					}
```
dd32: I was going to suggest using OR, but.. this is pretty ugly with no benefit:
```suggestion
					( canonical.match( /\/ticket\/(\d+)$/ ) || [,0] )[1]
```

Using Nullish operators can also one-line it, which feels much cleaner to me, as it can be easily used on more complicated code without having to nest multiple levels of if checks.
(This ends up as `ticket` being `undefined` which is fine for this use-case)
```suggestion
					canonical.match( /\/ticket\/(\d+)$/ )?.[0]
```
_(This is what I'm going to merge, you'll still get props for it)_

However, in this case, there's no need to check the match value, as if it matches due to the static regex you can be sure it'll be set.
```suggestion
					let match = canonical.match( /\/ticket\/(\d+)$/ );
					if ( match ) {
						ticket = match[1];
					}
```

hbhalodia: Thanks @dd32 For review.

Initially I had also used `canonical.match( /\/ticket\/(\d+)$/ )?.[0]` the nullish operator, but I searched the codebase if optional chaining is used or not for codebase to be consistent, which I ends up using the if checks.

I will update PR with what is being suggested.

Also, we are just matching ticket number instead of whole pattern, I guess it should be `canonical.match( /\/ticket\/(\d+)$/ )?.[1]`?

Thanks,

PR #564, opened 2026-03-17
Plugin Directory: UX changes to submitted plugins information.

frantorres:
Changes to the "Upload new version" link:
- It is now a button instead of a link.
- The wording is now `Upload updated "%s" plugin for review.` instead of `Upload new version...` given that there has been some confusion in the past regarding the need to update the version (which is not the case)

Other visual changes to improve the content hierarchy and appearance. Screenshot of the new aspect.

image


frantorres:
Added submission date so it can be compared with the new queue date #567 

image

frantorres:
Added an additional line showing information related to the email that they are gonna receive. 
Some plugin authors are not aware that they will receive an email, some of them receive it on spam, some of them don't know to which email that will be sent to.

Status new
image

Status pending
image

⚠️ Status pending includes data from a HelpScout email object that I wasn't able to test locally. It has guardrails to avoid issues, but it would need to be checked once it is online.

PR #570, opened 2026-03-23
Plugin Directory: New endpoints to integrate reviewers tools.

frantorres:
The internal reviewer tools are ready to integrate better with WordPress.org to make reviews faster and make further automations in the near future. 

Added 3 new endpoints:
- POST plugins/v1/plugin-review/{plugin_id}-{token}/assign     new → pending, also sets the reviewer
- POST plugins/v1/plugin-review/{plugin_id}-{token}/slug       renames a plugin awaiting review
- POST plugins/v1/plugin-review/{plugin_id}-{token}/approve    → approved

Authentication: As this will be executed from our tools and an application password has quite broad permissions I came up to use:
- the per-plugin private token that we already used to read information from the plugins privately, and
- to add a bit more of security, an Authorization: Bearer shared secret, PLUGIN_REVIEW_ENDPOINT_SECRET 
⚠️ We need the meta team to add this constant with a complex key and send it privately to us.

The reviewer is passed as user_id and the actions are verified by the authentication system and limited to the ones in the endpoints.

Other changes:
- Status_Transitions is only hooked in wp-admin, so I rewired that to work in this endpoints as well.
- Fixed a PHP warning in class-plugin.php

## Summary by CodeRabbit

- **New Features**
  - Added an endpoint for assigning an eligible reviewer while transitioning a plugin from “new” to “pending.”
  - Added secure request validation, reviewer eligibility checks, and clear error responses.
  - Improved status-transition handling across administrative and API workflows.
- **Bug Fixes**
  - Prevented plugin translation projects from being incorrectly marked inactive when a plugin is self-closed.
  - Improved handling when a plugin has no reviews to display.
- **Documentation**
  - Updated endpoint documentation to reflect required reviewer information, authentication, and response behavior.


wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-plugin-review.php
```diff
@@ -136,4 +144,43 @@ public function plugin_review_info( $request ) {
...
 	}
+
+	/**
+	 * Permission check for assigning a reviewer.
+	 *
+	 * @param \WP_REST_Request $request The Rest API Request.
+	 * @return bool
+	 */
+	public function assign_reviewer_permissions_check( $request ) {
+		return current_user_can( 'plugin_approve' );
```
dd32: This should be  checking`edit_post` and/or that checked in assign_reviewer.

Always check specific meta caps, and allow that to require the basic caps through map_meta_cap.

frantorres: Thanks for the feedback, applied!

wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-plugin-review.php
```diff
@@ -16,20 +22,82 @@ class Plugin_Review extends Base {
...
+				'required'    => true,
+			),
+		);
+
+		// An API Endpoint to change the status of a plugin from new to pending and assign a reviewer to it.
+		register_rest_route(
+			'plugins/v1',
+			$plugin_route . '/assign',
+			array(
+				'methods'             => WP_REST_Server::EDITABLE,
```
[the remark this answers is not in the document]
frantorres: Test made 👍

davidperezgar:
I've checked thanks to Codex and I found two remaining concurrency issues that should be addressed before merging:
1. The assign endpoint is not atomic. Two simultaneous requests can both see a plugin in new status, assign different reviewers, and return success. The later request overwrites the earlier reviewer. Please claim the plugin with a conditional new → pending update or a lock, returning a conflict when it has already been claimed.
2. A slug collision during rename can leave an unintended persisted change. If another process claims the slug after the availability check, WordPress may save a suffixed slug. The endpoint then returns an error, but the plugin has already been renamed. Please safely restore and verify the old slug, or make the update conditional.
The current CI checks are green otherwise.

frantorres:
Hi!
1. Regarding the assign endpoint being atomic, for this case maybe that is not needed because the team already has safeguards to avoid this type of collisions between reviewers (we cannot take plugins took by other reviewers in the control panel), even if we don't have it it is unlikely that it happens and even if that happens it is not critical as we are using the assignation mostly for stats and does not affect other processes.
2. Regarding the slug collision, again that's unlikely due to the small timeframe between checking and assigning it and even if that happens it is not critical as the slug can be still changed during the review process. Additional checks may be done at the end of the request, but I don't think it worth the while as we are not even going to do any further action if the slug if different to the requested one due to this.

wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-plugin-review.php
```diff
@@ -16,20 +22,82 @@ class Plugin_Review extends Base {
...
+
 		// An API Endpoint to expose more detailed plugin data for a pending plugin.
-		register_rest_route( 'plugins/v1', '/plugin-review/(?P<plugin_id>\d+)-(?P<token>[a-f0-9]{32})/?', array(
-			'methods'             => WP_REST_Server::READABLE,
-			'callback'            => array( $this, 'plugin_review_info' ),
-			'permission_callback' => array( $this, 'plugin_info_permission_check' ),
-		) );
+		register_rest_route(
+			'plugins/v1',
+			$plugin_route . '/?',
```
obenland, line 31:
```suggestion
			$plugin_route,
```

Do we do this in other endpoints? Can it be removed? Not sure that's needed.

wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-plugin-review.php
```diff
@@ -139,4 +224,266 @@ public function plugin_review_info( $request ) {
...
+		);
+
+		if ( is_wp_error( $result ) ) {
+			$result->add_data( [ 'status' => 500 ] );
+			return $result;
+		}
+
+		// `wp_insert_post()` alters the slug in a few cases, so check the plugin actually got it.
+		if ( get_post_field( 'post_name', $post->ID ) !== $new_slug ) {
+			return new WP_Error( 'slug_not_updated', 'Failed to update the plugin slug', [ 'status' => 500 ] );
```
obenland, line 496: > slug_not_updated

Not sure it needs changing, but it _did_ update the slug, just to something that's unexpected. And it doesn't restore the old one or surface the unexpected one.

wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-plugin-review.php
frantorres: This was the previous behaviour of this endpoint (this one was refactored, not changed). 
In the places I checked this URL is being shown with the final "/" and the tools we have are also processing it like that, so we should not remove the / (maybe we can do it otherwise, making it fixed, but I'm not sure if somewhere this is being checked without the /)

wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-plugin-review.php
frantorres: Thanks for pointing this out. I have changed the implementation to take this into account, and it now returns the new slug (rather than simply replying true/WP_Error) so we can handle that case downstream.

PR #582, opened 2026-03-31
Plugin Directory: Validate the required parameter for the bulk stats update endpoint

obenland:
## Summary
- The `/plugins/v1/update-stats` endpoint didn't define any argument requirements, so requests missing the `plugins` parameter would reach the callback and trigger a PHP warning when iterating over a null value.
- Adds a schema definition that validates the shape and types of the incoming data at the REST API layer, rejecting invalid requests before the callback runs.

## Test plan
- [ ] Confirm valid requests with a `plugins` object still work as expected
- [ ] Confirm requests missing `plugins` return a REST API validation error instead of triggering a PHP warning

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-internal-stats.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-internal-stats.php
@@ -23,6 +23,21 @@ class Internal_Stats extends Base {
 			'methods'             => \WP_REST_Server::CREATABLE,
 			'callback'            => array( $this, 'bulk_update_stats' ),
 			'permission_callback' => array( $this, 'permission_check_internal_api_bearer' ),
+			'args' => array(
+				'plugins' => array(
+					'required'             => true,
+					'type'                 => 'object',
+					'additionalProperties' => array(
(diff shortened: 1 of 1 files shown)


obenland:
@dd32 Could I ask for a sanity check for these changes? It doesn't look like the two places calling this endpoint should be bothered by getting an error returned when `plugins` is empty, but I wanted to double check.

This fixes:
```
E_WARNING: foreach() argument must be of type array|object, null given in wp-content/plugins/plugin-directory/api/routes/class-internal-stats.php:48
foreach() argument must be of type array|object, null given
```

https://github.com/WordPress/wordpress.org/blob/9209afdd3f0373f99f473f28d5931e2288c9c92a/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-internal-stats.php#L46-L48

dd32:
@obenland I'm thinking this validation isn't needed, because the sender should only be sending valid data; Something is broken if it's sending invalid data.. Pentesters aren't responsible here surely, as they won't pass the permission callbacks to reach that far.

I was going to update the error handler to understand rest-api requests and include a subset of the payload to diagnose this warning.

obenland:
Sounds good!

PR #583, opened 2026-03-31
PHPCS: Allowlist custom capabilities instead of disabling the sniff

obenland:
## Summary
- Replace the blanket `<severity>0</severity>` override on `WordPress.WP.Capabilities.Unknown` with an explicit allowlist of all custom capabilities used across the codebase, grouped by plugin.
- Fix four capability bugs surfaced by enabling the sniff:
  - **Plugin Directory:** ES Status tool used the role name `plugin_admin` instead of a capability — replaced with `plugin_approve`.
  - **Plugin Directory:** Comment row actions checked `manage_comments`, which is not a WordPress capability — replaced with `moderate_comments`.
  - **Photo Directory:** Moderation used the role name `photos_moderator` instead of a capability — replaced with `edit_photos`.
  - **BuddyPress:** Admin redirect checked four role names (`contributor`, `author`, `editor`, `administrator`) instead of capabilities — replaced with a single `edit_posts` check.

## Test plan
- [ ] Run `composer install && vendor/bin/phpcs --sniffs=WordPress.WP.Capabilities .` and confirm no warnings or errors
- [ ] Verify the Plugin Directory ES Index Status tool is still accessible to plugin admins
- [ ] Verify Plugin Directory internal note comment actions still work for reviewers
- [ ] Verify Photo Directory moderation capabilities still work for photo moderators
- [ ] Verify BuddyPress admin redirect still allows contributors and above into wp-admin

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -44,9 +44,85 @@
 
 		<!-- Short array syntax is already used throughout the codebase. -->
 		<exclude name="Universal.Arrays.DisallowShortArraySyntax" />
+	</rule>
 
-		<!-- Custom capabilities are valid in the WordPress.org environment; an allowlist is proposed in PR #583. -->
-		<exclude name="WordPress.WP.Capabilities.Unknown" />
+	<!-- Custom capabilities used across the WordPress.org environment. -->
+	<rule ref="WordPress.WP.Capabilities">
+		<properties>
+			<property name="custom_capabilities" type="array">
+				<!-- Plugin Directory -->
+				<element value="plugin_add_committer"/>
+				<element value="plugin_add_support_rep"/>
+				<element value="plugin_admin_edit"/>
+				<element value="plugin_admin_view"/>
+				<element value="plugin_approve"/>
+				<element value="plugin_close"/>
+				<element value="plugin_dashboard_access"/>
+				<element value="plugin_disable"/>
+				<element value="plugin_edit"/>
+				<element value="plugin_edit_others"/>
+				<element value="plugin_edit_pending"/>
+				<element value="plugin_manage_releases"/>
+				<element value="plugin_reject"/>
+				<element value="plugin_remove_committer"/>
+				<element value="plugin_remove_support_rep"/>
+				<element value="plugin_review"/>
+				<element value="plugin_self_close"/>
+				<element value="plugin_self_transfer"/>
+				<element value="plugin_set_category"/>
+				<element value="plugin_set_section"/>
+				<element value="plugin_toggle_public_preview"/>
+				<!-- Theme Directory -->
+				<element value="suspend_themes"/>
+				<element value="reinstate_themes"/>
+				<element value="theme_configure_categorization_options"/>
+				<!-- PHPUnit Test Reporter -->
+				<element value="edit_results"/>
+				<element value="publish_results"/>
+				<element value="edit_others_results"/>
+				<element value="read_private_results"/>
+				<!-- Translation Events -->
+				<element value="manage_translation_events"/>
+				<element value="create_translation_event"/>
+				<element value="view_translation_event"/>
+				<element value="edit_translation_event"/>
+				<element value="trash_translation_event"/>
+				<element value="delete_translation_event"/>
+				<element value="edit_translation_event_attendees"/>
+				<element value="edit_translation_event_title"/>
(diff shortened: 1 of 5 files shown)


obenland:
@dd32 Wanted to give you a chance to veto. It's a long allowlist, but it enables catching misuses like the other changes in this PR.

wordpress.org/public_html/wp-content/plugins/photo-directory/inc/moderation.php
```diff
@@ -291,7 +291,7 @@ public static function disable_own_post_editing( $caps, $cap, $args, $user ) {
 		}
 
 		// Bail if user isn't a moderator.
-		if ( ! user_can( $user->ID, 'photos_moderator' ) ) {
+		if ( empty( $caps['edit_photos'] ) ) {
```
dd32: I think this needs a `user_can()`, because super-admin roles don't have the caps.

wordpress.org/public_html/wp-content/plugins/plugin-directory/admin/tools/class-elasticsearch-status.php
```diff
@@ -27,14 +27,14 @@ public function add_to_menu() {
 			'plugin-tools',
 			'ES Index Status',
 			'ES Index Status',
-			'plugin_admin',
+			'plugin_approve',
```
dd32, line 30: I meant to target the `plugin admin` role, and not `plugin reviewer`, but you're right, `plugin_approve` is appropriate here to target that role.

phpcs.xml.dist
```diff
@@ -119,9 +119,89 @@
 		<severity>0</severity>
 	</rule>
 
-	<!-- Custom capabilities are valid in the WordPress.org environment. -->
-	<rule ref="WordPress.WP.Capabilities.Unknown">
-		<severity>0</severity>
+	<!-- Custom capabilities used across the WordPress.org environment. -->
+	<rule ref="WordPress.WP.Capabilities">
```
dd32, line 51: I'm not really a fan of this style of nitty-gritty capability listing.. but I'm not veto'ing this.

Perhaps we can do a middle-ground of warning severity or notice severity, and not error?

Let's also document it here that PRs can update the list as needed, and this isn't a strict requirement, rather just to catch those typos.


```suggestion
	<!-- Custom capabilities used across the WordPress.org environment. -->
	<!-- Please note: Using custom capabilities is allowed, just document them here please! -->
	<rule ref="WordPress.WP.Capabilities">
```

dd32 (approved):

wordpress.org/public_html/wp-content/plugins/photo-directory/inc/moderation.php
dd32: Hmmm.. actually... this is a `user_has_cap` filter.. Using user_can and friends in that is discouraged in cap filters.. due to loops..

I think the proper check here would be to be either requiring a cap by appending to $caps, or appending `do_not_allow` to block.. or.. something.. else...

obenland: I suppose it depends on how we want this to behave. It sets `edit_photos` to `false` when the photo in question is the moderator's own photo.

If we want super admins to be able to edit their own submissions we'd append a `|| is_super_admin( $user->ID )` check to bail. Otherwise a `&& ! is_super_admin( $user->ID )` to apply to them, too?

obenland: Applied in bc8f9c5 — switched to `user_can( $user->ID, 'edit_photos' )`.

On the loop concern: the re-entry terminates here. The nested check maps to `[ 'edit_photos' ]` with no object ID, so when the filter re-enters it bails at the `isset( $args[2] )` guard before doing any work. The pre-PR code already called `user_can()` inside this filter — this restores that pattern, just with a registered capability instead of the role name the sniff flags.

On the super-admin question: they never reach the restriction either way — `WP_User::has_cap()` short-circuits for super admins before `user_has_cap` filters run, so they can always edit their own submissions without needing an `is_super_admin()` bail. Caped users (who aren't network super admins) now pass the moderator check via the priority-11 grant, so the own-submission restriction applies to them, which seems right.

obenland:
@copilot resolve the merge conflicts in this pull request

PR #591, opened 2026-04-03
Plugin Directory - translation sync: save overwritten strings as old

timse201:
related to trac ticket: [6320](https://meta.trac.wordpress.org/ticket/6320)

fixes: [8217](https://meta.trac.wordpress.org/ticket/8217)

If we sync strings from dev to stable, the overwritten (fuzzy) string in stable will fire gp_translation_prepare_for_save again. Since there is no check for old strings, this one becomes current again because it adopts the status from the map. As a result, there are 2 current strings in stable.


obenland:
@ocean90 Heads up, in case you want to follow-up on this.

PR #592, opened 2026-04-03
Set_Plugin_Project - validate JSON via json_validate()

timse201:
fixes: https://meta.trac.wordpress.org/ticket/8218

Replaces the manual string check with PHP 8.3 native json_validate() function.


dd32:
At present we're running PHP 8.1 on WordPress.org (PHP 8.3 to come shortly).

This check is primarily supposed to catch HTML output from the endpoint, in the case of errors.

PR #606, opened 2026-04-14
Decodes plugin titles in email subjects and bodies

nimesh-xecurify:
Introduces a helper method to ensure plugin titles are HTML entity decoded when used in email plain-text contexts. This prevents titles containing HTML entities from displaying incorrectly.

Updates all relevant email classes to utilize this new method for consistent output.

Trac Ticket : https://meta.trac.wordpress.org/ticket/8231

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-base.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-base.php
@@ -163,6 +163,13 @@ https://make.wordpress.org/plugins', 'wporg-plugins' );
 	}
 
 	/**
+	 * Return the plugin's title with HTML entities decoded for plain-text use (e.g., email subjects).
+	 */
+	protected function plugin_title() {
+		return html_entity_decode( $this->plugin->post_title, ENT_QUOTES | ENT_HTML5, 'UTF-8' );
+	}
+
+	/**
 	 * A simple way to convert a WP_User object to a displayable username.
 	 * This shouldn't be needed, but unfortunately often is on WordPress.org.
 	 */
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-committer-added.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-committer-added.php
@@ -22,7 +22,7 @@ class Committer_Added extends Base {
 		return sprintf(
 			/* translators: 1: Plugin Name */
 			__( 'New committer added to %s', 'wporg-plugins' ),
-			$this->plugin->post_title
+			$this->plugin_title()
 		);
 	}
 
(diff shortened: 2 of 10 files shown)


dd32 (commented):
AI-assisted review (prompted by @dd32):

The change looks good — clean fix for HTML entities showing literally in email subjects/bodies. One suggestion:

Consider using `wp_specialchars_decode( $this->plugin->post_title, ENT_QUOTES )` instead of `html_entity_decode( ..., ENT_QUOTES | ENT_HTML5, 'UTF-8' )`. It's the WordPress wrapper that only decodes the entities WordPress itself encodes (`&amp;`, `&lt;`, `&gt;`, `&quot;`, `&#039;`), and is what `get_the_title()` uses internally.

We considered `get_the_title()` directly, but it runs the `the_title` filter which could add extra markup or decorations from themes/plugins — not safe for plain-text email context. Calling `wp_specialchars_decode()` directly on `post_title` is more predictable here.

wordpress.org/public_html/wp-content/plugins/plugin-directory/email/class-base.php
```diff
@@ -162,6 +162,13 @@ public function get_email_signature() {
 https://make.wordpress.org/plugins', 'wporg-plugins' );
 	}
 
+	/**
+	 * Return the plugin's title with HTML entities decoded for plain-text use (e.g., email subjects).
+	 */
+	protected function plugin_title() {
+		return html_entity_decode( $this->plugin->post_title, ENT_QUOTES | ENT_HTML5, 'UTF-8' );
```
dd32, line 169: AI-assisted review (prompted by @dd32):

~~Consider using `wp_specialchars_decode()` instead — it's the WordPress wrapper that only decodes the entities WordPress itself encodes.~~

**Edit 2:** Retracted. After checking the database, there are ~60 plugin titles with entities beyond the 5 that `esc_html()` produces (`&reg;`, `&trade;`, `&deg;`, `&ccedil;`, `&oacute;`, `&nbsp;`, `&#9733;`, `&#x1F3E6;`, etc.). These are stored directly in the DB from readme/header parsing, not added on read. `wp_specialchars_decode()` would leave these undecoded in emails. The original `html_entity_decode(ENT_QUOTES | ENT_HTML5, 'UTF-8')` is the correct approach.

~~**Edit:** Correction — `get_the_title()` does not use `wp_specialchars_decode()`. It is `the_title_attribute()` that does.~~

We considered `get_the_title()` directly, but it runs the `the_title` filter which could add extra markup from themes/plugins — not safe for plain-text email context.

~~```suggestion
		return wp_specialchars_decode( $this->plugin->post_title, ENT_QUOTES );
```~~

dd32:
AI totally got this wrong :) We want to do a full decode here for Email, as plugin authors can put limited HTML into their plugin names in WordPress; Specifically entities, such as `&copy;` or `&reg;`

Issue #603, opened 2026-04-13
Feature request: add "Job" section to WordPress.org profiles


annezazu:
[Pulling this comment from Matt](https://wordpress.slack.com/archives/C18723MQ8/p1776055702621089?thread_ts=1776036987.890829&cid=C18723MQ8) into this issue:

>even better, what would it look like if we actually made our profiles as useful and iconic as Githubs, and how can we embed jobs and job-seeking into the product of the main website rather than a side thing.

To head in that direction, I think we need a few things:

- New editable jobs section of our wordpress.org profiles that includes the ability to show open to work, list latest jobs, and list key achievements in WordPress. The last point ties in nicely with talking about output not input, like contribution hours. 
- Corresponding jobs information listed on public profile, including the ability to show "open to work" in Gravatar.
- The ability to browse folks open to work with a link from https://jobs.wordpress.net/ (similar to browse jobs, could browse candidates with some filtering). 

For the first two, here's a quick vibe coded design of what I imagine we could do, using what we already have and adding to it:

https://github.com/user-attachments/assets/7f252bd1-6ae3-4ad2-bd6f-7f5c6f41c65c

Here are links to the prototypes:

- Edit view: https://annezazu.github.io/jobs-profile/                                                                                          
- Public profile view: https://annezazu.github.io/jobs-profile/profile.html                                                                                                                                                              
                                                         
Feedback welcomed but let's see what we can move on. cc @m in case you have more to add.

jeffpaul:
In lieu of the "Add a position" bit, could we allow uploading a resume/CV or linking to LinkedIn (or perhaps a resume/cv page from someone's personal site) instead as a simple, first approach?

annezazu:
If we can add a similar "connect to linkedin" that we have for GitHub, that would be a great solution to auto-fill. I defer to @dd32 what is most feasible to do in the short term.

annezazu:
To track my thinking out loud, I can do a Make Core post after an initial version lands to get more feedback around what would be helpful for folks and what they like most about places like LinkedIn. I imagine much of it is network effects but we can try to replicate aspects.

dd32:
> If we can add a similar "connect to linkedin" 

We could add a OIDC connect flow for LinkedIn, but, it does not return the users vanity url / link. It also doesn't return job history or any meaningful data other than their avatar, fullname, and email address. The only thing it would do is "Yes, this user has a linkedin profile". Useless.

[...]

PR #611, opened 2026-04-21
Add accordion panel styles to jobs FAQ page

annezazu:
## Summary

- Mirrors existing `dd` CSS rules for `.wp-block-accordion-panel` in the jobs FAQ page stylesheet
- Enables the FAQ at jobs.wordpress.net/faq/ to be converted from a definition list (`dl/dt/dd`) to accordion blocks while maintaining correct styling for lists and links

## Test plan

- [ ] After merging, update the FAQ page content at jobs.wordpress.net/wp-admin to use accordion blocks
- [ ] Verify list items render with proper markers and spacing
- [ ] Verify nested lists render correctly (disc → circle)
- [ ] Verify links display with correct color and underline styling
- [ ] Verify link hover state works correctly

Refs: https://github.com/WordPress/wordpress.org/issues/603

### DIFF
--- a/jobs.wordpress.net/public_html/wp-content/themes/jobswp/style.css
+++ b/jobs.wordpress.net/public_html/wp-content/themes/jobswp/style.css
@@ -1314,6 +1314,33 @@ input[type="submit"].submit-job {
 	color: var(--color-deep-blueberry);
 }
 
+.page-faq .entry-content .wp-block-accordion-panel ul {
+	margin: var(--space-10) 0 0;
+	padding-left: var(--space-20);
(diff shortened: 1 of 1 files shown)


annezazu:
For some human context, I am doing this in order to use the new accordion block from WordPress 6.9 on this page rather than the custom HTML blocks. The theme is super old though so the styling needs to be updated. This makes me want to overhaul the entire theme to be a block theme next but one step at a time.

dd32:
@annezazu The B-Bot is very hungry, submit PRs from a fork :)

PR #622, opened 2026-05-07
Plugin Directory: server-render screenshots gallery + Gallery Lightbox Enhancements polyfill

dan-zakirov:
Replaces the bespoke `wporg-plugins-2024` theme JS screenshots gallery with a server-rendered `[wporg-plugins-screenshots]` shortcode in the Plugin Directory plugin, and adds a small polyfill plugin (Gallery Lightbox Enhancements) that restores two long-standing Gutenberg gaps locally so the wp.org migration can ship without waiting on core.

### Layout cases — counts 1 to 9

| Case | Screenshot |
|---|---|
| **N=1** — single tile, max-width 720 px | one-no-caption |
| **N=2** — 2-col row-aligned grid | two-with-captions |
| **N=4** — 3-up grid + last figure as full-width anchor | four-was-two-col-bug |
| **N=5** — 3+2 row-aligned grid | five-screenshots |
| **N=6** — 3×2 grid | six-no-captions |
| **N=6** — 3×2 grid | six-with-captions |
| **N=7, mixed aspect** — brick masonry, natural ratios preserved | seven-mixed-aspect |
| **N=8** — brick masonry | eight-screenshots |
| **N=9 (reveal threshold edge)** — full grid, no Show-all button | nine-edge-no-button |

### Reveal flow (N≥10)

| State | Screenshot |
|---|---|
| **Show-all button + fade overlay** — collapsed wrap clipped at 32rem | button-hover |
| **After Show-all click** — wrap unfolds to full content, fade and button removed | after-reveal |

### Mobile (≤599 px)

| Case | Screenshot |
|---|---|
| **Collapse suppressed** — every figure rendered, no fade or Show-all button on small viewports | mobile-collapse-suppressed |
| **N=33 mobile** — full gallery, vertical scroll only | mobile-thirty-three |

## What changes

- **`wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php`** — new shortcode that renders every screenshot through `core/gallery` + `core/image` blocks, lightbox enabled per tile. Galleries up to 9 figures show the full grid; larger ones get a `Show all N screenshots` reveal button. Layout is deterministic: N=1 single tile, N=2-4 row-aligned grid, N≥5 brick masonry by default with a row-aligned grid upgrade when every screenshot shares an aspect ratio.
- **`wp-content/plugins/gallery-lightbox-enhancements/`** — polyfill plugin that adds a `is-style-masonry` block style for `core/gallery` and renders `figcaption` text inside the lightbox overlay. Both behaviours are gated behind `Version_Guard::should_load()`, which lets the plugin self-disable once core ships the features.
- **`wp-content/themes/pub/wporg-plugins-2024/`** — removes the legacy theme implementation: `client/screenshots/`, the dedicated `theme.js` entry, the matching SCSS partials, and the `build:old:js` npm script.

## Why

Server-side rendering makes the gallery cacheable by Varnish, works without JS, and gives third-party scrapers and SEO crawlers the full markup. Every `` ships with `loading="eager"`, intrinsic `width` / `height` attributes, and `fetchpriority="high"` on the cap-window thumbnails, so CLS stays at 0 while screenshots decode. Aspect ratios and pixel dimensions come from a single Memcached-cached probe (md5-keyed by URL list, so a screenshot revision auto-invalidates).

## Closes / refs

Meta Trac (the long-standing screenshot UX gaps this PR addresses):

- Closes [meta#2273](https://meta.trac.wordpress.org/ticket/2273) — Screenshot UI redesign (open ~9 years, primary tracking ticket)
- Closes [meta#8083](https://meta.trac.wordpress.org/ticket/8083) — Add lightbox for plugin screenshots
- Closes [meta#8204](https://meta.trac.wordpress.org/ticket/8204) — Chrome 146 table-cell layout bug (the new shortcode renders through CSS Grid / columns, the legacy table-cell path is dropped)
- Refs [meta#5190](https://meta.trac.wordpress.org/ticket/5190) — Lazy loading screenshots (we go the opposite way: eager + `fetchpriority`, see "Why" above)
- Refs [meta#7878](https://meta.trac.wordpress.org/ticket/7878) — Screenshot navigation (handled by the core lightbox arrows)

Upstream Gutenberg gaps that the polyfill plugin addresses on the user side:

- [gutenberg#28247](https://github.com/WordPress/gutenberg/issues/28247) — Gallery masonry layout (open since January 2021)
- [gutenberg#60469](https://github.com/WordPress/gutenberg/issues/60469) — Lightbox captions excluded from overlay (regression since WP 6.5)

## Prior art on this task

Earlier attempts at the same migration that fed into this PR's design:

- [WordPress/wordpress.org#524](https://github.com/WordPress/wordpress.org/pull/524) — @ryelle's Gallery block migration draft.
- [WordPress/wordpress.org#67](https://github.com/WordPress/wordpress.org/pull/67) — older screenshot redesign attempt (predates Gutenberg gallery, kept for context).
- [WordPress/wporg-parent-2021#178](https://github.com/WordPress/wporg-parent-2021/pull/178) — lightbox component on the parent theme.

## Safety / scope

The change is additive on the Plugin Directory side and isolated for the polyfill, with one purposeful subtraction in the theme. Map of every touched path:

| Path | Risk |
|---|---|
| `wp-content/plugins/gallery-lightbox-enhancements/` (new plugin, 9 files) | None on its own. Self-disables via `Version_Guard::should_load()` once core ships the features. Activation lives outside this PR (mu-plugin loader on production). |
| `wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php` | Rewritten body, same registration hook. Shortcode tag, file name and class FQN are unchanged, so anything that already calls `do_shortcode( '[wporg-plugins-screenshots]' )` keeps working. |
| `wp-content/plugins/plugin-directory/shortcodes/assets/screenshots.css`, `screenshots.js` | New files. Enqueued only inside the shortcode's render path, so other pages on wp.org/plugins do not load them. |
| `wp-content/plugins/plugin-directory/class-plugin-directory.php` (+1) | One `add_filter( 'wp_resource_hints', [ Screenshots::class, 'add_resource_hints' ] )` that contributes `preconnect` / `dns-prefetch` URLs for the Photon host on single-plugin pages. Hint only, no blocking behaviour. |
| `wp-content/themes/pub/wporg-plugins-2024/client/screenshots/` (deleted) | Main subtraction. A repo-wide grep finds no external imports of these modules; happy to add a shim if a downstream consumer is missed. |
| `wp-content/themes/pub/wporg-plugins-2024/css/style.css`, `style-rtl.css` | Regenerated build artefacts after dropping the screenshot SCSS partials and the `build:old:js` script. No theme JS or CSS for screenshots remains — the shortcode owns the whole flow. |

The migration does not touch DB schema, post types, taxonomies, REST endpoints, or any sibling shortcode in `plugin-directory/shortcodes/`.

## Notes for reviewers

- All PHP passes the project `phpcs.xml.dist` ruleset (WordPress-Core / Docs / Extra, PHP 8.4 compat).
- No theme JS or CSS left for screenshots — the shortcode owns the whole flow.
- Polyfill is fully self-contained: drop in, drop out once core ships the features.
- cc @dd32 — would value your eyes on the `Version_Guard` placeholder strategy and on the deletion of the old `client/screenshots/` modules.
- cc @ryelle — this picks up the direction from #524; happy to align further on the layout decisions if you have feedback.


dan-zakirov:
A bit of context on how we got here: we first proposed an upstream Gutenberg fix for lightbox captions and, separately, a masonry block style for `core/gallery`. The masonry proposal was declined upstream and the captions one has not had movement; rather than block the wp.org migration on either, we ship the same behaviour as a self-disabling polyfill plugin so the gallery on `wordpress.org/plugins/{slug}/` can move forward today. If either fix lands in core later, flip `Version_Guard::CORE_VERSION_WITH_FEATURES` to that release tag and the plugin steps aside.

On captions specifically: every figure ships its caption text, but it surfaces only inside the lightbox overlay and never on the gallery itself. Plugin authors routinely write multi-sentence captions, and rendering those under each thumbnail would tear the brick layout apart with uneven figure heights. The lightbox is the right surface for that copy - it has room for it without compromising the grid.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -1,41 +1,754 @@
...
+	/**
+	 * Reads the width and height from an image's header bytes, without
+	 * touching GD or the PHP warning surface that `getimagesizefromstring`
+	 * emits on partial / malformed input. Supports PNG, JPEG, and GIF —
+	 * the formats Plugin Directory accepts as screenshots.
+	 *
+	 * @param string $bytes Raw image bytes (the first 32 KB is enough).
+	 * @return array|false `[ $width, $height ]` on success, false otherwise.
+	 */
+	private static function parse_image_dimensions( $bytes ) {
```
dd32: I think I'd still prefer to use `getimagesize()` here, but instead of having it within the Screenshots handler, fetch the resolutions during the plugin import process.

If the file has malformed image headers, we can also skip importing those images... Although I'm betting this would exclude soem animated images _(..what a shame...)_

We can backfill the existing image metadata, such that we can have this as assuming all data is stored.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -1,41 +1,754 @@
...
+	 *
+	 * @param string $rendered_gallery Output of do_blocks() for the gallery.
+	 * @param int    $count            Total number of screenshots.
+	 * @return string
+	 */
+	protected static function wrap_with_show_all_button( $rendered_gallery, $count ) {
+		$label = sprintf(
+			/* translators: %d: total number of screenshots in the gallery. */
+			_n( 'Show all %d screenshot', 'Show all %d screenshots', $count, 'wporg-plugins' ),
+			$count
```
dd32:
```suggestion
			/* translators: %s: total number of screenshots in the gallery. */
			_n( 'Show all %s screenshot', 'Show all %s screenshots', $count, 'wporg-plugins' ),
			number_format_i18n( $count )
```
As a general thing, best to always format with %s and number_format_i18n for the 1000+ use-case, and as such, keep this standardised with those formatters in this <100 use-case.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -1,41 +1,754 @@
...
+			file_exists( $css_path ) ? (string) filemtime( $css_path ) : '1.0'
+		);
+
+		wp_enqueue_script(
+			'wporg-plugins-screenshots',
+			plugins_url( $js_rel, $plugin_main ),
+			array(),
+			file_exists( $js_path ) ? (string) filemtime( $js_path ) : '1.0',
+			true
+		);
```
dd32: Let's just hard-code these to use `filemtime()`, and duplicate the path in the plugins_url and filemtime calls.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -1,41 +1,754 @@
...
+	 * browser can warm up the TLS handshake while the page HTML is
+	 * still streaming. Saves ~50–150 ms on the first thumbnail paint
+	 * for cold visitors. Hooked from `class-plugin-directory.php`.
+	 */
+	public static function emit_preconnect() {
+		if ( ! is_singular( 'plugin' ) ) {
+			return;
+		}
+		$env = function_exists( 'wp_get_environment_type' ) ? wp_get_environment_type() : 'production';
+		if ( 'production' !== $env ) {
```
dd32:
```suggestion
		if ( 'production' !== wp_get_environment_type() ) {
```

Let's just target modern WP

dd32: Actually; Let's just always include these via the `wp_resource_hints` on singular views.  Production or local doesn't really matter.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dd32: I'd also move these to plugins/plugin-directory/{css,js}/** instead of assets within the shortcode folder.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/gallery-lightbox-enhancements.php
```diff
@@ -0,0 +1,30 @@
...
+ * @package GalleryLightboxEnhancements
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+require_once __DIR__ . '/includes/class-version-guard.php';
+
+if ( ! GalleryLightboxEnhancements\Version_Guard::should_load() ) {
+	return;
+}
```
dd32, line 24: The main problem we have here is that we run WordPress Trunk and the latest Gutenberg release on WordPress.org.

Realistically, as a result, WordPress.org will need to update this plugin to note the new version guard asap, which is probably fine, as there should be a few weeks hopefully between PR merge and gutenberg inclusion.

dd32 (commented):
@dan-zakirov Thanks for the PR! I've added some commentary below, I've only code reviewed, not actually tried running it, but it looks like a strong start!

I'm going to fire up a PR to [extract the resolution of assets at import time](https://github.com/WordPress/wordpress.org/pull/622#discussion_r3216533310), and backfill them, to make the code cleaner and allow such improvements to also be used for the WordPress.org plugins API.

I haven't checked, but I think I recall the output of this is used in the WordPress.org plugins api somewhere, somehow..

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dd32: Initial swipe at it #628  (untested..)

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: Done in b8b63bccb - switched to %s + number_format_i18n( $count ), kept the same _n() pluralisation. Translator comment updated to %s as well.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: Done in d6a9458bb - switched to a wp_resource_hints filter on singular plugin views. Dropped the function_exists guard and the production-only check as you suggested - hint is harmless in any environment and we target modern WP

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: Done in 64a58067b - moved screenshots.{css,js} to plugin-directory/css/ and plugin-directory/js/ (alongside edit-form.{css,js}), inlined filemtime() directly into plugins_url() / version arg without the file_exists() fallback or helper variables.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: Waiting on #628 to land - once it does I'll drop the parser here and read width/height straight from the assets_screenshots post meta.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -1,41 +1,764 @@
...
+	protected static function enqueue_assets() {
+		$css_rel = 'shortcodes/assets/screenshots.css';
+		$js_rel  = 'shortcodes/assets/screenshots.js';
+
+		$plugin_dir  = dirname( __DIR__ );
+		$plugin_main = $plugin_dir . '/plugin-directory.php';
+
+		$css_path = $plugin_dir . '/' . $css_rel;
+		$js_path  = $plugin_dir . '/' . $js_rel;
+
```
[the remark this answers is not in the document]
dan-zakirov: Outdated - paths were updated in 64a58067b. enqueue_assets() now points to plugin-directory/css/screenshots.css and plugin-directory/js/screenshots.js, which match the file locations in this branch's HEAD.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/includes/class-lightbox-captions.php
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Stores the figcaption text alongside the lightbox state for the given image.
+	 *
+	 * @param string $content Rendered block markup.
+	 * @param array  $block   Parsed block.
+	 * @return string Unchanged content; we only mutate Interactivity API state.
+	 */
+	public static function add_caption_to_state( $content, $block ) {
+		if ( ! str_contains( $content, 'wp-lightbox-container' ) ) {
```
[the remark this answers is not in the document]
dan-zakirov: Done in f2207fbce - replaced str_contains() with `false === strpos( ..., 'wp-lightbox-container' )` to match the plugin's declared Requires PHP: 7.4 minimum.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/includes/class-version-guard.php
```diff
@@ -0,0 +1,52 @@
...
+	public static function should_load() {
+		// Vanilla WordPress core: plugin is the only source of these features.
+		if ( ! defined( 'GUTENBERG_VERSION' ) ) {
+			return true;
+		}
+
+		// Sentinel: no upstream release tracked yet, polyfill stays on.
+		if ( '' === self::CORE_VERSION_WITH_FEATURES ) {
+			return true;
+		}
```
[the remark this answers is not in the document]
dan-zakirov: Done in b672d29ae - added a WP_VERSION_WITH_FEATURES constant alongside CORE_VERSION_WITH_FEATURES so the guard self-disables on vanilla WordPress installs once the features ship in core, not only on installs running the Gutenberg plugin.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/includes/class-version-guard.php
```diff
@@ -0,0 +1,52 @@
...
+
+		return version_compare(
+			GUTENBERG_VERSION,
+			self::CORE_VERSION_WITH_FEATURES,
+			'<'
+		);
```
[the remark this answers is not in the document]
dan-zakirov: Done in b672d29ae - added a WP_VERSION_WITH_FEATURES constant alongside CORE_VERSION_WITH_FEATURES so the guard self-disables on vanilla WordPress installs once the features ship in core, not only on installs running the Gutenberg plugin.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/gallery-lightbox-enhancements.php
dan-zakirov: Noted - and after b672d29ae the guard exposes two constants now: CORE_VERSION_WITH_FEATURES for the Gutenberg-plugin track and WP_VERSION_WITH_FEATURES for the core track. Setting either to a real release tag steps the polyfill aside on installs running that version or newer; for wordpress.org you'd only need to bump CORE_VERSION_WITH_FEATURES once the next Gutenberg release ships these features. Happy to file a meta-repo note when that lands so the constant bump doesn't get forgotten.

wordpress.org/public_html/wp-content/plugins/plugin-directory/js/screenshots.js
```diff
@@ -0,0 +1,298 @@
...
+ * Every figure is in the DOM from first paint — masonry has already
+ * balanced columns and never needs to re-balance. Collapse / expand
+ * happens entirely on the wrap via `max-height`:
+ *
+ *   - On init, measure the natural bottom of the ninth figure across
+ *     the columns and write it into `--collapse-height` so the wrap
+ *     clips at exactly the right line.
+ *   - On click, measure the wrap's full `scrollHeight` and write it
+ *     into `--full-height`, then flip the `is-revealed` class so the
+ *     CSS transition animates from collapse-height to full-height.
```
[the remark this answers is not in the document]
dan-zakirov: Done in 814a32dbb - removed the unused COMPACT_VISIBLE constant, documented the $dimensions param on build_gallery_markup(), and refreshed the CSS/JS/class-header comments that still described the dropped dynamic --collapse-height measurement.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/includes/class-lightbox-captions.php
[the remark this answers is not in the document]
dd32: Pointless change though, since WordPress contains a `str_contains` polyfill :)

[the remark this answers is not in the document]
dan-zakirov: TIL, thanks for the heads-up - will keep WP's compat.php polyfills in mind for future PRs.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dd32: #628 has landed, and all plugins asset metadata has been updated, Haven't confirmed if the local environment gets it yet...

dan-zakirov: Done in 6af7759 - `get_dimensions()` now plucks width/height straight from each $screenshot entry; fetch_dimensions(), parse_image_dimensions() and the verdict cache are gone. Verified the data is live via wp-json/wp/v2/plugin/163.

bor0:
Hi! Thanks for your work here. Since @dd32 already reviewed the code earlier, I didn't look at the code (yet), but I did a first round of testing. Things look pretty good!

Only found one small thing: when opening the page with javascript disabled, you can click on an image but nothing happens.

| What | Screenshot |
| ---- | ---------- |
| Plugin with >10 screenshots (mobile) | Screenshot 2026-05-28 at 23 49 47 |
| Plugin with >10 screenshots (mobile) | Screenshot 2026-05-28 at 23 49 20 |
| screenshots.js loaded properly | Screenshot 2026-05-28 at 23 48 54 |
| Old theme.js no longer loaded | Screenshot 2026-05-28 at 23 48 44 |
| Performance | Screenshot 2026-05-28 at 23 48 26 |
| Plugin with 6 screenshots | Screenshot 2026-05-28 at 23 47 35 |
| Open a screenshot | Screenshot 2026-05-28 at 23 46 50 |
| Plugin with 6 screenshots (mobile) | Screenshot 2026-05-28 at 23 46 32 |
| Open a screenshot (mobile) | Screenshot 2026-05-28 at 23 46 42 |

Plan to do another round of extra testing.

dan-zakirov:
> Hi! Thanks for your work here. Since @dd32 already reviewed the code earlier, I didn't look at the code (yet), but I did a first round of testing. Things look pretty good!
> 
> Only found one small thing: when opening the page with javascript disabled, you can click on an image but nothing happens.

@bor0

Thanks for the round and screenshots.

The no-JS click is inherited from core/image's lightbox attribute, same applies to any lightbox-enabled core image, so it's not a PR regression.

If you want, I can wrap the img in `<a href="{src}" target="_blank" rel="noopener">` so no-JS users get the full-res, and the core lightbox JS does its preventDefault as usual. Or keep it consistent with core. Your call.

This PR is a first cut focused on retiring the old gallery quickly. Edge cases by screenshot count and height will land in follow-ups.

bor0:
> no-JS users get the full-res, and the core lightbox JS does its preventDefault as usual

This sounds like a good UX improvement to me! Especially that it's low effort, I'd propose to go with it :)

dan-zakirov:
> > no-JS users get the full-res, and the core lightbox JS does its preventDefault as usual
> 
> This sounds like a good UX improvement to me! Especially that it's low effort, I'd propose to go with it :)

Done in 7f291542. Wrapped each img in `<a href="{src}" target="_blank" rel="noopener">`. The core lightbox JS attaches its click handler on hydration so the JS path is unchanged.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/includes/class-lightbox-captions.php
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		$caption = trim( wp_strip_all_tags( $matches[1] ) );
+		$id      = isset( $block['attrs']['id'] ) ? absint( $block['attrs']['id'] ) : 0;
+
+		if ( '' === $caption || ! $id ) {
+			return $content;
+		}
+
+		wp_interactivity_state(
```
thilinah, line 56: The light box is readign the caption from image alt attribute. So do we need this?

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -1,41 +1,754 @@
...
+		$attrs = wp_json_encode(
+			array(
+				'id'              => $id,
+				'sizeSlug'        => 'large',
+				'linkDestination' => 'none',
+				'lightbox'        => array( 'enabled' => true ),
+			)
+		);
+
+		$srcset = self::photon_srcset( $src );
```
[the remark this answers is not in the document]
thilinah: @dan-zakirov I think is a small change worth addressing.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -1,41 +1,754 @@
...
+		// Probe every screenshot for its real width / height once and
+		// pass the result down. `get_dimensions()` keeps a Memcached
+		// cache keyed by an md5 of the URL list so the network round-
+		// trip happens at most once per screenshot revision; on
+		// subsequent renders the call costs a single Memcached lookup.
+		// Knowing the dimensions lets every `<img>` ship with `width`
+		// and `height` attributes — the browser reserves slot height
+		// from the intrinsic aspect ratio, so the gallery has zero
+		// layout shift no matter how slowly the screenshots decode.
+		$dimensions = self::get_dimensions( $screenshots );
```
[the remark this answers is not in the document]
thilinah: @dan-zakirov I think it would be better to address this too.

thilinah:
I've trsted this PR on a sandbox and it mostly work as expected. Add a few comments on some Copilot suggestions.

The image gallery loads fine:
Screenshot 2026-06-05 at 14 06 59

But when I click on an image it opens a new tab with a full size image. 

Then I have to go back to the previous tab to see the light box.



https://github.com/user-attachments/assets/90dd96e1-bd4d-46a5-a787-90ea0f9be5fb


I think we should stop opening the full size image since we show the light box.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/includes/class-lightbox-captions.php
dan-zakirov: Yes this should stay. The current screenshots happen to mirror caption into alt but the polyfill should use the stored figcaption state directly. I will update the front end side for that

dan-zakirov:
@thilinah @bor0

> But when I click on an image it opens a new tab with a full size image.

Fixed in d6c2106e. The no JS fallback link stays available and the JS path now opens only the lightbox.

> This sounds like a good UX improvement to me!

Follow up fixed in d6c2106e. The image still works without JavaScript and the JavaScript path now opens only the lightbox.

thilinah (approved):
Looks good to me.

thilinah:
> The no JS fallback link stays available and the JS path now opens only the lightbox.

@dan-zakirov I see this is fixed. I have approved the PR. Could you rebase your branch onto the latest trunk and resolve the conflicts?

dan-zakirov:
Thanks @thilinah. I rebased the branch onto the latest trunk and resolved the conflicts. I also rechecked that the no-JS fallback stays available, the JS path still opens only the lightbox, and the legacy React screenshots bundle is not enqueued.

dan-zakirov:
Happy to see this make it onto WordPress.org :) Thanks @bor0, @thilinah, and @dd32 for the help at different stages. At some point this stopped being just a fix and turned into something closer to a whole three-phase story: we spent a long time thinking it through, iterated a lot, and it ended up being challenging but fun in a good way

PR #625, opened 2026-05-08
Plugin Directory: Detect wp_add_dashboard_widget() during import

dd32:
## Summary

- Adds `find_dashboard_widgets_in_file()` to the plugin importer, scanning PHP files for `wp_add_dashboard_widget()` calls.
- Assigns the `dashboard-widgets` `plugin_section` term to plugins where any call is detected, and stores each detected widget label as `dashboard_widget_name` post meta.
- The label is the first quoted string literal inside the second argument, so wrappers like `__()`, `_x()`, and `esc_html__()` are handled.

Context: https://wordpress.slack.com/archives/C0ATCQF2PMM/p1778193159754749

## Notes

- The `dashboard-widgets` term is assumed to exist on production (manual creation, same approach as `featured`, `popular`, `block`). The local environment seeds it via `environments/plugin-directory/bin/after-start.sh`.
- A backfill helper, `bin/process-dashboard-widgets.php`, is included for convenience. It pulls candidate slugs from a Veloria CSV export and re-imports them. **This script likely does not need to be merged** and can be dropped from the PR if preferred.
- Unit tests for the new helper are intentionally not included here. They will be proposed as a follow-up branch/PR alongside coverage for the existing `find_blocks_in_file()`.

## Test plan

- [ ] Run `bin/import-plugin.php elementor` (or another known dashboard-widget plugin) in a dev environment and confirm the `dashboard-widgets` `plugin_section` term is set and `dashboard_widget_name` meta is populated.
- [ ] Visit `/plugins/browse/dashboard-widgets/` and confirm the plugin lists.
- [ ] Confirm a re-import of a plugin with no `wp_add_dashboard_widget()` calls does not gain the section.


dd32:
Thanks for the review. Triage:

1. **Auto-creation of misnamed term** — leaving as-is. The pre-existing-term assumption matches how `adopt-me`, `featured`, `popular`, `block`, etc. are handled, and is the explicitly chosen approach for this PR; the term will be created on production before deploy and the local environment seeds it via `environments/plugin-directory/bin/after-start.sh`.
2. **Duplicate labels** — fixed in 3211e9a8b: `array_unique()` inside `find_dashboard_widgets_in_file()` so the helper hands back a clean list.
3. **Recursive scan cost** — accepted scope. This mirrors the fallback in `find_blocks_in_file()` (which also walks every PHP file when no `block.json` is present), so it is consistent with existing import behavior. Vendor directories are now skipped (3211e9a8b) which removes the worst-offender case.
4. **`token_get_all` for accurate parsing** — deferred. A separate PR will introduce a shared `token_get_all`-based extractor and migrate both block detection and dashboard widget detection to it at the same time, so both benefit equally.
5. **Escaped quotes in labels** — addressed by the same follow-up. Token-based extraction returns the unescaped value natively, which is the right place to handle this.

PR #626, opened 2026-05-08
Plugin Directory: Tokenise PHP source for block / dashboard widget detection

dd32:
## Summary

Replaces the regex paths in `find_blocks_in_file()` (PHP branches) and `find_dashboard_widgets_in_file()` with a single token-based extractor in a new `Tools\Tokenisation_Helpers` class. Fixes false positives inside comments, strings, and method/static calls; restores title extraction for both `register_block_type()` and `new WP_Block_Type()`; and corrects the dashboard widget label capture for plugins that pass class constants as the widget ID (e.g. Jetpack, where the previous regex picked up the word `Stats` from an adjacent doc comment).

Builds on #625 (merged).

## Helper API

`Tokenisation_Helpers` exposes a single public method that walks PHP tokens once per source and returns structured per-call data:

```php
Tokenisation_Helpers::find_function_calls( string $contents, string $function_name ): array
```

For each matched call, returns a positional list of resolved arg values. Each value is:

- **string** — literal-string value (bare, or unwrapped from a recognised i18n function: `__`, `_e`, `_x`, `_ex`, `_n`, `_nx`, `esc_html__`, `esc_html_e`, `esc_html_x`, `esc_attr__`, `esc_attr_e`, `esc_attr_x`, `translate`, `translate_with_gettext_context`).
- **array** — inline `array(...)` / `[...]` literal, recursively resolved (only string-keyed entries appear; values follow the same string|array|null contract).
- **null** — any other expression (variable, class const, method call, concat, non-i18n wrapper, etc.).

Calls inside comments, string literals, method/static-method accesses, function declarations, and constructor mismatches are skipped at the walker stage.

Prefix `$function_name` with `'new '` to match constructor calls (e.g. `'new WP_Block_Type'`).

The implementation uses `PhpToken::tokenize()` and includes a `stripos()` early-out plus a single-entry token cache so back-to-back scans of the same source for different function names tokenise only once.

## Behavioural changes

| Form | Original regex | This PR |
|---|---|---|
| Bare literal arg | matched | matched |
| Wrapped in `__`, `_x`, `esc_html__`, etc. | first quote in arg | inner literal of the i18n call |
| Class const as 1st arg + literal label later | order-sensitive; could pick text from inline doc comments | label correctly extracted regardless of preceding non-literal args |
| Inline doc comment between args | comment contents could be captured | ignored |
| Escaped quote inside literal | truncated value | correctly unescaped |
| Call inside line/block comment | false positive | ignored |
| Call inside a string literal | false positive | ignored |
| `$obj->name()`, `Class::name()`, `function name()` declaration | false positive | ignored |
| `\register_block_type()` (leading backslash) | incidental substring match | matched (`T_NAME_FULLY_QUALIFIED`) |
| `Foo\register_block_type()` (arbitrary namespace) | incidental substring match | not matched (documented shortcut) |
| `title` in second-arg options array | captured for `new WP_Block_Type` only | captured for both, including `__()`-wrapped values, in long or short array form |
| Concat name `'prefix-' . $var` | skipped via negative lookahead | resolves to `null` (no partial literal) |
| Non-i18n wrapper, e.g. `some_helper( 'X' )`, `sprintf( __( 'X' ), ... )` | inner literal could be captured | resolves to `null` |
| Variable label in `wp_add_dashboard_widget` | call dropped | call reported with empty string (section term still applies; meta is still skipped) |

## Examples

Block titles via the options array (now for both functions, long and short array form, including i18n-wrapped values):

```php
register_block_type( 'my-plugin/foo', array( 'title' => 'Foo' ) );
register_block_type( 'my-plugin/bar', [ 'title' => __( 'Bar', 'td' ) ] );
new WP_Block_Type( 'my-plugin/baz', array( 'title' => 'Baz' ) );
```

`Tokenisation_Helpers::find_function_calls( $src, 'register_block_type' )` yields:

```php
[
    [ 'my-plugin/foo', [ 'title' => 'Foo' ] ],
    [ 'my-plugin/bar', [ 'title' => 'Bar' ] ],
]
```

Class const + inline doc comment between args (the original Jetpack pattern):

```php
wp_add_dashboard_widget(
    My_Widget_Class::DASHBOARD_WIDGET_ID,
    /** This is a comment */
    'My Widget',
    array( $instance, 'render' )
);
```

yields `[ [ null, 'My Widget', [] ] ]`. The previous regex was sensitive to the contents of any inline comment between args — when the comment contained quote characters, the inner regex would match those quotes first and capture the wrong text (the original Jetpack source has `/** "Stats" is a product name. */` between the ID and the label, which is what triggered the regression).

Variable as label still tags the section:

```php
wp_add_dashboard_widget( 'id', $variable, 'cb' );
```

yields `[ [ 'id', null, 'cb' ] ]`. `find_dashboard_widgets_in_file()` converts the null label to `''`, and `import_from_svn` filters empty strings before storing `dashboard_widget_name` meta, so the call is detected (the `dashboard-widgets` `plugin_section` term is applied) but no empty meta is stored.

False positives are now ignored:

```php
// wp_add_dashboard_widget( 'x', 'Y', 'cb' );
/* wp_add_dashboard_widget( 'x', 'Y', 'cb' ); */
$str = "wp_add_dashboard_widget( 'x', 'Y', 'cb' );";
$obj->wp_add_dashboard_widget( 'x', 'Y', 'cb' );
SomeClass::wp_add_dashboard_widget( 'x', 'Y', 'cb' );
function wp_add_dashboard_widget( $id, $name, $cb ) {}
```

All return `[]`.

## Documented shortcuts

Two test cases under `Shortcut (to reduce complexity):` doc comments record deliberate trade-offs:

- Only `\register_block_type` (`T_NAME_FULLY_QUALIFIED` with a leading backslash) is treated as the global function. Arbitrary namespaced calls like `Foo\Bar\register_block_type(...)` are not matched.
- Nested wrapping calls are not recursively unwrapped through non-allowlisted functions: `esc_html( __( 'X', 'td' ) )` resolves to `null` because the outer `esc_html` is not in the translation allowlist.

## Real-world smoke test

Scanned six representative plugins (Jetpack, WooCommerce, Yoast SEO, Akismet, Gutenberg, Contact Form 7 — ~226 MB of plugin source) before and after the refactor:

| Plugin | Result |
|---|---|
| Jetpack | `"Jetpack Stats"` and `"Jetpack Newsletter"` correctly captured. (The Stats case is exactly the class-const + inline doc comment pattern that motivated this PR; the previous regex returned `"Stats"`.) |
| WooCommerce | 3 dashboard widgets correctly captured. |
| Yoast SEO | 2 calls detected; both labels resolve to `null` because Yoast wraps them in `sprintf( __( '%s Posts Overview', 'td' ), 'Yoast SEO' )`. Section term still applies; the label is genuinely dynamic and not statically capturable. |
| Gutenberg | 1 direct `register_block_type()` call. Most Gutenberg blocks go through `register_block_type_from_metadata()` / `block.json`, which is a separate detection path this PR does not touch. |
| Akismet, Contact Form 7 | No dashboard widgets, no direct block registrations. |

No false positives across the corpus.

## Test plan

- [x] 35 unit tests in `tests/Tokenisation_Helpers_Test.php` cover bare literals, i18n wrappers, false-positive prevention, non-literal args, namespaced calls, array-literal resolution (long/short, nested, mixed, translation-wrapped, non-string keys), and the two documented shortcuts.
- [x] PHPCS clean (`phpcs-branch.php` exit 0).
- [x] Real-world smoke test against the six plugins above.
- [ ] Re-import the Jetpack plugin in production and confirm `dashboard_widget_name` post meta now contains `Jetpack Stats` (was `Stats`).
- [ ] Confirm a plugin that calls `wp_add_dashboard_widget` with a `sprintf( __() )` label still gets the `dashboard-widgets` section term assigned, and no empty `dashboard_widget_name` rows are stored.


dd32:
Refactor in 72692eeee addresses two of the Copilot review threads:

- **`@token_get_all()` suppression** — dropped. The call is now `token_get_all( $contents )` with no error suppression.
- **`yield` wording vs. materialised return** — the per-call walker is now an actual `iterable` generator (`walk_calls()` uses `yield`), and the public `find_function_calls()` builds the materialised list via `array_map()`. Doc comments match.

The other open Copilot threads:
- PHP 8.0+ feature use (`str_starts_with`, `T_NAME_FULLY_QUALIFIED`, `T_NULLSAFE_OBJECT_OPERATOR`) — kept as-is. This codebase targets PHP 8.1+ (see `phpcs.xml.dist`: `testVersion=8.4-`).
- Double tokenisation in `find_blocks_in_file()` — still applies; each `find_function_calls()` call re-tokenises. The cost is small per file (microseconds), but worth revisiting if importer wall time becomes a concern.

dd32:
Performance feedback addressed in f3d8325fd:

- **`stripos()` early-out** in `Tokenisation_Helpers::find_function_calls()` — when the source does not contain the function name (case-insensitive), the method returns `[]` without calling `token_get_all()`. The check also covers the `\name(` leading-backslash form since `stripos` is a substring match.
- **Single-entry token cache** — back-to-back scans of the same file (e.g. `find_blocks_in_file()` running `register_block_type` then `new WP_Block_Type` against the same `$contents`) tokenise only once. Different `$contents` replaces the cached entry, so memory stays bounded.

Real-world smoke test across Jetpack, WooCommerce, Yoast SEO, Akismet, Gutenberg, and Contact Form 7 (~226 MB of plugin source) completes in ~0.5s wall time and produces identical output to the previous commit.

dd32:
Both Copilot points addressed:

- **`find_dashboard_widgets_in_file()` docblock** — updated in 9cace13ad to spell out the allowlist (the 14 recognised i18n functions) and call out that non-allowlisted wrappers like `sprintf( __() )` or `esc_html( __() )` resolve to an empty string while the call itself is still reported so the section term can apply.
- **PR description references obsolete API** — refreshed the PR body to describe the current single-method API (`Tokenisation_Helpers::find_function_calls()`) and removed mentions of `find_function_call_arg_strings()` / `find_function_call_first_arg_and_array_value()`.

PR #633, opened 2026-05-13
Plugin Directory: Send plugin updates to Gandalf for advisory scans

Luc45:
After a plugin import, carry the importer release context through the existing `scan_plugin:{slug}` cron job and send the current ZIP URL to Gandalf.

Gandalf callbacks are accepted through a new authenticated REST route. The Plugin Directory keeps only local integration state: pending scan correlation, latest integration error, and Slack verdict dedupe. Full scan history and reports stay in Gandalf.

This is advisory only for now, so that we can monitor false positive rate. Plugin releases are not blocked if Gandalf is unavailable or returns findings.

Reuses the existing Plugin Check cron path rather than adding a second per-plugin update scan cron job.

Testing:

- Added Gandalf dispatch/callback tests.
- Added WPORG-owned contract fixtures for Gandalf compatibility checks.
- Added integration tests on the Gandalf side that tests against those contracts.
- Verified syntax, PHPCS, and targeted PHPUnit.


dd32:
Re-opening this, as I committed (badly) a variant of this via #634, but with greatly less error handling checks.

I didn't see the point in being super strict on the payloads here, when it's only an internal service that's being called / responding.

This allows for a more gracious path going forward where the payloads can be updated independently without having to be super-strict about needing to be updated in the same second.

It might also best to update the scanner to accept any WordPress.org ZIP url as well, including our CDN's, as I'd have liked to include the `?nostats=1` to avoid bumping our internal download counters.

dd32:
Closing in favour of what was added via https://github.com/WordPress/wordpress.org/pull/634

PR #663, opened 2026-06-02
Add aria-hidden to stars

joedolson:
Partially fixes https://meta.trac.wordpress.org/ticket/8276 by adding `aria-hidden` to the ratings star output.

However, the container already has an `aria-label`. This is run through `wp_kses_post` at a few points, but I wouldn't have expected that to remove `aria-label`, so I'm not sure why it's gone.

In block contexts, this is rendered via `wporg/ratings-stars`, but I have no idea where that is.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/class-template.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/class-template.php
@@ -715,7 +715,7 @@ class Template {
 		$args = is_numeric( $args ) ? array( 'rating' => $args ) : $args;
 		$args = wp_parse_args( $args, array(
 			'rating'   => 0,
-			'template' => '<span class="%1$s"></span>',
+			'template' => '<span class="%1$s" aria-hidden="true"></span>',
 		) );
 
 		$rating         = round( $args['rating'] / 0.5 ) * 0.5;


t-hamano:
I think this PR likely aims to fix the review content on the plugin page.

rating

In this section, the rating is outside the heading, so even if the title link is focused, screen readers apparently don't read out the rating.

```html
<div class="header-top">
  <div class="wporg-ratings" aria-label="3 out of 5 stars" data-title-template="%s out of 5 stars" data-rating="3" style="color:#ffb900;">
    <span class="dashicons dashicons-star-filled"></span>
    <span class="dashicons dashicons-star-filled"></span>
    <span class="dashicons dashicons-star-filled"></span>
    <span class="dashicons dashicons-star-empty"></span>
    <span class="dashicons dashicons-star-empty"></span>
  </div>
  <h3 class="review-title">
    <a class="url" href="https://wordpress.org/support/topic/good-direction-but-still-lacks-flexibility/">Good Direction, But Still Lacks Flexibility</a>
  </h3>
</div>
```

Perhaps the solution is to inject screen reader text into the heading?

```html
<div class="header-top">
  <div class="wporg-ratings" aria-label="3 out of 5 stars" data-title-template="%s out of 5 stars" data-rating="3" style="color:#ffb900;">
    <span class="dashicons dashicons-star-filled"></span>
    <span class="dashicons dashicons-star-filled"></span>
    <span class="dashicons dashicons-star-filled"></span>
    <span class="dashicons dashicons-star-empty"></span>
    <span class="dashicons dashicons-star-empty"></span>
  </div>
  <h3 class="review-title">
    <span class="screen-reader-text">3 out of 5 stars</span>
    <a class="url" href="https://wordpress.org/support/topic/good-direction-but-still-lacks-flexibility/">Good Direction, But Still Lacks Flexibility</a>
  </h3>
</div>
```

Additionally, I found that to fix the original ticket, we will likely need to make the fix in a different repository. Here it is.

https://github.com/WordPress/wporg-theme-directory/blob/af069c6691ef7e7d93903015f13833b47ad2980e/source/wp-content/plugins/wporg-ratings/wporg-ratings.php#L221-L243

joedolson:
So many different places these things are kept... :D

t-hamano:
This PR should resolve the original issue. https://github.com/WordPress/wporg-theme-directory/pull/163

softglazee:
Checked the current state on production today, 22 August.

The profile path now carries `role="img"`, `aria-label` and `aria-hidden="true"` on all five stars, so it looks fixed, presumably by WordPress/wporg-theme-directory#163.

The review path on plugin pages is unchanged: `aria-label` present, `role` absent, and no `aria-hidden` on any of the stars. So this PR still applies.

One thing worth weighing alongside the `screen-reader-text` idea above. On the review path the `aria-label` sits on a plain `<div>` with no role, whereas the profile path has `role="img"`. A generic element does not reliably take an accessible name, so hiding the stars may leave the label unannounced rather than announced. Reading attributes only, no screen reader used.

Full attribute readings for both paths are on the ticket: https://meta.trac.wordpress.org/ticket/8276

PR #665, opened 2026-06-04
Plugin Directory: Scope the plugin card pointer cursor to the cards, not the container

derpaschi:
## What

The plugin cards grid applied `cursor: pointer` to the entire `.plugin-cards` container. As a result the pointer cursor appeared over the grid gaps and padding — areas that aren't clickable — and not just over the cards themselves.

This change scopes the rule to `.plugin-cards .is-style-cards-grid li`, which lines it up with the existing `:hover` and `:focus-within` selectors that already target the individual cards.

## Why

The pointer cursor is a clickability affordance. Showing it over the non-interactive container space is a little misleading, and it was inconsistent with the hover/focus styling that was already scoped to the cards. Tightening the selector keeps the cursor behaviour consistent with the rest of the card interactions.

## Testing

- Recompiled the theme styles; `css/style.css` and `css/style-rtl.css` are regenerated from the SCSS source.
- Verified manually: hovering a plugin card shows the pointer cursor, while hovering the space between cards shows the default cursor.

## Notes

This is a focused, presentational-only change with no markup or behaviour changes. Feedback very welcome — happy to adjust the selector if there's a preferred convention for this theme.


wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/css/style.css.map
obenland, line 1: @derpaschi Let's remove that file.

It looks like it's an unintended artifact from the sass options, probably worth a follow-up PR.

PR #674, opened 2026-06-05
Plugin Directory: Search: Only boost whole-slug matches, to prevent slug squatting. See #8225.

dd32:
Fixes the slug boost in plugin search so it rewards _searching for a plugin by its slug_ without letting plugins **squat on common keywords** via their slug. See https://meta.trac.wordpress.org/ticket/8225.

### Problem

`slug_text` was part of the high-weight (`boost: 5`) `most_fields` title clause. `slug_text` is word-tokenized — `wp-google-maps` → `wp`, `google`, `maps` — so a generic single-word query like `map` matched a _fragment_ of every `*-map(s)` slug and collected the full slug boost. This is a **slug-squatting** vector: stuffing a popular term into a slug buys ranking on that term.

Concretely, for the query `map`, every plugin with `map` as one slug word got the boost, and the only top results _without_ it (`wp-google-maps`, `google-maps-widget`, whose slugs contain `maps`, not `map`) were the ones penalised — the opposite of intent.

### Fix

Match the query against the whole-slug `slug` keyword field, normalised with `sanitize_title()` (the same function that generated the slug), via a `constant_score` clause. The slug boost now fires only when the query _is_ the plugin's identifier, not when it shares one word with it. `slug_text` is removed from the title `most_fields` clause; title matching is unchanged.

### Behaviour — slug-searches still match the expected plugin

| Search query | Normalised | Result |
|---|---|---|
| `map` | `map` | **no slug boost** (no plugin's whole slug is `map`) — squatting prevented |
| `maps` | `maps` | **no slug boost** — squatting prevented |
| `wp-google-maps` | `wp-google-maps` | boosts `wp-google-maps` ✓ |
| `WP Google Maps` (spaces) | `wp-google-maps` | boosts `wp-google-maps` ✓ |
| `wordpress-seo` | `wordpress-seo` | boosts `wordpress-seo` ✓ |
| `leaflet-map` | `leaflet-map` | boosts `leaflet-map` ✓ |
| `contact-form-7` | `contact-form-7` | boosts `contact-form-7` ✓ |
| `woocommerce` | `woocommerce` | boosts `woocommerce` ✓ (single-word brand slugs still match) |

Verified against the live index: single common words (`map`/`maps`) match zero whole-slugs, while every real slug search returns exactly its plugin with a flat, predictable score.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-search.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-search.php
@@ -274,15 +274,42 @@ class Plugin_Search {
 			],
 		];
 
-		// A direct slug match
+		// Boost on a title match.
 		$should_match[] = [
 			'multi_match' => [
 				'query'  => $search_phrase,
-				'fields' => $this->localise_es_fields( [ 'title', 'slug_text' ] ),
+				'fields' => $this->localise_es_fields( [ 'title' ] ),
 				'type'   => 'most_fields',
 				'boost'  => 5,
 			],
 		];
+
+		// A direct slug match.
+		//
+		// The slug boost is meant to reward searches that are looking for a
+		// specific plugin by its slug (e.g. `wordpress-seo`, `wp-google-maps`),
+		// not searches for a topic that happens to appear as one word of a slug.
+		//
+		// Previously `slug_text` was included in the title `most_fields` match
+		// above. Because `slug_text` is word-tokenized (`wp-google-maps` becomes
+		// `wp`, `google`, `maps`), a generic single-word query like `map` matched
+		// a fragment of every `*-map(s)` slug and collected the full slug boost.
+		// That let plugins squat on common terms by stuffing them into the slug.
+		//
+		// Instead we match the query against the whole-slug `slug` keyword field,
+		// normalised the same way the slug itself was generated. `sanitize_title()`
+		// turns `WP Google Maps` into `wp-google-maps`, so an actual slug search
+		// still scores a decisive, constant boost, while a single common word
+		// only matches a plugin whose entire slug is that word.
+		$slug_candidate = sanitize_title( $search_phrase );
+		if ( $slug_candidate ) {
+			$should_match[] = [
+				'constant_score' => [
+					'filter' => [ 'term' => [ 'slug' => $slug_candidate ] ],
+					'boost'  => 8,
+				],
+			];
+		}
 
 		$should_match[] = [
 			'multi_match' => [


dd32:
### Optional follow-up: a forgiveness tier for multi-word slug searches

This PR deliberately implements only the **exact whole-slug** tier, which keeps the anti-squatting property airtight. The trade-off is that the slug boost is strict: the query has to normalise to the _entire_ slug. A search that drops or adds a word — e.g. `google maps widget pro` for the slug `google-maps-widget`, or `wp google maps openstreetmap` for `wp-google-maps` — gets no slug boost and falls back to title/content/popularity scoring.

If we decide that's too strict, we could add a **second, lower-weighted tier** gated to multi-word queries:

```php
if ( str_contains( trim( $search_phrase ), ' ' ) ) {  // multi-word only
    $should_match[] = [
        'match_phrase' => [
            'slug_text' => [
                'query' => $search_phrase,
                'boost' => 3,   // below the exact-slug tier (8)
            ],
        ],
    ];
}
```

**Why we might want it:**

- **Forgiveness for near-miss slug searches.** Users often paste/type a slug with extra or missing words, or in a slightly different order. A phrase match on the word-tokenized `slug_text` still rewards a strong in-order overlap (`google maps` → `google-maps-widget`) without requiring an exact identifier match.
- **Bridges the gap between "exact slug" and "title relevance."** Today everything between those two extremes is unrewarded; a moderate phrase boost fills it.

**Why it's _not_ in this PR (the reasons to be cautious):**

- **It partially reopens the squatting surface.** A `match_phrase` on `slug_text` still scores _fragments_ of a slug — just multi-word ones. A plugin could squat slug `best-seo-plugin` and pick up a boost for the query `seo plugin`. The single-word gate only raises the bar from one word to two; it doesn't close the hole.
- **Two-word phrases are far less gameable than single words**, and the boost is lower (`3` vs `8`), so the risk is smaller — but it's a genuine policy call, not a clean win, which is why it belongs in discussion rather than bundled into the squatting fix.
- **Keeps this PR's claim simple.** "Slug boost requires the whole slug" is easy to reason about and verify; adding partial matching makes the behaviour fuzzier to audit.

Happy to add the tier (with tests/examples) in this PR or a follow-up if reviewers think the strictness is a problem in practice. My suggestion: ship the strict version, watch for complaints about legitimate multi-word slug searches not ranking, and add the tier only if the data shows it's needed.

dd32:
@gibrown Any thoughts on this? This PR is AI generated by me telling AI my understanding of the reason we do slug matching, and the concerns I have about the slug matching that was added back in.

I think this is a good middle-ground that seems to be reasonable to me.

PR #689, opened 2026-06-09
Plugin Directory: lightbox image is rendered smaller than expected

thilinah:
## Summary

The screenshot gallery lightbox (added in #622) had two production-only bugs:

1. **Image too small** — the enlarged image was capped at the 900px Photon
   thumbnail instead of showing the full-resolution screenshot.
2. **Image duplicated** — the screenshot rendered twice, stacked vertically.

Both are fixed here. They were related: fixing (1) is what surfaced (2).

## Background

Plugin screenshots are external `ps.w.org` assets with no real attachment, so each `core/image` block is given a *synthetic* attachment ID. Core's lightbox renderer treats that ID as a genuine attachment:

So the lightbox had no full-resolution source to enlarge and no real dimensions, falling back to the thumbnail's Photon-shrunk size.

This was not surfaced during sandbox testing because the Photon srcset is gated to production only  photon_srcset() bails on non-production environments.

With no srcset on the sandbox, the thumbnail  loads the full-resolution ps.w.org original, so the thumbnail's natural size is the full image and the lightbox looked correct. Only on production, where Photon caps the thumbnail at ≤900px, did the bug appear — and the duplicate-image bug below was likewise masked because the enlarged container had no source to render.

## Fix

1. Repair the lightbox metadata (class-screenshots.php)

A new render_block_core/image filter (priority 20, after core's priority-15 lightbox pass) overrides the interactivity state for our screenshot blocks with the real uploadedSrc, targetWidth/targetHeight, and clears lightboxSrcset so the enlarged view loads the lossless original instead of a capped Photon candidate. Core keys this state by a per-render uniqid(), so the filter reads that key back out of the rendered markup. wp_interactivity_state() merges later-call-wins, so the corrected values replace the broken ones. Applied on all environments so sandbox and production behave identically.

2. Stop stacking the lightbox's two image containers (gallery-lightbox-enhancements)

The caption polyfill forced core's two center-overlapping .lightbox-image-container siblings to position: relative, dropping them into normal flow and stacking them vertically. This only ever looked correct because the enlarged container was empty (bug 1 above). Once that container held a real image, the screenshot rendered twice.

- CSS: removed the position: relative override so core's absolute center-overlap is restored.
- JS: the caption is now mounted into the top (full-resolution) container so it isn't occluded by the enlarged image.

## Testing

- Before startign to test, comment out this line: http://github.com/WordPress/wordpress.org/blob/53180a2d8dec3ab9ad4acc5dea5d732e667b9e1c/wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php#L588
- Visit a plugin page e.g: https://wordpress.org/plugins/woocommerce/
- The enlarged image is shown on lightbox.
- The image renders once, not stacked/duplicated.
- Captions still pin to the bottom of the picture. (e.g: https://wordpress.org/plugins/wordpress-seo/)

## Closes / refs
- Closes https://meta.trac.wordpress.org/ticket/8285

## Screenshots

Before:

production-lightbox

After:

Screenshot 2026-06-08 at 22 37 34

Screenshot 2026-06-08 at 22 43 51

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.css
+++ b/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.css
@@ -42,7 +42,7 @@
 	margin: 0;
 	padding: 3.5em 1.25em 1em;
 	color: #fff;
-	text-align: center;
+	text-align: start;
 	font-size: 0.95rem;
 	font-weight: 500;
 	line-height: 1.45;
--- a/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.css
+++ b/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.css
@@ -25,16 +25,12 @@
 }
 
 /*
- * Do NOT override `.lightbox-image-container` positioning. Core stacks
- * its two containers — the zoom-thumbnail and the full-resolution image
- * — as absolutely positioned, center-overlapping siblings, with the
- * full-resolution one painted on top. Forcing them to `position:
- * relative` drops both into normal flow and stacks the images
- * vertically, rendering the picture twice the moment both containers
- * hold a real image (e.g. plugin screenshots, whose lightbox source is
- * wired up server-side). The caption is mounted into the top container
- * and pins to its frame, using that absolute container as its
- * containing block — no positioning override required here.
+ * Do NOT set `.lightbox-image-container` to `position: relative`. Core
+ * stacks its two containers (zoom-thumbnail + full-res) as absolutely
+ * positioned, overlapping siblings; forcing them into normal flow stacks
+ * the images vertically and renders the picture twice once both hold a
+ * real image (e.g. plugin screenshots). The absolute container is already
+ * the caption's containing block — no override needed.
  */
 
 .wp-lightbox-overlay .lightbox-image-container > figcaption.wp-lightbox-caption {
--- a/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.css
+++ b/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.css
@@ -12,22 +12,30 @@
  */
 
 /**
- * The core zoom-mode `.scrim` is white at 0.9 opacity by default, but
- * core CSS doesn't pin it (no `top`/`left` declared). With the two
- * `.lightbox-image-container` siblings each `position: relative; top:
- * 450px`, the absolute scrim falls through to its normal-flow position
- * and lands somewhere around `y: 1444px` — visually below the viewport
- * fold. Pin it to the overlay's top-left so the dimmer actually covers
- * the page as intended.
+ * Defensive pin for the core zoom-mode `.scrim` (white, 0.9 opacity).
+ * Core leaves it `position: absolute` with no `top`/`left`, relying on
+ * its two `.lightbox-image-container` siblings being out of normal flow
+ * so the scrim's static position resolves to the overlay's top-left.
+ * We keep those containers on core's own positioning (see note below),
+ * so this is belt-and-braces — harmless if core already covers the page.
  */
 .wp-lightbox-overlay .scrim {
 	top: 0 !important;
 	left: 0 !important;
 }
 
-.wp-lightbox-overlay .lightbox-image-container {
-	position: relative;
-}
+/*
+ * Do NOT override `.lightbox-image-container` positioning. Core stacks
+ * its two containers — the zoom-thumbnail and the full-resolution image
+ * — as absolutely positioned, center-overlapping siblings, with the
+ * full-resolution one painted on top. Forcing them to `position:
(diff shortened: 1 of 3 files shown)


wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.css
```diff
@@ -12,22 +12,26 @@
...
  */
 .wp-lightbox-overlay .scrim {
 	top: 0 !important;
 	left: 0 !important;
 }
 
-.wp-lightbox-overlay .lightbox-image-container {
-	position: relative;
-}
+/*
```
thilinah, line 27: As suggested by @bor0 here (https://github.com/WordPress/wordpress.org/pull/686#pullrequestreview-4451000032), I've trimmed the comment. But it'll be useful to keep trimmed down version of this comment.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -396,6 +433,85 @@ protected static function build_image_block( $screenshot, $id, $above_fold = fal
...
+		if ( $id < self::SYNTHETIC_ID_OFFSET || ! isset( self::$lightbox_meta[ $id ] ) ) {
+			return $block_content;
+		}
+
+		if ( ! function_exists( 'wp_interactivity_state' ) ) {
+			return $block_content;
+		}
+
+		// Recover the render-time state key core assigned to this image.
+		$processor = new \WP_HTML_Tag_Processor( $block_content );
```
thilinah, line 483: >Question from @bor0 on original PR:
> Is there a way to make the initial lightbox render properly? This kind of fixing may cause issues in the future

Thanks for raising this, @bor0 — you're right that it looks more convoluted than it should, and that's down to how plugin screenshots differ from normal images.

Plugin screenshots are external ps.w.org files, not WordPress attachments — there's no attachment post behind them. Core's lightbox builds its enlarged-image source and dimensions from wp_get_attachment_url() and wp_get_attachment_metadata(), but both of those return false and exit before their filters run when the attachment post doesn't exist. So there's no hook where we can hand core the correct values up front, which is why the current code fixes up the rendered output afterward instead.

Claude suggested this long term solution:

> The truly proper long-term fix is upstream: a Gutenberg/core filter that lets a block supply lightbox metadata for external (non-attachment) images

For this PR I would skip implementing it as we are addresing a bug, and there is some urgency to push this.

I'll open a seperate trac issue for a long term solution.

thilinah:
Suggestion from @alanjacobmathew on [closed PR](https://github.com/WordPress/wordpress.org/pull/686#issuecomment-4653590330)

> The 'wp-lightbox-caption', better if it is placed outside the image. That will make sure the screenshots can follow the brand-guidelines too if needed. Right now, if the image has a white shade, the caption becomes nearly unreadable. And most screenshot taken of the admin-screen is nearly white(Take the example of Yoast SEO screenshot above)

Thanks for the suggestion @alanjacobmathew. Since the fix for the original issue (small lightbox image) is a bit urgent, I'll differ this to another PR.

bor0 (approved):
Tests well! Let's address that comment soon though before it breaks anything.

Thanks!

Bovelett:
Thank you, @thilinah! Where can I find that PR / Issue for the text outside the image? A second point to that is that the text should be left-aligned. Centered it's hard to read for people with dyslexia (I'm one of the 15% of all users this affects as soon as it's more than one line).

> Suggestion from @alanjacobmathew on [closed PR](https://github.com/WordPress/wordpress.org/pull/686#issuecomment-4653590330)
> 
> > The 'wp-lightbox-caption', better if it is placed outside the image. That will make sure the screenshots can follow the brand-guidelines too if needed. Right now, if the image has a white shade, the caption becomes nearly unreadable. And most screenshot taken of the admin-screen is nearly white(Take the example of Yoast SEO screenshot above)
> 
> Thanks for the suggestion @alanjacobmathew. Since the fix for the original issue (small lightbox image) is a bit urgent, I'll differ this to another PR.

cc: @stevejonesdev

thilinah:
> A second point to that is that the text should be left-aligned. Centered it's hard to read for people with dyslexia

Thanks for bringing this up @Bovelett. I've updated the text alignment based on locale, to either left or right.

Screenshot 2026-06-10 at 11 45 04

Screenshot 2026-06-10 at 11 44 37




> Where can I find that PR / Issue for the text outside the image?

I'll open the new PR after merging this.

amberhinds:
@thilinah re: 
> Since the fix for the original issue (small lightbox image) is a bit urgent, I'll differ this to another PR.

It's safe to say that the caption readability is a larger failure than the image size. Small images are annoying, unreadable text is a literal blocker, and a failure of the Web Content Accessibility Guidelines 2.2 standard that WordPress says it will meet.

Here is an example of an image on my plugin, [Accessibility Checker:](https://wordpress.org/plugins/accessibility-checker/)
Screenshot on the Accessibility Checker plugin with a caption overlaid the bottom half of the image in white text that is completely unreadable.

This issue that @alanjacobmathew reported should be treated as urgent.

amberhinds:
There's a ticket from @stevejonesdev for the caption [issue in meta trac here.](https://meta.trac.wordpress.org/ticket/8289)

PR #690, opened 2026-06-09
Environments: Add support forums environment.

Clorith:
This introduces the support forums as an environment in the WordPress.org suite of sites. Complete with full dependency setup, including:

- Multisite setup
- Plugin and theme directory subsites (scaffolds for plugin and theme support forums, not for their development)
- Full set of every usertype available
- Default forum creation, mimicking the current WordPress.org forums

There is a caveat that Gutenberg is _currently_ pinned to version 17.9, due to the Blocks Everywhere plugin, and that Blocks Everywhere is not rendering properly pending an update in https://github.com/Automattic/blocks-everywhere/pull/211

## Summary by CodeRabbit

* **New Features**
  * Added a local WordPress.org Support Forums environment with multisite support, seeded forums, topics, replies, users, and directory content.
  * Added local support for forum ratings, profile associations, two-factor account links, and safe mail handling.
  * Added commands for starting, refreshing, and testing the Support Forums environment.

* **Documentation**
  * Documented setup, usage, network structure, seeded content, and local environment limitations.

* **Tests**
  * Added automated test configuration and coverage for Support Forums functionality, including forum block restrictions.


environments/support-forums/bin/after-start.sh
```diff
@@ -0,0 +1,140 @@
...
+$WP wp post create --post_type=plugin --post_status=publish --post_author=4 --post_title='Hello Dolly' --post_date='2022-08-20 01:00:00' --post_modified='2022-08-20 01:00:00' --post_modified_gmt='2022-08-20 01:00:00' --url=localhost:8888/plugins
+
+###
+# Set up the `themes` sub-site with associated content.
+###
+
+echo "Setting up minimum required themes sub-site..."
+
+# Set up the site identity.
+$WP wp option update blogname 'WordPress.org Forums'
```
obenland:
```suggestion
$WP wp option update blogname 'WordPress.org Themes' --url=localhost:8888/themes
```
Is this meant to be for themes?

environments/support-forums/bin/after-start.sh
```diff
@@ -0,0 +1,140 @@
+#!/bin/bash
+#
+# Runs after wp-env start. Sets up permalinks, creates pages, and imports plugins.
+#
+
```
obenland, line 5:
```suggestion

set -euo pipefail

```

Let's make failures debuggable in this script

environments/support-forums/bin/after-start.sh
Clorith: Ahhh no, this is the title of the main site (the support forums), it's just me having placed it in an odd location, the Themes and Plugin subsites that are dependencies are set up at L28 and L31 with simplified titles

obenland: Hm, this appears to be in the wrong place then? It's in the "Set up the `themes` sub-site with associated content." section.

environments/support-forums/bin/.htaccess
```diff
@@ -0,0 +1,15 @@
...
+# Any changes to the directives between these markers will be overwritten.
+<IfModule mod_rewrite.c>
+RewriteEngine On
+RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+RewriteBase /
+RewriteRule ^index\.php$ - [L]
+RewriteCond %{REQUEST_FILENAME} !-f
+RewriteCond %{REQUEST_FILENAME} !-d
+RewriteRule . /index.php [L]
+</IfModule>
```
obenland: These are single-site rewrite rules, but the environment is a subdirectory multisite.

environments/support-forums/bin/after-start.sh
```diff
@@ -0,0 +1,142 @@
...
+fi
+
+###
+# Create multisite framework
+###
+echo "Setting up multisite framework..."
+$WP wp core multisite-convert
+
+# Network enable the themes
+$WP wp theme enable wporg-support-2024
```
obenland: Looks like it's missing the `--network` flag to enable it network wide

environments/support-forums/bin/seed.php
```diff
@@ -0,0 +1,523 @@
...
+	/*
+	 * Production serves the archive from /forums/ but each forum from
+	 * /forum/<slug>/. bbPress only drops the root slug from those child
+	 * permalinks when the root is excluded.
+	 */
+	if ( $forums ) {
+		update_option( '_bbp_include_root', false );
+	}
+
+	flush_rewrite_rules( false );
```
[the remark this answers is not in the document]
obenland: Fixed in acd444f1c — confirmed, and it was worse than described.

Before the fix, the forums' rules were stored against both directory sub-sites:

```
blog 1 forums   rules=260  forum-ish=95   plugin-cpt=0
blog 2 plugins  rules=285  forum-ish=128  plugin-cpt=0
blog 3 themes   rules=285  forum-ish=128  plugin-cpt=0
```

`/themes/twentytwentyfour/` was a 404 as a result.

I took a different remedy than suggested: rather than moving the flush into each seeded path in `seed-site.php`, `configure_site()` now calls `delete_option( 'rewrite_rules' )` inside the switch, so each site regenerates on its first request with its own plugins loaded. One call site instead of three, and it cannot drift when another sub-site is added. Core takes the same care in `wp_install_defaults()`, which reinitialises `$wp_rewrite` before flushing.

After:

```
blog 2 plugins  rules=113  forum-ish=0   plugin-cpt=5
blog 3 themes   rules=96   forum-ish=0
blog 4 rosetta  rules=213  forum-ish=85
```

Fixing the routing also exposed a second fault the 404 had been hiding: the seeded repopackage carried no versioned meta, and `Themes_API::get_theme()` indexes `_screenshot` by version without checking its shape, so the theme page became a TypeError the moment it resolved. The seed now writes the keys `import-themes.php` writes. Both directory permalinks render clean.

.github/unit-tests-suites.yml
```diff
@@ -124,6 +124,13 @@ suites:
     paths:
       - 'environments/translate/**'
 
+  support-forums:
+    type: wordpress
+    name: Support Forums
+    script: support:test
+    paths:
+      - 'environments/support-forums/**'
+
```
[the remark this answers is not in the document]
obenland: Not applied — the workflow already derives that exact path, so adding it would duplicate it.

`.github/workflows/unit-tests.yml` lines 47-51 append each wordpress suite's activated plugins to its own path filter:

```bash
# Add each wordpress suite's environment plugins to its paths so the two can't drift.
plugins=$( jq -c '[ .plugins[] | select( startswith( "../" ) ) | sub( "^\\.\\./"; "" ) + "/**" ]' \
  "environments/$suite/.wp-env.test.json" )
suites=$( jq -c --arg suite "$suite" --argjson plugins "$plugins" '.suites[ $suite ].paths += $plugins' <<< "$suites" )
```

Run verbatim against `environments/support-forums/.wp-env.test.json` that emits:

```
["wordpress.org/public_html/wp-content/plugins/support-forums/**"]
```

which is the suggested addition character for character, and it covers `tests/` and the bootstrap under the same glob.

`plugin-directory` relies on the identical mechanism and lists only `environments/plugin-directory/**`. The registry header documents it, and `make` is the one entry that adds a plugin path by hand, with the note explaining why: "Mounted for its suite but not activated, so not read from the environment's plugin list." The support forums plugin *is* activated, so it is read automatically.

obenland:
> Test constructor-registered filters. — `bootstrap.php:48-54`

Valid, and fixed in 368b9e29a.

The premise checks out: every behaviour test calls `limit_blocks()` or `block_pre_render()` on the instance directly, so nothing covered the wiring. `Blocks` is also instantiated outside the `is_main_forums` conditional (gated on Blocks Everywhere instead), so a single-site bootstrap reaches it, as noted.

Added `test_constructor_registers_the_content_filters()` to `Block_Limits_Test`, asserting the `bbp_new_*`/`bbp_edit_*_pre_content` and `bbp_get_*_content` registrations for all three types plus `pre_render_block`. It asserts the priorities rather than just presence, since those are part of the contract: 100 puts the limiter after bbPress' own content filters and before storage, and 7 puts it before Blocks Everywhere renders at 8.

Verified it guards rather than merely passes — removing the `bbp_get_{$type}_content` registration fails it with `Failed asserting that false is identical to 7`. Suite is at 14 tests / 31 assertions, green.

One correction to the location: the test class lives in `tests/Block_Limits_Test.php`, not `tests/bootstrap.php:48-54` — line 48 of the bootstrap is the tests-directory lookup, not the class under discussion.

PR #693, opened 2026-06-12
Fix Photon caching for Plugin Directory screenshots

dan-zakirov:
This keeps Photon for responsive Plugin Directory screenshot thumbnails while sourcing the variants directly from plugins.svn.wordpress.org.

The screenshot revision is forwarded through Photon's q parameter, so replaced assets receive a new cache key. The direct ps.w.org src and full-size lightbox URL remain unchanged.

Fixes https://meta.trac.wordpress.org/ticket/8331.

Based on https://make.wordpress.org/systems/2026/07/20/writable-storage-behind-ps-w-org-s-w-org/.

Tested locally with https://github.com/WordPress/wordpress.org/pull/697 to confirm thumbnails, captions, navigation, and full-size lightbox sources remain compatible.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
@@ -363,7 +363,7 @@ class Screenshots {
 			)
 		);
 
-		$srcset = self::photon_srcset( $src );
+		$srcset = self::photon_srcset( $screenshot );
 		$class  = 'wp-block-image size-large';
 
 		// Record the full-resolution source and intrinsic dimensions for the
@@ -555,40 +555,24 @@ class Screenshots {
 	}
 
 	/**
-	 * Builds a Photon-powered `srcset` (and matching `sizes`) attribute string
-	 * for a `ps.w.org` screenshot URL. Returns an empty string when the source
-	 * URL is not on `ps.w.org`, so the original `src` is used unchanged.
+	 * Builds a revision-aware Photon `srcset` for a screenshot.
 	 *
-	 * Plugin authors upload screenshots at full resolution but we render them
-	 * inside a 3-column grid, so the browser otherwise downloads the full
-	 * asset (often 300–800 KB) only to scale it down to a ~250 px tile.
-	 * Routing the URL through `i0.wp.com` (Photon) returns a re-encoded,
-	 * width-bound copy at ~10× smaller payload — see
-	 * https://developer.wordpress.com/docs/photon/ for the resize/optim
-	 * options. The grid thumbnail therefore loads a small Photon candidate;
-	 * the lightbox is pointed back at the full-resolution `ps.w.org` original
-	 * separately by {@see self::fix_lightbox_metadata()} so users still get
-	 * the lossless image when they enlarge a screenshot.
-	 *
-	 * @param string $src Original asset URL.
+	 * @param array $screenshot Screenshot metadata.
 	 * @return string Attribute fragment ready to interpolate into `<img>`,
 	 *                including the leading space, or empty string.
 	 */
-	protected static function photon_srcset( $src ) {
-		if ( ! preg_match( '#^https?://ps\.w\.org/#', $src ) ) {
-			return '';
-		}
-
-		// Photon (i0.wp.com) only runs on production and staging. In local
-		// or other environments the proxy may not be reachable, which would
-		// leave the gallery silently empty until the cold cache warmed up.
-		// Fall back to the unoptimised `ps.w.org` URL there.
+	protected static function photon_srcset( $screenshot ) {
 		$env = function_exists( 'wp_get_environment_type' ) ? wp_get_environment_type() : 'production';
 		if ( 'production' !== $env && 'staging' !== $env ) {
 			return '';
 		}
 
-		$photon_base = preg_replace( '#^https?://#', 'https://i0.wp.com/', $src );
+		$source_url   = Template::get_asset_url( null, $screenshot, false );
+		$source_query = wp_parse_url( $source_url, PHP_URL_QUERY );
+		$photon_base  = str_replace( 'https://', 'https://i0.wp.com/', remove_query_arg( 'rev', $source_url ) );
+
+		// Photon only forwards the source query string when it is passed through `q`.
+		$photon_base = add_query_arg( 'q', $source_query, $photon_base );
 		$widths      = array( 300, 600, 900 );
 		$srcset      = array();
 


medavidallsop:
I also noticed this on a few plugins. Added info to Slack:

> On a plugin, I updated the screenshot assets, im seeing after 48 hours the thumbnails on the plugin page still haven't updated, but if you enlarge the screenshots they show the updated images - the enlarged screenshots come from [ps.w.org](http://ps.w.org/) and they are fine, but the thumbs come from [i0.wp.com/ps.w.org](http://i0.wp.com/ps.w.org) and those remain the old ones

https://wordpress.slack.com/archives/C02QB8GMM/p1784192370772129

obenland:
@dan-zakirov @thilinah @bor0 Any chance this could be prioritized? It keeps getting reported. https://meta.trac.wordpress.org/ticket/8331

bor0:
Hey @obenland, @dan-zakirov and I already discussed this in DM. Appreciate the nudge.

This PR fixes staleness by dropping Photon, but that means full-res screenshots (500KB-1MB+) load in the grid, which risks LCP on this page.

Meanwhile, Dan checked the importer and we can generate revisioned 300/600/900px thumbnails whenever a screenshot changes. Only open question, AFAIK, is storage: ps.w.org serves straight from plugin SVN, and Dan couldn't find a writable path for generated assets.

Do you happen to know if we have one behind ps.w.org or s.w.org? Once that's sorted, Dan can start right away.

obenland:
> Do you happen to know if we have one behind ps.w.org or s.w.org?

I'm afraid I don't know. https://make.wordpress.org/systems/ would be a good place to ask and make it easy to find in the future.

bor0:
Good call, thanks @obenland!

https://make.wordpress.org/systems/2026/07/20/writable-storage-behind-ps-w-org-s-w-org/

cc @dan-zakirov

bor0:
Got feedback from Systems: https://make.wordpress.org/systems/2026/07/20/writable-storage-behind-ps-w-org-s-w-org/

Barry's take is that this PR has it backwards: keep Photon (dropping it loses resizing/UX), and fix the cache-busting by pointing it at plugins.svn.wordpress.org with the revision baked into the query string instead. This way, we won't need new storage.

Let's rework this PR in that direction. Roughly, we should:

- Revert removing `photon_srcset()` and the resource hints
- Point Photon at `plugins.svn.wordpress.org` instead of `ps.w.org`
- Keep `rev` in the query string for cache-busting
- Leave the plain `src`/lightbox URL unchanged
- Verify on sandbox that a revision bump actually busts the thumbnail

@dan-zakirov, lmk if you need any help with it.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -555,40 +555,52 @@ protected static function wrap_with_show_all_button( $rendered_gallery, $count )
...
+			! is_array( $src_parts ) ||
+			empty( $src_parts['scheme'] ) ||
+			! in_array( strtolower( $src_parts['scheme'] ), array( 'http', 'https' ), true ) ||
+			empty( $src_parts['host'] ) ||
+			'ps.w.org' !== strtolower( $src_parts['host'] ) ||
+			empty( $src_parts['path'] ) ||
+			empty( $src_parts['query'] )
+		) {
 			return '';
 		}
```
obenland: This feels over complicated. Screenshot urls are created in `Template::get_asset_url()` and are in our control.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -555,40 +555,52 @@ protected static function wrap_with_show_all_button( $rendered_gallery, $count )
...
-		// or other environments the proxy may not be reachable, which would
-		// leave the gallery silently empty until the cold cache warmed up.
-		// Fall back to the unoptimised `ps.w.org` URL there.
+		$query_args = array();
+		wp_parse_str( $src_parts['query'], $query_args );
+		$revision = isset( $query_args['rev'] ) && is_string( $query_args['rev'] ) ? $query_args['rev'] : '';
+
+		if ( '' === $revision || ! ctype_digit( $revision ) || (int) $revision < 1 ) {
+			return '';
+		}
```
obenland: Same as above, rev is always there.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -555,40 +555,52 @@ protected static function wrap_with_show_all_button( $rendered_gallery, $count )
...
+
+		/*
+		 * Photon only runs on production and staging. Other environments use
+		 * the original `ps.w.org` source so local galleries do not depend on
+		 * the external image proxy.
+		 */
```
obenland:
```suggestion
```
Not helpful.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -555,40 +555,52 @@ protected static function wrap_with_show_all_button( $rendered_gallery, $count )
...
 		$env = function_exists( 'wp_get_environment_type' ) ? wp_get_environment_type() : 'production';
 		if ( 'production' !== $env && 'staging' !== $env ) {
 			return '';
 		}
 
-		$photon_base = preg_replace( '#^https?://#', 'https://i0.wp.com/', $src );
+		$photon_base = 'https://i0.wp.com/plugins.svn.wordpress.org' . $src_parts['path'];
+		$photon_base = add_query_arg( 'q', 'rev=' . $revision, $photon_base );
```
obenland: What does `q` do? Is `rev=` needed?

obenland (changes requested):
There should be a way to simplify this. `Template::get_asset_url( null, $screenshot, false )` returns the `plugins.svn.wordpress.org`-based screenshot URL, instead of CDN'ing the screenshots and then replacing all that, can it be done more elegantly up the call chain somewhere?

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: `q` carries the original source query string through Photon. The value needs to be `rev=<revision>`, rather than just the revision number, because Photon expects the complete source query. [Jetpack handles it the same way](https://github.com/Automattic/jetpack/blob/926f30fcd22925b190eb2cdd24546df66b061cad/projects/packages/image-cdn/src/class-image-cdn-core.php#L220-L239)

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: Thanks, good point. My first pass was rushed, so I overcomplicated this. It now uses `Template::get_asset_url()` directly.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: Removed the redundant checks and now rely on the revision from `Template::get_asset_url()`.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: Removed the comment.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: The code now passes the query returned by `Template::get_asset_url()` through `q` directly.

obenland (approved):
Thanks, `srcset` works much better now. Is it expected that `src` is still using `ps.w.org`?

dan-zakirov:
> Thanks, `srcset` works much better now. Is it expected that `src` is still using `ps.w.org`?

Yes, this is expected. The browser uses the Photon candidates from `srcset` for thumbnails, while `src` stays on `ps.w.org` as the cached fallback and full-size lightbox source

PR #695, opened 2026-06-13
Plugin/Theme directory - calculate sanitize_active_installs dynamically

### LINKED ISSUE #694: refactor sanitize_active_installs
We can round active installs dynamically without an if else loop.

timse201:
fixes #694


obenland:
Thanks for the suggestion, but let's not move forward with this. The existing version is about 2.3× faster and capped the result (10M for plugins, 1M for themes), which is part of the intention of these functions.

timse201:
I never thought it would cost so much more.

PR #697, opened 2026-06-14
Plugin Directory: Improve screenshot lightbox captions

dan-zakirov:
As part of https://meta.trac.wordpress.org/ticket/8289, this updates screenshot gallery lightbox captions in the Plugin Directory.

Long captions should not cover the image, tall screenshots should stay within the viewport, and short captions should keep the clean centered presentation that works well for most plugins.

Overall, the opened gallery view is improved while keeping the existing Gallery block output, navigation, and image behavior intact.

| Case | Screenshot | Notes |
| --- | --- | --- |
| Short caption | 01-yoast-seo-one-line-caption | Keeps the simple centered caption style |
| Two line caption | 02-accessibility-checker-two-line-caption | Shows longer text without covering the image |
| Long caption | 04-accessibility-checker-long-paragraph-caption | Covers the main reported case from the ticket |
| Vertical screenshot | 05-accessibility-checker-vertical-caption | Checks tall screenshots with captions below the image |

Captions stay centered to avoid breaking the presentation for other plugins.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
+++ b/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
@@ -707,19 +707,17 @@
 		overlay.wporgCaptionSyncTimer = window.setTimeout( function () {
 			overlay.wporgCaptionSyncTimer = null;
 			markOverlaySettled( overlay );
-			clearOverlayMetricCache( overlay );
-			syncCaption( overlay, false );
 			scheduleCaptionReveal( overlay, ensureCaption( overlay ) );
 		}, CAPTION_REVEAL_DELAY );
 	}
 
 	/**
-	 * Queues a two-phase sync so captions do not mutate the lightbox geometry
-	 * while core is still animating the current image.
+	 * Delays the caption reveal until core finishes the opening animation.
 	 *
-	 * @param {HTMLElement} overlay The `.wp-lightbox-overlay` element.
+	 * @param {HTMLElement} overlay       The `.wp-lightbox-overlay` element.
+	 * @param {boolean}     forceGeometry Whether to recalculate during a pending reveal.
 	 */
-	function requestSettledSync( overlay ) {
+	function requestSettledSync( overlay, forceGeometry ) {
 		var shouldSyncImmediately = isOverlaySettled( overlay );
 
 		if ( ! isOverlayActive( overlay ) ) {
@@ -742,22 +740,18 @@
 			return;
 		}
 
-		if ( shouldSyncImmediately ) {
-			if ( isCaptionRevealPending( overlay ) ) {
-				var pendingKey = overlay.dataset.wporgCaptionRevealKey || '';
-				var currentKey = getActiveImageKey( overlay );
+		if ( isCaptionRevealPending( overlay ) ) {
+			var pendingKey = overlay.dataset.wporgCaptionRevealKey || '';
+			var currentKey = getActiveImageKey( overlay );
 
-				clearOverlayMetricCache( overlay );
-				syncCaption( overlay, false );
-
-				if ( currentKey !== pendingKey ) {
-					clearCaptionReveal( overlay );
-					scheduleDeferredSync( overlay );
-				}
-
+			if ( currentKey === pendingKey && ! forceGeometry ) {
 				return;
 			}
 
+			clearCaptionReveal( overlay );
+		}
+
+		if ( shouldSyncImmediately ) {
 			clearDeferredSync( overlay );
 			clearCaptionReveal( overlay );
 			clearOverlayMetricCache( overlay );
@@ -766,7 +760,9 @@
 		}
 
 		clearOverlayMetricCache( overlay );
+		markOverlaySettled( overlay );
 		syncCaption( overlay, false );
+		markOverlayUnsettled( overlay );
 		scheduleDeferredSync( overlay );
 	}
 
@@ -907,12 +903,12 @@
 		classObserver.observe( overlay, { attributes: true, attributeFilter: [ 'class' ] } );
 
 		window.addEventListener( 'resize', function () {
-			requestSettledSync( overlay );
+			requestSettledSync( overlay, true );
 		} );
 
 		if ( window.visualViewport ) {
 			window.visualViewport.addEventListener( 'resize', function () {
-				requestSettledSync( overlay );
+				requestSettledSync( overlay, true );
 			} );
 		}
 	}
--- a/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
+++ b/wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
@@ -14,7 +14,10 @@
  */
 
 ( function () {
-	var CAPTION_REVEAL_DELAY  = 430;
+	// Delay the first caption reveal until core's lightbox zoom sizing has settled.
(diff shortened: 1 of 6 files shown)


wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
```diff
@@ -140,9 +899,24 @@
 		 * transitions can move which container is on screen).
 		 */
 		var classObserver = new MutationObserver( function () {
-			syncCaption( overlay );
+			if ( isOverlayActive( overlay ) ) {
+				requestSettledSync( overlay );
+				return;
+			}
+
+			requestSettledSync( overlay );
```
bor0: This seems incorrect? Perhaps just

```suggestion
			requestSettledSync( overlay );
```

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
```diff
@@ -363,14 +335,12 @@ protected static function build_image_block( $screenshot, $id, $above_fold = fal
 			)
 		);
 
-		$srcset = self::photon_srcset( $src );
```
bor0: The old code served 300/600/900px srcset candidates via i0.wp.com for the grid thumbnails.

The new code loads full-resolution ps.w.org images directly in the gallery grid.

While I understand you wanted to simplify some things, this simplification could also cause performance issues.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
```diff
@@ -1,23 +1,22 @@
...
+ * grows beyond a short-label treatment.
  *
  * `data-wp-text` / `data-wp-bind` attributes added after the
  * Interactivity runtime parsed the page do not bind, so the caption is
  * driven imperatively via a MutationObserver on every overlay <img>.
  */
 
 ( function () {
+	var CAPTION_REVEAL_DELAY  = 430;
+	var SCREENSHOT_ID_OFFSET = 9000000;
```
bor0: Can we describe what these do?

bor0 (commented):
This tested well!

Tested with some screenshots on https://wordpress.org/plugins/accessibility-checker/#screenshots and https://wordpress.org/plugins/wordpress-seo/#screenshots

Thanks for your work @dan-zakirov! 🙌 

I added some comments, looking forward to your thoughts/suggestions

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
dan-zakirov: Good catch, updated to the single requestSettledSync( overlay ) call

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-screenshots.php
dan-zakirov: Good catch. I split this out of #697; the Photon/performance part is handled separately

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
dan-zakirov: Added short comments for both constants

bor0:
@dan-zakirov this is very close, thanks for the updates!

I noticed one more thing - some screenshots "jump" at the initial time you click them, but not on the second time. Other screenshots don't "jump" at all. I guess it depends on whether the text is too large or resize needs to happen.

https://github.com/user-attachments/assets/a4cd8daa-3ad6-402b-80c4-562477e490ef

dan-zakirov:
> @dan-zakirov this is very close, thanks for the updates!
> 
> I noticed one more thing - some screenshots "jump" at the initial time you click them, but not on the second time. Other screenshots don't "jump" at all. I guess it depends on whether the text is too large or resize needs to happen.
> 
>  Untitled.mp4

@bor0 I fixed the jump on the first open. Geometry is now calculated before the animation, with only the caption reveal delayed. Please take another look when you have a moment 🙂

bor0 (approved):
Tests well, thanks!

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
```diff
@@ -1,23 +1,25 @@
...
  * Interactivity runtime parsed the page do not bind, so the caption is
  * driven imperatively via a MutationObserver on every overlay <img>.
  */
 
 ( function () {
+	// Delay the first caption reveal until core's lightbox zoom sizing has settled.
+	var CAPTION_REVEAL_DELAY = 430;
+
+	// Synthetic Plugin Directory screenshot IDs start here; ordinary images stay unmanaged.
+	var SCREENSHOT_ID_OFFSET = 9000000;
```
thilinah, line 22: SCREENSHOT_ID_OFFSET must stay in lockstep with SYNTHETIC_ID_OFFSET (class-screenshots.php:60). If PHP ever changes the offset, every caption silently disappears (managed-image detection fails closed). At minimum, cross-reference each constant in the other file's comment.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
```diff
@@ -57,29 +59,719 @@
...
+				! currentImageHeight ||
+				! currentLightboxScale ||
+				newContainerHeight === currentContainerHeight
+			) {
+				return;
+			}
+
+			scale = newContainerHeight / currentContainerHeight;
+
+			overlay.style.setProperty(
```
thilinah, line 555: Nit: worth a comment noting these --wp--lightbox-* vars are core Image-block internals (set in setOverlayStyles, [pinned link]) and that core rewrites them on open/nav/resize. A core lightbox refactor would break this silently, so a breadcrumb here will save future debugging time.
https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/image/view.js#L585-L597

thilinah (approved):
Thanks @dan-zakirov. Tested and working as expected. I've added a few non-blocking comments. I'll give you some time to address them before deploying.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
dan-zakirov: Added cross-references in both files so the shared offset stays explicit and in sync. Fixed in afbdb97bf.

wordpress.org/public_html/wp-content/plugins/gallery-lightbox-enhancements/assets/lightbox-captions.js
dan-zakirov: Added a comment noting that these are core Image block internals, when core rewrites them, and linked the relevant Gutenberg source. Fixed in afbdb97bf.

thilinah:
Hi @dan-zakirov, could you rebase this onto current trunk.

dan-zakirov:
@thilinah rebased this onto the latest trunk and rechecked it with https://github.com/WordPress/wordpress.org/pull/693 in place. Captions, galleries without captions, navigation, and the mobile lightbox still work as expected. All checks pass.

PR #698, opened 2026-06-14
Plugin directory: Prevent duplicated plugin submissions

frantorres:
Sometimes plugin authors inadvertently submit the same plugin multiple times; when spoken to, they explain that they thought the process had stalled and had tried to resubmit the form. 

This adds two methods to prevent duplicated plugin submissions.
- Frontend: Prevents click again in the submission button and changes the text to "Uploading…"
- Backend: Creates an "upload lock" for that user and that plugin (0 when new) that gets removed after the upload has been processed (if for some reason the process fails, it is saved as wp_cache_add for 5 minutes so they can try again after waiting a bit)


timse201:
Plugin translations are imported automatically when it lands in trac. The best solution would be to allow imports only for PTEs. This prevents such activity on a permanent basis without disrupting legitimate activities.

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-upload.php
```diff
@@ -97,8 +97,19 @@ public static function display() {
...
+			$lock_key = 'plugin_upload_lock_' . get_current_user_id() . '_' . $for_plugin;
+			if ( false === wp_cache_add( $lock_key, time(), 'wporg-plugins', 5 * MINUTE_IN_SECONDS ) ) {
+				$upload_result = new \WP_Error(
+					'upload_in_progress',
+					__( 'Your previous upload is still being processed. Please wait a moment before trying again.', 'wporg-plugins' )
+				);
+			} else {
+				$upload_result = $uploader->process_upload( $for_plugin );
+				wp_cache_delete( $lock_key, 'wporg-plugins' );
+			}
```
obenland, line 112: Is this strictly needed? Maybe we could try disabling the button first and see how it goes?

wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/js/upload.js
```diff
@@ -49,4 +49,19 @@
...
+	$( 'form.plugin-upload-form' ).on( 'submit', function() {
+		var $button = $(this).find( 'input[type="submit"]' );
+
+		$button.prop( 'disabled', true );
+
+		if ( ! $button.data( 'defaultValue' ) ) {
+			$button.data( 'defaultValue', $button.val() );
+		}
+
+		if ( 'undefined' !== typeof pluginUpload && pluginUpload.l10n && pluginUpload.l10n.uploading ) {
```
obenland: I've not tried it, but I wonder if we could pass this string as a data attribute on the button instead?

wordpress.org/public_html/wp-content/plugins/plugin-directory/shortcodes/class-upload.php
frantorres: For the feedback given I suspect that they think the process is stuck, they open a new window or reload the page and submit it again (while it is still processing), so yes, this would be needed. 

Disabling the button is an additional just-in-case, but looks that the lock is the main thing.

obenland: It wouldn't fail with a duplicate slug?

frantorres: It doesn't, when the duplicated submissions happen the system adds a number to the slug. 
For example for a plugin under the name "PRT SEO" we receive prt-seo, prt-seo-2, prt-seo-3 and so on.

wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/js/upload.js
frantorres: True, much cleaner that way, done! Thanks!

PR #706, opened 2026-07-08
Add dynamic social share images for plugin pages (Meta #5926)

mathetos:
Fixes https://meta.trac.wordpress.org/ticket/5926

Plugin pages now get a proper preview image when someone shares a link on social media. You get the title, description, icon, and a few key stats in one clean 1200x630 image.

This follows what Dion asked for on #5926: build it in PHP, keep it simple, and serve dynamic images the same way we already do with geopattern icons so the CDN can cache them. Jetpack stays installed, and on plugin pages we let it step aside for Open Graph and Twitter tags so social platforms use our image.

The layout uses a small set of constants and a straightforward stats grid that holds up across different plugin titles and descriptions. The change stays focused on the plugin directory and theme meta tags, and it passes PHPUnit and PHPCS.

## Sample output
share-image-sample-gutenberg

## Test plan

- [x] PHPCS passes on new share image files
- [x] Plugin Directory PHPUnit suite passes (127 tests)
- [ ] Visit `/share-image/{slug}.jpg` on a plugin page and confirm JPEG renders
- [ ] View page source on a plugin page and confirm `og:image` points at the share image URL
- [ ] Confirm Jetpack does not output duplicate `og:image` tags on plugin pages


environments/plugin-directory/.wp-env.json
```diff
@@ -15,14 +15,15 @@
 		"wp-content/mu-plugins": "./mocks/mu-plugins",
 		"wp-content/mu-plugins/pub": "../wordpress.org/public_html/wp-content/mu-plugins/pub",
 		"wp-content/mu-plugins/wporg-mu-plugins": "WordPress/wporg-mu-plugins#build",
-		"wp-content/env-bin": "./plugin-directory/bin"
+		"wp-content/env-bin": "./plugin-directory/bin",
+		"style": "../wordpress.org/public_html/style"
 	},
 	"lifecycleScripts": {
-		"afterStart": "bash plugin-directory/bin/after-start.sh"
+		"afterStart": "node plugin-directory/bin/after-start.js"
```
obenland, line 22: What makes this change necessary?

wordpress.org/public_html/wp-content/plugins/plugin-directory/docs/share-image-sample-gutenberg.png
obenland, line 1: This should be removed.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-share-image.php
```diff
@@ -0,0 +1,966 @@
+<?php
+namespace WordPressdotorg\Plugin_Directory;
+
+require_once __DIR__ . '/class-plugin-share-image-layout.php';
```
obenland, line 4: This should be covered by the autoloader?

wordpress.org/public_html/wp-content/plugins/plugin-directory/tests/Plugin_Share_Image_Layout_Test.php
```diff
@@ -0,0 +1,70 @@
...
+ * @package WordPressdotorg\Plugin_Directory\Tests
+ */
+
+use PHPUnit\Framework\TestCase;
+use WordPressdotorg\Plugin_Directory\Plugin_Share_Image_Layout;
+
+/**
+ * @group share-image
+ */
+class Plugin_Share_Image_Layout_Test extends TestCase {
```
obenland, line 14: The tests in this class are not helpful.

obenland (commented):
This looks like a good start!

Overall I think it could benefit from a few iterations on simplification, WordPress coding and docs standards, and tests. Removing abstractions, be more specific to what we actually need, have tests be meaningful, etc. Can `Plugin_Share_Image` be half the size?

mathetos:
Working on this based on @obenland 's comments above.

environments/plugin-directory/.wp-env.json
mathetos: Good pushback. It isn't necessary for this PR. I reverted it to trunk's bash plugin-directory/bin/after-start.sh.

The node swap was me working around `afterStart` failing on Windows when WSL has no `/bin/bash`. That's a local env issue, not something this PR actually really needed.

wordpress.org/public_html/wp-content/plugins/plugin-directory/docs/share-image-sample-gutenberg.png
mathetos: Removed.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-share-image.php
mathetos: Also good pushback. I removed that require statement. The layout class it was loading is gone too, so there's nothing left to autoload.

wordpress.org/public_html/wp-content/plugins/plugin-directory/tests/Plugin_Share_Image_Layout_Test.php
mathetos: Ya, cleaned that one up too.

mathetos:
Thanks so much @obenland for the detailed review. The class went from roughly 950 lines (with its layout class) to about 640, so not literally half, but what's left is far more efficient and easier to navigate and troubleshoot.

The biggest wins in this pass was committing to the Inter font that wporg-mu-plugins already ships, instead of probing system fonts. That deleted the bitmap fallback, the fake width math, and all the “font might be missing” branching. The footer is a simple horizontal blend now, and the stat icons use a fixed baseline instead of bounding-box math.
 
On the env file, I kept one thing that really is still needed: style, so the WordPress logotype exists inside wp-env. 

I also mapped wporg-mu-plugins into the test env so PHPUnit can find Inter.ttf and render() can produce a real JPEG. The tests now stay on `PHPUnit\Framework\TestCase` so they run in this suite (PHPUnit 11). They cover install formatting, unpublished and non-plugin posts returning false, and a 1200x630 JPEG for a published plugin, including a multibyte title that has to truncate. That last one used to be able to fatal in GD.

One trade-off worth saying out loud: with a single font dependency, a missing Inter.ttf means no share image at all (the route errors rather than serving a degraded image), and Inter doesn't cover CJK, so non-Latin plugin titles on locale sites will render with gaps. The old DejaVu fallback had the same CJK limitation, but now it's a deliberate choice rather than an accident, so flagging it in case you see it differently.

PHPCS and the Plugin Directory PHPUnit tests all pass now. Happy to keep iterating if find anything else.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-share-image.php
```diff
@@ -0,0 +1,639 @@
...
+	 */
+	public static function get_data( $plugin ) {
+		$plugin = get_post( $plugin );
+
+		if ( ! $plugin || 'plugin' !== $plugin->post_type || 'publish' !== $plugin->post_status ) {
+			return null;
+		}
+
+		$icons = Template::get_plugin_icon( $plugin );
+		$icon  = $icons['icon_2x'] ?? $icons['icon'] ?? '';
```
obenland: `Template::get_plugin_icon()` returns `false` (not `null`) for a missing `icon_2x`, so `??` never falls back — plugins with only a 1x raster icon render with an empty icon box. Use `?:` or explicit falsy checks.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-share-image.php
```diff
@@ -0,0 +1,639 @@
...
+	/**
+	 * Build the public share-image URL for a plugin.
+	 *
+	 * @param int|\WP_Post|null $post Optional. Post ID or post object.
+	 * @return string|false
+	 */
+	public static function get_url( $post = null ) {
+		$plugin = get_post( $post );
+
+		if ( ! $plugin || 'plugin' !== $plugin->post_type ) {
```
obenland: This checks `post_type` but not `post_status`. Closed/disabled are public statuses, so those pages now emit og:image/twitter:image URLs that the route 404s — previously no image tag was emitted for them. Checking for `publish` here would match `get_data()` and the route.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-share-image.php
```diff
@@ -0,0 +1,639 @@
...
+	public static function output( $plugin ) {
+		$bytes = self::render( $plugin );
+
+		if ( false === $bytes ) {
+			return false;
+		}
+
+		status_header( 200 );
+		header( 'Content-Type: image/jpeg' );
+		header( 'Cache-Control: public, max-age=' . YEAR_IN_SECONDS );
```
obenland: A year-long public cache on a mutable image whose URL has no version token pins the first render at CDNs/scrapers with no purge path. The geopattern route this mirrors is a pure function of its URL (slug + color act as the cache buster); this image embeds live title, installs, rating, and locale count. A single 5s icon-fetch timeout also produces an icon-less JPEG that then caches for a year. Consider a version token in the URL (e.g. hash of `last_updated`) or a short TTL.

obenland (commented):
A few ways `class-plugin-share-image.php` could be simplified significantly:

- **Reuse existing stats logic instead of reimplementing it**: `count_locales()` duplicates the Meta widget's `available_languages()` query, `count_contributors()` is a third variant of the widget/API contributor logic, and `format_install_count()` is mostly `Template::format_active_installs_for_display()` plus K+/M+ compaction. With shared helpers this class becomes a pure renderer (~350 lines instead of ~640).
- **Inline the one-shot constants** — roughly 20 of the 24 class constants are used exactly once and nothing is configurable; folding the layout numbers into `render()` cuts a lot of indirection.
- **Gradient footer**: stretching a 2×1 image with `imagecopyresampled()` gets GD to interpolate the gradient in one call, replacing 1,200 `imagecolorallocate()`/`imageline()` iterations.
- **Fake bold**: `draw_stats_row()` duplicates the double-draw trick from `draw_text_block()`; a small shared text helper removes it. Dropping the dashicon glyphs from the stats row would remove another ~45 lines if the labels alone are acceptable visually.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-share-image.php
```diff
@@ -0,0 +1,639 @@
...
+					'network_id' => WPORG_GLOBAL_NETWORK_ID,
+					'public'     => 1,
+					'path'       => '/',
+					'locale__in' => $wp_locales,
+					'number'     => '',
+					'count'      => true,
+				)
+			);
+
+			if ( $count ) {
```
obenland: `if ( $count )` conflates a legitimate zero Rosetta-site count with "no global network" and falls through to the unfiltered language-pack count — e.g. a plugin whose packs are all variant locales would show a Locales stat while the Meta widget (same query) shows none. Return `(int) $count` unconditionally when the constant is defined.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-share-image.php
```diff
@@ -0,0 +1,639 @@
...
+	}
+
+	/**
+	 * Count plugin contributors, including the plugin owner.
+	 *
+	 * @param \WP_Post $plugin Plugin post object.
+	 * @return int
+	 */
+	protected static function count_contributors( $plugin ) {
+		$contributors = get_terms(
```
obenland: This is a third divergent copy of the contributors-plus-owner logic (`widgets/class-contributors.php`, `api/routes/class-plugin.php`). Unlike the widget it doesn't filter out nicenames that no longer resolve to a user, so the count can disagree with the sidebar. Worth extracting a shared helper instead.

wordpress.org/public_html/wp-content/plugins/plugin-directory/class-plugin-directory.php
```diff
@@ -1509,6 +1516,38 @@ function geopattern_icon_route() {
...
+	 * Output a JPEG share image for a given plugin.
+	 */
+	public function share_image_route() {
+		$slug = get_query_var( 'plugin_share_image' );
+
+		if ( ! $slug ) {
+			return;
+		}
+
+		$plugin = get_posts(
```
obenland: Consider `self::get_plugin_post( $slug )` instead — it sanitizes the slug and is backed by the `plugin-slugs` cache group including negative caching, so bot probes of nonexistent slugs don't each run a fresh `WP_Query`. Then: `if ( ! $plugin || 'publish' !== $plugin->post_status ) { 404 }`.

wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/functions.php
```diff
@@ -337,10 +337,13 @@ function social_meta_data() {
 	printf( '<meta name="twitter:card" content="summary_large_image">' . "\n" );
 	printf( '<meta name="twitter:site" content="@WordPress">' . "\n" );
 
-	if ( $banner['banner_2x'] ) {
+	$share_image = Template::get_share_image_url();
```
obenland, line 340: `get_url()` never checks renderability, so if `render()` fails systematically (no GD, or the Inter.ttf build-input path in wporg-mu-plugins goes away), every plugin page points at a 500 and the banner branches below never run — scrapers then get no image at all. The `elseif` chain also drops `og:image` entirely in the `banner_2x` fallback path, where the old code emitted both tags independently.

mathetos:
Thanks for the detailed pass. The class was still a third copy of directory stats, so I stopped patching nits and made it a renderer.

Stats, icons, and install formatting now come from Template. The Contributors widget and the plugins API share the same helper, so the card matches the sidebar. Closed plugins no longer emit a URL the route 404s. If GD or Inter is missing, get_url() returns false and the banner tags run independently again. The route uses get_plugin_post().

The image URL has a version token so a year-long cache can still bust, and a failed icon fetch only caches for five minutes. I dropped the dashicons as you suggested. I think it definitely makes for a way cleaner looking design: 
image

You suggested stretching a tiny 2-pixel strip to get the gradient. GD didn't blend that well at all, it just split navy and blue down the middle, so I paint the gradient across one row of pixels instead. Happy to keep going if you see anything else.

PR #712, opened 2026-07-14
Theme Directory: Restrict search results to published themes

### LINKED ISSUE #69: Theme search is not working
Possibly related to https://meta.trac.wordpress.org/ticket/7277.

On the new theme, [searching for "twenty"](https://wordpress.org/themes/search/twenty/?new-theme=1) returns only 3 items on the first page, but claims to have 445 items, and 38 pages. Page 2 has 9 items, and so on. All pages should have 12 items until the end.

Screenshot 2024-05-14 at 6 29 29 PM

I think the search is using ElasticSearch, which is returning 12 items, but 9 of those are draft themes.

It looks like the same thing is [happening on production](https://wordpress.org/themes/search/twenty/?new-theme=0), but the theme doesn't have normal pagination so it's not a visible error.

Screenshot 2024-05-14 at 6 33 37 PM

obenland:
## Problem

On the Themes Directory search results pages, each page shows far fewer theme cards than the reported total, leaving large empty gaps in the grid.

Reproduced live on `/themes/search/elementor/` — the header claims **1,012 themes** across 49 pages, but the pages render inconsistent, short counts:

| Page | Cards rendered |
|------|----------------|
| 1 | 4 |
| 2 | 12 |
| 3 | 12 |
| 4 | 12 |
| 5 | 6 |

See https://github.com/WordPress/wporg-theme-directory/issues/69.

## Root cause

Theme searches are served by **Jetpack Search (Elasticsearch)** via `Classic_Search::filter__posts_pre_query()`:

1. `found_posts` (which drives the result count and pagination) is set from the **ES total hit count**.
2. The page's post IDs are then loaded for display with `WP_Query([ 'post__in' => $ids, 'perm' => 'readable', ... ])` — with **no `post_status` constraint**.
3. Jetpack's ES query builder (`convert_wp_es_to_es_args`) never carries over the originating query's `post_status`, so the index returns — and counts — themes in non-public statuses (`draft`, `suspend`, `delist`, …).

Those hidden themes inflate the total, but are dropped by `perm => readable` at load time, so each page renders only the published subset of its hits.

## Fix

Add a `post_status: publish` filter to the ES query via the `jetpack_search_es_query_args` filter, so the ES total and the rendered results stay in sync. This mirrors the exact hook and `filter['and']` merge pattern the plugin directory already uses in production (`plugin-directory/class-plugin-search.php`).

## Testing

- New unit test (`Search_Published_Filter_Test`) covering the filter-merge logic across all `es_query_args['filter']` shapes (empty, single filter, existing `and` group) and the non-`repopackage` guard.
- Verified the new code is phpcs-clean and adds no new violations.

Fixes https://github.com/WordPress/wporg-theme-directory/issues/69

https://claude.ai/code/session_01MV5GwLs4jUFbf34tvwt4YT

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/theme-directory/query-modifications.php
+++ b/wordpress.org/public_html/wp-content/plugins/theme-directory/query-modifications.php
@@ -188,6 +188,45 @@ function wporg_themes_pre_get_posts( $query ) {
 add_action( 'pre_get_posts', 'wporg_themes_pre_get_posts' );
 
 /**
+ * Restricts Jetpack Search (Elasticsearch) theme searches to published themes.
+ *
+ * Without this, the ES hit total counts themes in non-public statuses that are
+ * then dropped when the results are loaded for display, leaving search pages
+ * with fewer cards than the total implies.
+ *
+ * @see https://github.com/WordPress/wporg-theme-directory/issues/69
+ *
+ * @param array    $es_query_args The raw Elasticsearch query args.
+ * @param WP_Query $query         The originating WP_Query object.
+ * @return array
+ */
+function wporg_themes_restrict_search_to_published( $es_query_args, $query ) {
+	if ( ! $query instanceof WP_Query || 'repopackage' !== $query->get( 'post_type' ) ) {
+		return $es_query_args;
+	}
+
+	$status_filter = array(
+		'terms' => array(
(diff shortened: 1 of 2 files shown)


Stiofan:
Linked my GitHub now!

PR #714, opened 2026-07-16
Hold changed lines and new files to the full WordPress standard

obenland:
Let's use `WordPress` as our coding standard. With AI writing most of our code now, it's trivial to have it conform new and updated code to the full standard, and everyone reading the code get to benefit from it.

## What

Adds `phpcs-clean.xml.dist` — the stock **WordPress** coding standard with **no sniff exclusions** — and points the branch linter (`.github/bin/phpcs-branch.php`) at it. Newly written code is now held to the unmodified standard, while the existing tree is only ever checked on the lines a branch actually touches.

## How it works

The linter already scopes itself to changed lines + new files:

- **Modified files** → `phpcs-changed`, reporting only on lines the branch changed.
- **New files** → full `phpcs` on the whole file.

Both paths now pass `--standard=./phpcs-clean.xml.dist`. Existing code is untouched until someone edits it.

## Scope of each ruleset

| File | Standard | Used by |
|------|----------|---------|
| `phpcs.xml.dist` | WordPress + ~50 exclusions | `composer lint` (whole repo) — **unchanged** |
| `phpcs-clean.xml.dist` | Stock `WordPress`, no exclusions | `phpcs-branch.php` (changed lines + new files) |

`phpcs-clean.xml.dist` keeps only the environment config (`minimum_wp_version`, `testVersion`, `PHPCompatibilityWP`) and the shared file/vendor `exclude-pattern`s.

## Note

The stock `WordPress.WP.Capabilities` sniff is now active on changed code and will flag wporg-specific custom capabilities as unknown. When that happens, the cap gets documented in place rather than silenced:

```xml
<rule ref="WordPress.WP.Capabilities">
    <properties>
        <property name="custom_capabilities" type="array">
            <element value="plugin_admin_edit" />
        </property>
    </properties>
</rule>
```

Because this only fires on changed lines and new files, the list grows only with caps that actually touch new code.

https://claude.ai/code/session_0178d9rQN6YbCuz5LetN7Rzi

### DIFF
--- a/.github/bin/phpcs-branch.php
+++ b/.github/bin/phpcs-branch.php
@@ -12,7 +12,7 @@
 namespace WordPressOrg\Bin\PHPCS_Changed;
 
 function run_phpcs( $file, $bin_dir ) {
-	exec( "$bin_dir/phpcs $file -snq", $output, $exec_exit_status );
+	exec( "$bin_dir/phpcs $file --standard=./phpcs-clean.xml.dist -snq", $output, $exec_exit_status );
 	echo implode( "\n", $output );
 	return $exec_exit_status;
 }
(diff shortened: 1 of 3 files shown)


obenland:
Superseded by #771: since CI only enforces the ruleset on changed lines and new files, stripping the exclusions from the single `phpcs.xml.dist` achieves the same goal without a second config that can drift from what CI enforces (see 83744292f). The clean-config copy here also predated the fixes in #771 (stale `*.js[x]?` exclude patterns, unescaped shell arguments in the branch script).

PR #720, opened 2026-07-20
Plugin Directory: Block a release when a security scan reports a high risk score

obenland:
## Summary

Adds the **block** side of the Gandalf scan integration: when a scan reports a high `risk_score` for a plugin version still inside its release cooldown, that version is held out of the update API **indefinitely** — the previously served version keeps being served — until a reviewer force-releases it.

This is the counterpart to #717 (serve a clean scan early); this PR approaches the same callback from the other direction. Gandalf does not send a `risk_score` yet, so the feature is read defensively and stays inert (today's behaviour) until it does.

## How it works

- **Threshold** — `Plugin_Scan_Gandalf::RISK_SCORE_BLOCK_THRESHOLD` (default `8`). A completed scan whose `risk_score` is at or above it blocks; below, or absent, it does not.
- **The hold** — `handle_callback()` records a `release_block` (scan id, score, timestamp) on the release row. A single gate in `API_Update_Updater::update_single_plugin()` honours it, so both the deferred cooldown cron and the backup reconciliation cron keep the new version out of `update_source`.
- **Precedence** — block (risk ≥ threshold) → findings alert (existing) → clean.
- **Only within cooldown** — a verdict for a version that is already live, was superseded by a newer commit mid-scan, or had no delay can't un-ship anything; it is escalated to Slack for manual action instead of held.
- **Unblock** — a reviewer force-release clears the block and serves the version; the override is recorded in the audit log.
- **Reviewer visibility** — the Controls metabox shows a blocked-state notice alongside the force-release control.
- **Notifications** — the findings and block Slack alerts now share one sender, and their user-facing wording says "security scan" rather than naming the scanner.

## Tests

- `Gandalf_Risk_Score_Block_Test` — drives `handle_callback()` directly for the handler's branches (blocks, below/absent/non-numeric score, superseded version, no delay, force-release override).
- `gandalf-risk-score-block-callback-test.php` — the full chain over the REST route (auth, routing, the write to `update_source`, force-release closing the loop, the block outlasting the cooldown), built on a shared `Gandalf_Callback_Test_Case` base.

## Notes for reviewers

- **Draft pending SecOps.** Like #717, this is a policy change gated on the Gandalf scanner itself.
- **Overlap with #717** — three points to reconcile when the branches meet:
  - The **test-environment fix** (first commit) is byte-for-byte #717's `c1edf5b11`; whichever PR merges second should drop it.
  - `API_Update_Updater::get_served_version()` is the same helper #717 extracts.
  - When #717's callback args-schema lands, `risk_score` should join it as an optional `{ "type": "number", "minimum": 0 }`. It's read defensively here because trunk's route validates none of the callback body.

### DIFF
--- a/environments/plugin-directory/.wp-env.test.json
+++ b/environments/plugin-directory/.wp-env.test.json
@@ -4,7 +4,13 @@
 	"plugins": [
 		"../wordpress.org/public_html/wp-content/plugins/plugin-directory"
 	],
+	"mappings": {
+		"wp-content/env-bin": "./plugin-directory/bin"
+	},
 	"lifecycleScripts": {
 		"afterStart": "bash plugin-directory/bin/after-start-test.sh"
+	},
+	"config": {
+		"PLUGINS_TABLE_PREFIX": "wp_"
 	}
 }
--- a/environments/plugin-directory/bin/after-start-test.sh
+++ b/environments/plugin-directory/bin/after-start-test.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 #
 # Runs after wp-env start for the test environment.
-# Installs PHPUnit 11 and Yoast polyfills in the test container.
+# Installs PHPUnit 11 and Yoast polyfills, and creates the stub tables tests read.
 #
 
 CONFIG="--config plugin-directory/.wp-env.test.json"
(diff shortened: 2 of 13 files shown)


obenland:
@Luc45 This draft PR assumes a risk_score being passed, just so that I have something testable. This can be updated to what ever is convenient on the Gandalf side when you get to it.

PR #729, opened 2026-07-22
Plugin Directory: Block a release when a security scan reports a high risk score

obenland:
## Summary

Adds the scan-driven half of the release block: when a Gandalf scan reports a high `risk_score`
for a plugin version still inside its release cooldown, that version is held out of the update
API indefinitely — the previously served version keeps being served — until a reviewer
force-releases it.

This is the second half of #720, which is now split in two. It builds on #728, which adds the
blocking mechanism itself and the reviewer control.

> [!IMPORTANT]
> **Stacked on #728.** GitHub can't base a fork PR on another fork branch, so this PR's diff
> includes #728's commits. **Review only the last commit** — `Block a release when a security
> scan reports a high risk score`. This should merge after #728.

## How it works

- **Threshold** — `Plugin_Scan_Gandalf::RISK_SCORE_BLOCK_THRESHOLD` (default `8`). A completed
  scan at or above it blocks; below, or absent, it does not.
- **Precedence** — block (risk ≥ threshold) → findings alert (existing) → clean.
- **Only within cooldown** — a verdict for a version that is already live, was superseded by a
  newer commit mid-scan, or had no delay can't un-ship anything; it's escalated to Slack for
  manual action instead of held.
- **Attribution** — `force_release()` and the Controls metabox name the risk score when a scan
  set the block, rather than describing it as a reviewer's.
- **Notifications** — the findings and block Slack alerts now share one sender, and their
  user-facing wording says "security scan" rather than naming the scanner.

## What ships inert

**Auto-blocking on a risk score is disabled.** A high score alerts Slack; a human decides. The
call is one commented-out line in `handle_callback()` to restore.

Gandalf doesn't send `risk_score` yet either, so the field is read defensively and the path stays
inert until both change.

## Tests

- `Gandalf_Risk_Score_Block_Test` drives `handle_callback()` directly for the handler's branches.
- `Gandalf_Risk_Score_Block_Callback_Test` covers the full chain over the REST route — auth,
  routing, the write to `update_source`, force-release closing the loop, the block outlasting
  the cooldown.
- Fixtures build on #728's `Release_Block_Test_Case` via `Gandalf_Scan_Test_Case` →
  `Gandalf_Callback_Test_Case`.

190 tests pass locally, **8 skipped**.

> [!NOTE]
> Because auto-blocking is off, the 8 tests that assert a hold are skipped, and several of the
> remaining `..._does_not_block` tests pass vacuously — nothing blocks, so "it didn't block"
> is not evidence the threshold logic works. Real coverage of this path only exists once the
> commented-out call is restored. Flagging it so the green checkmark isn't read as more than
> it is.

## Notes for reviewers

- **Draft pending SecOps.** Like #717, the policy question is upstream of the code.
- When #717's callback args-schema lands, `risk_score` should join it as an optional
  `{ "type": "number", "minimum": 0 }`. It's read defensively here because trunk's route
  validates none of the callback body.

### DIFF
--- a/environments/plugin-directory/.wp-env.test.json
+++ b/environments/plugin-directory/.wp-env.test.json
@@ -4,7 +4,13 @@
 	"plugins": [
 		"../wordpress.org/public_html/wp-content/plugins/plugin-directory"
 	],
+	"mappings": {
+		"wp-content/env-bin": "./plugin-directory/bin"
+	},
 	"lifecycleScripts": {
 		"afterStart": "bash plugin-directory/bin/after-start-test.sh"
+	},
+	"config": {
+		"PLUGINS_TABLE_PREFIX": "wp_"
 	}
 }
--- a/environments/plugin-directory/bin/after-start-test.sh
+++ b/environments/plugin-directory/bin/after-start-test.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 #
 # Runs after wp-env start for the test environment.
-# Installs PHPUnit 11 and Yoast polyfills in the test container.
+# Installs PHPUnit 11 and Yoast polyfills, and creates the stub tables tests read.
 #
 
 CONFIG="--config plugin-directory/.wp-env.test.json"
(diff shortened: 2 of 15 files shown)


obenland:
Closing in favor of #777.

PR #738, opened 2026-07-25
REST API: Validate and sanitize every consumed route parameter

obenland:
An audit of every \`register_rest_route()\` call in the repo found 22 route registrations whose callbacks consume request parameters that were never declared in an \`args\` schema, leaving them without validation or sanitization at the route layer. The wporg-login routes were already fixed in r15003; this PR closes out the rest.

Two validation postures are used deliberately:

- **Coerce, don't reject** on routes that api.wordpress.org proxies raw client input into (\`query-plugins\`, plugin info \`locale\`): custom sanitize callbacks (\`absint\`, \`sanitize_text_field\`, \`wp_parse_list\`, locale character filtering) replace strict schema validation, so today's tolerated garbage (\`per_page=\`, unserialized ints for string params) keeps returning 200 instead of cascading into a \`Query Failed.\` 500 with a negative cache on api.wordpress.org.
- **Validate strictly** everywhere else (internal, capability-gated, or zero-caller routes), where a 400 for malformed input is the desired contract.

## Plugin Directory

- **query-plugins** (public, backs api.wordpress.org): declared all 11 consumed parameters with coercing sanitizers per the above.
- **update-stats**: \`plugins\` is a required object, so a scalar body returns a 400 instead of a PHP warning from \`foreach\` — but entries are unconstrained and the handler now skips malformed ones, so one bad entry cannot reject a whole cron batch.
- **committers/support-reps POST**: \`committer\`/\`support_rep\` are now required strings, matching the validation the DELETE routes already had. Route-placeholder args moved to shared route-level \`args\`.
- **blueprint.json**: declared \`zip_hash\`, \`url_hash\`, \`type\`, and \`lang\` — the hash parameters act as the endpoint's authorization and were previously undeclared.
- **release-confirmation**: \`confirmations_required\` typed as integer; \`rollout_strategy\` constrained to the enum from \`Template::get_rollout_strategies()\` instead of being stored raw.
- **gandalf-scan**: declared the callback body schema. Fields are typed but not required and tolerate nulls, so malformed callbacks still reach the handler, which records them to the \`_gandalf_scan_last_error\` audit trail and clears pending scans — rejecting at the REST layer would silently destroy that trail. The handler now guards its dereferences accordingly.
- Also declared: \`plugin-review\` (\`plugin_id\`, \`token\`), \`plugin\` info \`locale\`, \`upload\` \`ID\`, \`self-toggle-preview\` \`dismiss\`.

## Theme Directory

- **themes/1.x query/info/tags** (public; note api.wordpress.org's theme API calls \`Themes_API\` directly and does not pass through these REST routes): declared every parameter \`Themes_API\` accepts, with multi-type schemas (\`string|array\`) where the API tolerates both. \`wp_version\` is declared only on the features endpoint, the one action that consumes it; \`cache_buster\` declared on commercial-shops.
- **stats endpoints** (public): declared \`startDate\`, which reaches \`strtotime()\` and a prepared SQL query.
- **update-stats**: \`themes\` is a required object with unconstrained entries; the handler skips malformed entries (same batch fail-open behavior as the plugin directory).
- **github auto-review**: declared \`theme_slug\` and \`ticket_id\`. Note: typing \`ticket_id\` as integer changes the XML-RPC wire type from \`<string>\` to \`<int>\` on the Trac call — int is the documented Trac RPC type, but worth a one-ticket smoke test after deploy.
- **preview-blueprint POST**: declared the \`blueprint\` payload as required; also fixed two \`'required' => 'true'\` string-literal typos.
- Added a shared \`sanitize_locale()\` helper in the REST namespace.

## Also Viewing (support forums)

- \`page\` (an unconstrained \`.+\` URL capture) is now a required, route-level arg sanitized with the plugin's own \`sanitize_page_url_for_db()\`; \`isTyping\` is typed as boolean on the POST handler.

## Coding standards

New test files pass the full \`--standard=WordPress\` ruleset with one structural exception: PHPUnit derives test class names from file basenames and class names cannot contain hyphens, so test files can never satisfy the hyphenated-lowercase filename sniff (renaming them makes PHPUnit silently skip every test). \`phpcs.xml.dist\` now carries the same tests-scoped exemption WordPress core uses.

## Tests

One functional test suite per endpoint file (18 test files plus a shared base class per directory, loaded from each suite's bootstrap), testing what the endpoints do: querying finds published plugins and honors the \`slug:\` bypass and pagination, the info endpoints return plugin/theme data, update-stats writes whitelisted meta and skips unknown slugs and malformed entries, the tags endpoint lists and slices tags, slug changes work for plugin authors and reject invalid slugs, blueprints are generated for themes, gandalf callbacks are recorded, and every gated endpoint enforces its bearer token or capability check. Malformed-input behavior is pinned per posture: coercion on the proxied routes, 400s on the strict ones.

Writing these surfaced two dormant bugs, fixed here: the theme info endpoint's 404 branch read a nonexistent \`\$api->error\` property so unknown themes returned 200 (now reads the error from the API response), and the plugin info endpoint emitted PHP warnings when review data is unavailable (now guarded in \`get_plugin_reviews_markup()\`).

Known limitations, deliberately out of scope: the upload endpoint's real inputs (\`\$_FILES['zip_file']\`, \`\$_POST['comment']\`, \`\$_REQUEST['upload_token']\`) are read from superglobals inside \`Upload_Handler\` and are invisible to route schemas — declaring them without changing the read path would be decorative; and api.wordpress.org's 48-hour negative caching of failed internal \`plugin_information\` dispatches (\`class-plugins-info-api.php\`) predates this PR and deserves its own change.

Both suites pass under wp-env with no warnings: plugin-directory 191 tests, theme-directory 40 tests.

https://claude.ai/code/session_01JPDKVeVpBoWoB8euRB5hK2

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -79,6 +79,11 @@
 		<exclude phpcbf-only="true" name="PEAR.Functions.FunctionCallSignature" />
 	</rule>
 
+	<!-- PHPUnit derives test class names from file names, and class names cannot contain hyphens. -->
+	<rule ref="WordPress.Files.FileName.NotHyphenatedLowercase">
+		<exclude-pattern>*/tests/*</exclude-pattern>
+	</rule>
+
 	<rule ref="WordPress-Docs">
 		<!-- Descriptive names make explicit comments unnecessary in many cases. -->
 		<exclude name="Squiz.Commenting.FunctionComment.MissingParamComment" />
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/class-base.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/class-base.php
@@ -50,6 +50,22 @@ class Base {
 	}
 
 	/**
+	 * A sanitization callback for REST API locale parameters.
+	 *
+	 * Restricts the value to the characters present in WordPress locale slugs.
+	 *
+	 * @param string $value The requested locale.
+	 * @return string The sanitized locale.
+	 */
+	public function sanitize_locale_callback( $value ) {
+		if ( ! is_scalar( $value ) ) {
+			return '';
+		}
+
+		return preg_replace( '/[^A-Za-z0-9_-]/', '', (string) $value );
+	}
+
+	/**
 	 * A Permission Check callback which validates the request against the internal api-call token.
 	 *
 	 * @param \WP_REST_Request $request The Rest API Request.
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-gandalf-scan.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-gandalf-scan.php
@@ -30,9 +30,47 @@ class Gandalf_Scan extends Base {
 			[
 				'methods'             => \WP_REST_Server::CREATABLE,
 				'callback'            => [ $this, 'scan_callback' ],
+
+				/*
+				 * Fields are typed but not required, and tolerate nulls, so
+				 * that malformed callbacks reach the callback and are recorded
+				 * on the plugin instead of being rejected at the REST layer.
+				 */
 				'args'                => [
-					'plugin_slug' => [
+					'plugin_slug'     => [
 						'validate_callback' => [ $this, 'validate_plugin_slug_callback' ],
+					],
+					'scan_id'         => [
+						'type' => 'string',
+					],
+					'version'         => [
+						'type' => 'string',
+					],
+					'release_ref'     => [
+						'type' => 'string',
+					],
+					'status'          => [
+						'type' => 'string',
+					],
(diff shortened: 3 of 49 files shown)


obenland:
All new test files now pass the full `--standard=WordPress` ruleset (docblocks, formatting, and structure), with one deliberate exception: the file-name hyphenation sniff. PHPUnit 11 discovers test classes by deriving the class name from the file basename, and PHP class names cannot contain hyphens, so a hyphenated test file can never contain the class PHPUnit looks for — renaming the files to `class-*-test.php` silently dropped every new test from the run. WordPress core carries the same exemption for its test suite, so this PR adds the equivalent scoped rule to `phpcs.xml.dist` (`WordPress.Files.FileName.NotHyphenatedLowercase` excluded for `*/tests/*` only), which also unblocks the failing PHPCS check.

PR #760, opened 2026-08-04
Provide a public new-queue count endpoint

### LINKED ISSUE #758: Provide a low-privilege endpoint for the Plugins Team new-queue count
The Plugins Team Tools plugin currently records the daily number of plugins in the `new` queue.

At present, it queries the Plugins REST API endpoint:

```text
GET https://wordpress.org/plugins/wp-json/wp/v2/plugin?status=new&per_page=1
```

The plugin reads the total from the `X-WP-Total` response header. Because the `new` status requires authentication, the request uses Basic Authentication with a WordPress.org username and an Application Password.

```http
Authorization: Basic base64( WPORG_USERNAME:WPORG_APPLICATION_PASSWORD )
```

This is more privilege than the integration needs: it only requires a single aggregate number, but it must hold a user-scoped credential. Even if the credential is stored outside the database and injected through deployment secrets, its compromise could expose access broader than this read-only use case.

Could Meta provide a narrowly scoped, read-only way to retrieve this metric without a user Application Password?

For example, either:

- A dedicated endpoint returning only the current count of plugins with `status=new`, such as:

  ```text
  GET /plugins/wp-json/wp/v2/stats/new-queue-count
  ```

  ```json
  {
    "queue_count": 123
  }
  ```

- Or an authenticated service-to-service endpoint/token with permission limited to reading this aggregate queue statistic.

The Plugins Team Tools plugin would call this endpoint once per day and store the returned integer as an internal historical statistic. It does not need plugin records, author data, submission content, or any ability to modify data.

This would allow us to remove the Application Password from the integration and reduce the impact of a potential credential leak.

davidperezgar:
## Summary

- Adds a public `GET /plugins/wp-json/plugins/v1/stats` endpoint.
- Returns only the aggregate number of plugins awaiting initial review.
- Removes the need for a user-scoped Application Password to retrieve this metric.
- Adds REST API coverage for the endpoint.

## Testing

- `php -l` on changed PHP files
- PHPCS on new files

Closes #758

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-internal-stats.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/api/routes/class-internal-stats.php
@@ -24,6 +24,30 @@ class Internal_Stats extends Base {
 			'callback'            => array( $this, 'bulk_update_stats' ),
 			'permission_callback' => array( $this, 'permission_check_internal_api_bearer' ),
 		) );
+
+		register_rest_route(
+			'plugins/v1',
+			'/stats',
+			array(
+				'methods'             => \WP_REST_Server::READABLE,
+				'callback'            => array( $this, 'get_stats' ),
+				'permission_callback' => '__return_true',
+			)
+		);
+	}
+
+	/**
+	 * Gets public Plugin Directory statistics.
+	 *
+	 * @param \WP_REST_Request $request The REST API request.
+	 * @return array The public statistics.
+	 */
+	public function get_stats( $request ) {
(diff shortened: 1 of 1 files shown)


obenland:
@davidperezgar Looks like this needs a rebase.

There's no other endpoint this can be folded into?

davidperezgar:
Rebased onto the current trunk. I also folded this into the existing Internal_Stats API: the public endpoint is now GET /plugins/wp-json/plugins/v1/stats and returns only queue_count. The existing POST /update-stats endpoint remains authenticated for writes.

PR #772, opened 2026-08-07
Trac: Refresh WordPress.org header and footer templates

obenland:
Regenerates the Trac header/footer templates via `templates/update-headers.php`, picking up the current markup from WordPress.org:

- Updated style/script tags and layout container class hashes in `wporg-head.html`.
- Footer navigation changes, including replacing the Five for the Future link with the Mercantile (Swag) link.


obenland:
Hold on merging this: the source markup from the global-header-footer endpoint currently contains broken CDN URLs — `rewrite_assets_to_cdn()` in wporg-mu-plugins rewrites `includes_url()` (which has a trailing slash) to `https://s.w.org/wp-includes` (which doesn't), producing `https://s.w.org/wp-includesjs/...` URLs that redirect to the wordpress.org homepage. This affects both mediaelement stylesheets in the head and the emoji fallback script in the footer.

The templates should be regenerated once the upstream fix lands and deploys.

obenland:
Unblocked: WordPress/wporg-mu-plugins#752 is merged and deployed, and the templates have been regenerated — all `wp-includes` CDN URLs now resolve correctly (mediaelement styles, dashicons, and the emoji fallback script). Ready for review.

PR #776, opened 2026-08-10
Add /here access to channel manager for wp-ahmedabad

supernovia:
Adding Chetan so he can ping about events.

Fixes https://meta.trac.wordpress.org/ticket/8398.


iamchetanp:
Meta ticket. https://meta.trac.wordpress.org/ticket/8398

PR #784, opened 2026-08-11
Translate: link Meta project pages to their per-locale contributor pages

wppoland:
Ticket: https://meta.trac.wordpress.org/ticket/8396

### What this fixes

The per-locale contributor pages already exist for Meta projects and already render. `/locale/pl/default/meta/wordcamp/` returns 200 and lists 54 contributors with real per-project totals, plus the Translation Editors block, in the same `locale-project-contributors` markup the plugin pages use.

Nothing links to them. Measured on the rendered HTML:

| Page | links to `/locale/…` |
| --- | --- |
| `/projects/wp-plugins/codepress-admin-columns/` | 410 |
| `/projects/meta/wordcamp/` | 0 |

`inc/class-plugin.php` registers wporg project routes for `wp-plugins` and `wp-themes` only, so those get `templates/projects-wp-plugins.php`, whose locale column links to `gp_url_join( 'locale', $locale, $set_slug, $project->path )`. Meta has no route, so it falls through to GlotPress core's `gp-templates/project.php`, whose locale column links into the editor instead. Core has no filter around that link, and wporg does not override `project.php`.

### Approach

A `Routes\Meta` route plus a `projects-meta.php` template rendering the same translation-set table, with the locale name linking to its contributor page. The per-status counts (translated / fuzzy / untranslated / waiting) keep linking into the editor, so the path a translator actually uses is unchanged.

**Why this does not reuse `WP_Plugins::get_plugin_projects()`**, which was my first instinct: that method builds its table from `WHERE p.parent_project_id = $project->id`, which is right for `wp-plugins/<slug>` because that is a container over `dev` / `stable` / `dev-readme` / `stable-readme`. Meta projects are leaves. `/projects/meta/wordcamp/` has 0 sub-projects and 179 locale rows of its own, 140 with non-zero counts. A copy of that query would have returned an empty table. The data GlotPress core already renders is correct; only the destination of the locale link differs.

The template still renders a sub-projects list, so Meta projects that do have children are unaffected.

### Open question for reviewers

The ticket raises this and I did not want to decide it unilaterally: this makes the locale name link to contributors, matching the wp-plugins precedent. The alternative is to leave the locale name pointing at the editor and add a separate "Contributors" link per row. Happy to switch, it is a two-line change in the template.

There is also a more general option, noted in the ticket: add a hook to GlotPress's `gp-templates/project.php` so any install can append per-locale links, rather than wporg carrying a near-copy of that template. That is the better long-term shape but needs a GlotPress release, so this is the version that can ship now. If you would rather wait for the upstream hook, say so and I will close this and open the GlotPress issue instead.

### Checks

`php -l` clean on all three files. PHPCS against `phpcs.xml.dist` reports 0 issues on both new files; `class-plugin.php` has 15 pre-existing issues, none on the line this adds (line 170).

Not verified in a running environment: I do not have a translate.wordpress.org instance to load the route against, so this is verified by reading the routing, autoloading (`register_class_path` resolves `Routes\Meta` to `inc/routes/class-meta.php`) and the core route it mirrors, not by hitting the page. Worth someone loading `/projects/meta/wordcamp/` on sandbox before this goes near production.

Props motylanogha

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/projects-meta.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-customizations/templates/projects-meta.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * Project page for the Meta project family.
+ *
+ * Mirrors GlotPress's project template, except that the locale name links to
+ * the locale's contributors page rather than into the editor, matching what
+ * the wp-plugins and wp-themes project pages already do.
+ *
+ * @see https://meta.trac.wordpress.org/ticket/8396
+ *
+ * @package wporg-gp-customizations
+ */
+
+gp_title(
+	sprintf(
+		/* translators: %s: Project name. */
+		__( '%s project < GlotPress', 'glotpress' ),
+		esc_html( $project->name )
+	)
+);
+gp_breadcrumb_project( $project );
+gp_enqueue_scripts( array( 'gp-common', 'tablesorter' ) );
+gp_tmpl_header();
+?>
+
+<h2 class="project-name"><?php echo esc_html( $project->name ); ?></h2>
+
+<?php if ( $project->description ) : ?>
+	<div class="project-description"><?php echo $project->description; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Project descriptions allow markup. ?></div>
+<?php endif; ?>
+
+<?php if ( $translation_sets ) : ?>
+<div id="translation-sets">
+	<table class="gp-table translation-sets">
+		<thead>
+			<tr>
+				<th><?php esc_html_e( 'Locale', 'glotpress' ); ?></th>
+				<th class="stats percent"><?php esc_html_e( 'Percent', 'glotpress' ); ?></th>
+				<th class="stats translated"><?php esc_html_e( 'Translated', 'glotpress' ); ?></th>
+				<th class="stats fuzzy"><?php esc_html_e( 'Fuzzy', 'glotpress' ); ?></th>
+				<th class="stats untranslated"><?php esc_html_e( 'Untranslated', 'glotpress' ); ?></th>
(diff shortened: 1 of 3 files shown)


wppoland:
Correction, and it undercuts the premise of this PR.

@ocean90 pointed out on [#8396](https://meta.trac.wordpress.org/ticket/8396#comment:2) that the Meta contributor pages are already linked. I verified both of his examples before replying:

- `/locale/pl/default/meta/` links to `/locale/pl/default/meta/wordcamp/` — 3 occurrences, 43 links to locale Meta projects on that page
- `/stats/` carries 1144 links to `/locale/<locale>/default/meta/…`

So "nothing links to them" was wrong. I measured a single page, `/projects/meta/wordcamp/`, and generalised from it without checking the locale-first navigation at all.

What survives is much smaller: `/projects/wp-plugins/<slug>/` carries 410 `/locale/…` links and `/projects/meta/<slug>/` carries 0, so the contributors view is reachable from the project page for a plugin but not for a Meta project. Whether that inconsistency justifies any code is the maintainers' call, and if the answer is no, this PR should be closed as invalid — which I have offered to do on the ticket.

Leaving it open only so the diff is there if the inconsistency turns out to be worth addressing. If it is, the better shape is probably a hook in GlotPress's `gp-templates/project.php` rather than wporg carrying a near-copy of that template for one project family.

Props motylanogha

wppoland:
Closing: [#8396](https://meta.trac.wordpress.org/ticket/8396#comment:4) was closed as `wontfix`. @ocean90's reasoning is that `/projects/` is not an expected starting point for translation, so the inconsistency this patch fixes is not worth code. That is a reasonable call and I am not going to argue it, particularly as the ticket's original framing was mine and was wrong.

Leaving the branch in place in case the `/projects/` navigation is ever revisited, but this needs no review. Thanks for the quick look.

PR #807, opened 2026-08-13
Plugin Directory: Resolve release holds from the stable tag

obenland:
The release cooldown and #785's release block are resolved by looking up the release *named after the plugin's Version header*: `update_single_plugin()`, `block_release()`, and `force_release()` all call `get_release( $post, $version )`, which matches by tag. An author can commit one line into a held tag — changing `Version: 1.4.4` to `Version: 1.4.5` — and the lookup finds no release for "1.4.5": no block, no cooldown, and the tag's rebuilt package is written to `update_source` immediately under the new version label. The importer's `version_tag_mismatch` check is a warning, not a gate. The same rename can point at another, clean release's version to inherit its (absent) hold. With the release cooldown live in production, this is an active bypass of it today: the rename ships a held tag's rebuilt package immediately instead of waiting out the cooldown. It would equally bypass #777's scan-driven blocks once those land.

Builds on #785; complements #777 and #806.

### What it does

- `API_Update_Updater::get_current_release()` resolves the release being served or held from the **stable tag** — the source of the served package, so the hold follows the content, not the header's label — falling back to the version for trunk releases, which are keyed `trunk@{version}`. A rename inside a held tag now resolves to that tag's release, block and cooldown intact.
- `update_single_plugin()`, `block_release()`, and `force_release()` use it, so enforcement, scan-driven blocking, and the reviewer's release all target the same release row. Force-release keeps working on a renamed release rather than failing to find it.
- The two cooldown displays (wp-admin publish metabox, Release Management notice) use it too, so the UI shows the hold that is actually enforced.
- For the normal case — header matching its tag, or a trunk release — both lookups land on the same row and nothing changes. Plugins whose stable tag legitimately differs from their version (e.g. tag `1.4` carrying version `1.4.0`) were previously invisible to the cooldown entirely; they now resolve correctly.

### Testing

`tests/Current_Release_Resolution_Test.php` (6 tests): a renamed header keeps the cooldown hold and keeps a block's hold, a rename pointing at another clean release's version still holds, `block_release()` lands the block on the stable tag's release, force-release serves the renamed version, and trunk releases still resolve by version. Verified the first three fail against the previous resolution.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/admin/metabox/class-controls.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/admin/metabox/class-controls.php
@@ -60,7 +60,7 @@ class Controls {
 			return;
 		}
 
-		$release = Plugin_Directory::get_release( $post, $version );
+		$release = API_Update_Updater::get_current_release( $post );
 		if ( ! $release ) {
 			return;
 		}
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/admin/metabox/class-controls.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/admin/metabox/class-controls.php
@@ -55,17 +55,12 @@ class Controls {
 	protected static function display_release_cooldown() {
 		$post = get_post();
 
-		$version = get_post_meta( $post->ID, 'version', true );
-		if ( ! $version ) {
-			return;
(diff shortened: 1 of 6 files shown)


obenland:
@Luc45 This moves the version lookup from `Stable Tag` in plugin readme to the tag folder in SVN.

PR #825, opened 2026-08-21
Add Waveform player to the external libs

scruffian:
I'm not sure if this is the best way to add this, or if there's a process that generates it.


sirreal (approved):
I'm not familiar with maintaining this myself, but I had copilot research and it says this is the right thing to do.

Looks good to me.

SergeyBiryukov:
Thanks for the PR! Merged in [r15091](https://meta.trac.wordpress.org/changeset/15091).

PR #831, opened 2026-08-25
SSO: scope remote-login tokens to the destination's registrable domain

bor0:
### Summary

Scopes the SSO remote-login/logout token to the destination's **registrable domain** (via the existing `_get_targetted_host()`), rather than leaving it host-agnostic. A token stays valid across same-family canonical redirects (e.g. `2023.us.wordcamp.org` → `us.wordcamp.org`) but is not accepted on an unrelated family (`wordcamp.org` ≠ `wordpress.org`).

Also adds **temporary, deduped** logging (behind the `wporg_sso_log_binding_rejects` filter) so we can confirm during rollout that the scoping doesn't affect legitimate sign-ins — it logs nothing under normal traffic and can be removed once things look clean.

### Testing

Validated on a sandbox: legitimate login **and** logout across wordcamp / bbpress / buddypress, plus the `2023.us → us` canonical redirect, all work; tokens don't validate across unrelated families.

### Rollout

Draft — please hold until deployed. In-flight tokens (≤5 min TTL) simply re-issue after deploy.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved single sign-on security by binding login and logout tokens to the intended website domain.
  * Added consistent handling for domain capitalization, ports, and related host variations.
  * Added temporary diagnostic warnings when valid tokens are rejected because they belong to a different domain.
  * Improved privacy and reliability when recording diagnostic request details by limiting logged information to the relevant URL path.

### DIFF
--- a/common/includes/wporg-sso/wp-plugin.php
+++ b/common/includes/wporg-sso/wp-plugin.php
@@ -577,7 +577,8 @@ if ( class_exists( 'WPOrg_SSO' ) && ! class_exists( 'WP_WPOrg_SSO' ) ) {
 				array(
 					'action'         => 'remote-logout',
 					'redirect_to'    => urlencode( $logout_redirect ),
-					'sso_logout'     => urlencode( $this->_generate_remote_token( $user ) )
+					// The logout token is consumed on the SSO host (see _maybe_perform_remote_logout()), so bind it there.
+					'sso_logout'     => urlencode( $this->_generate_remote_token( $user, $this->sso_host ) )
 				),
 				$this->sso_host_url . '/wp-login.php'
 			);
@@ -685,7 +686,7 @@ if ( class_exists( 'WPOrg_SSO' ) && ! class_exists( 'WP_WPOrg_SSO' ) ) {
 			$redirect_host = parse_url( $redirect, PHP_URL_HOST );
 
 			if ( $user && $this->_is_valid_targeted_domain( $redirect_host ) && ! preg_match( '!wordpress.org$!i', $redirect_host ) ) {
-				$sso_token = $this->_generate_remote_token( $user );
+				$sso_token = $this->_generate_remote_token( $user, $redirect_host );
 				$redirect  = add_query_arg( 'sso_token', urlencode( $sso_token ), $redirect );
 			}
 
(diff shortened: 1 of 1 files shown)


common/includes/wporg-sso/wp-plugin.php
```diff
@@ -757,20 +779,30 @@ protected function _generate_remote_token( $user ) {
...
+		 * @param string $target_host The host the token is issued for. Binding the hash
+		 *                            to the host keeps a token issued for one host from
+		 *                            being accepted on another.
 		 */
-		protected function _generate_remote_token_hash( $user, $valid_until, $remember_me = false, $session_token = '' ) {
+		protected function _generate_remote_token_hash( $user, $valid_until, $remember_me = false, $session_token = '', $target_host = '' ) {
+			// Scope the token to the destination's registrable domain (wordcamp.org,
+			// bbpress.org, ...) rather than the exact host, so it stays valid across
+			// same-family canonical redirects (e.g. 2023.us.wordcamp.org -> us.wordcamp.org)
+			// but not on an unrelated family (wordcamp.org != wordpress.org).
```
obenland: Multiline comments should use php docs standards

https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-2-multi-line-comments

common/includes/wporg-sso/wp-plugin.php
```diff
@@ -798,14 +830,42 @@ protected function _validate_remote_token( $sso_token ) {
...
 
+			/*
+			 * Temporary rollout check: log tokens that are well-formed and unexpired
+			 * but don't validate for the current host's family, so we can confirm the
+			 * scoping above isn't affecting legitimate sign-ins. Deduped per host
+			 * (10 min) so it can't be flooded, and behind a filter so it's easy to turn
+			 * off. Under normal traffic this logs nothing. Remove after rollout.
+			 */
+			if ( ! $valid && $expiration_valid && $user && ! $valid_hash
+				&& apply_filters( 'wporg_sso_log_binding_rejects', true )
```
obenland: There's no phpcs rule for this, but filters and actions should have docs, even if it's just pointing to the file where the hook is first defined.

Maybe break this out in its own line and save the result in a variable to give it proper docs

obenland (commented):
Both optional

common/includes/wporg-sso/wp-plugin.php
bor0: Done — switched it to a `/* */` block comment.

common/includes/wporg-sso/wp-plugin.php
bor0: Good call — broke the filter out into a documented `$log_rejects` variable with a docblock pointing at where it fires.

PR #852, opened 2026-09-01
Openverse: Strip only the leading subpath when building the redirect target.

mcliwanow:
`wordpress.org/openverse/` redirects to the standalone Openverse site. `get_target_url()` builds the destination by removing the `/openverse` prefix from the request path and appending what is left to the configured origin.

The removal is written as:

```php
$count = 1; // Only replace the leading Openverse subpath.
$target_url .= str_replace( OPENVERSE_SUBPATH, '', $path, $count );
```

`str_replace()` has no limit parameter. Its fourth argument is a by-reference output that receives the number of replacements performed, so the `1` is overwritten and every occurrence of `/openverse` is removed rather than only the leading one.

Any request whose path or query contains the string a second time is forwarded to the wrong URL, including paths where it sits inside a longer segment:

| Request | Forwarded to today | Should be |
| --- | --- | --- |
| `/openverse/image/openverse-logo/` | `{origin}/image-logo/` | `{origin}/image/openverse-logo/` |
| `/openverse/tag/openverse/` | `{origin}/tag/` | `{origin}/tag/openverse/` |
| `/openverse/search/?q=/openverse` | `{origin}/search/?q=` | `{origin}/search/?q=/openverse` |

The remainder is appended to a bare origin, which the setting's `sanitize_callback` guarantees has no trailing slash, so the remainder has to supply the separator itself. It was appended unchanged, which produces a malformed URL whenever it does not already start with a slash.

Two smaller things in the same path. The setting's `sanitize_callback` is the only place the origin's trailing slash is dropped, and `wp theme mod set` bypasses it, so a value set that way doubles the slash in every forwarded URL. And the same target is echoed into an HTML comment unescaped when the redirect is switched off (`index.php`), a branch that renders for every request under the site because `use_index_php_as_template()` forces a 200, and which is the default since `ov_is_redirect_enabled` starts out false.

## What changed

* Remove the subpath only when it is a leading, whole path segment, so `/openverse-search` keeps its first segment instead of becoming `/-search`.
* Prefix what is left with a single slash, so it always contributes a path.
* Read `REQUEST_URI` through `wp_unslash()` and `esc_url_raw()`. `sanitize_text_field()` is not usable here: it strips percent-encoded octets and would turn `?q=cat%20dog` into `?q=catdog`.
* Drop the origin's trailing slash in `get_target_url()` rather than only in the Customizer, so a value set through `wp theme mod set` behaves the same.
* Read the on/off setting through `wp_validate_boolean()`. `wp theme mod set` stores the string it is given, so `ov_is_redirect_enabled false` stored `'false'`, which PHP reads as true and switched the redirect on.
* Check the target's host against the configured origin before redirecting, and render the page when it does not match. The path is appended straight after the origin, so the authority always comes from the setting and never from the request; comparing them makes that an enforced property rather than an assumption.
* Escape the target URL echoed by the disabled-redirect branch of `index.php`.

### 1. The suite

```bash
cd environments && npm run openverse:test
```

67 cases covering what each request maps to, that a forwarded URL always starts with the configured origin, the locale prefix, the origin's own variants, and the string forms the on/off setting can hold. This is the theme's first suite, so it adds `environments/openverse/` and an entry in `.github/unit-tests-suites.yml`.

### 2. The root cause, no WordPress needed

```bash
php -r '$c = 1; var_dump( str_replace( "/openverse", "", "/openverse/image/openverse-logo/", $c ), $c );'
```

```
string(12) "/image-logo/"
int(2)
```

The path lost `/openverse` twice, and `$c` came back as `2`. The `1` was an output slot, not a limit.

### 3. On a site running the theme

Set `ov_redirect_url` to `https://openverse.org` and tick `ov_is_redirect_enabled` in the Customizer, then read the `Location` header:

```bash
curl -sI --path-as-is "https://<site>/openverse/image/openverse-logo/" | grep -i '^location'
```

These three are wrong before the patch and right after it:

| Request path | Before | After |
| --- | --- | --- |
| `/openverse/image/openverse-logo/` | `https://openverse.org/image-logo/` | `https://openverse.org/image/openverse-logo/` |
| `/openverse/tag/openverse/` | `https://openverse.org/tag/` | `https://openverse.org/tag/openverse/` |
| `/openverse/search/?q=/openverse` | `https://openverse.org/search/?q=` | `https://openverse.org/search/?q=/openverse` |
| `/openverse-search` | `https://openverse.org/-search` | `https://openverse.org/openverse-search` |

These five must be identical before and after, and are the check that nothing legitimate broke:

| Request path | `Location`, unchanged |
| --- | --- |
| `/openverse/` | `https://openverse.org/` |
| `/openverse/search/?q=dog` | `https://openverse.org/search/?q=dog` |
| `/openverse/search/?q=cat%20dog` | `https://openverse.org/search/?q=cat%20dog` |
| `/openverse/search/?q=caf%C3%A9` | `https://openverse.org/search/?q=caf%C3%A9` |
| `/openverse/about/` | `https://openverse.org/about/` |

The last two matter because they are what rules out `sanitize_text_field()`.

Two deliberate changes to expect:

* `/openverse` with no trailing slash now resolves to `https://openverse.org/` instead of `https://openverse.org`. That is what the docblock and `README.md` already documented.
* `esc_url_raw()` percent-encodes square brackets, so `/openverse/search/?q=a[]=1&a[]=2` forwards as `?q=a%5B%5D=1&a%5B%5D=2`. The two are equivalent to any server, and Openverse's own parameters (`q`, `license`, `extension`, `source`) do not use bracket syntax. Characters outside `esc_url()`'s allowed set are dropped, but `wp_sanitize_redirect()` already dropped the same ones inside `wp_redirect()`, so that part is not new.
* `esc_url()` also runs `str_replace( ';//', '://' )`, so `/openverse/search/?q=http;//example.com` forwards as `?q=http://example.com`. Neither the raw `REQUEST_URI` nor `wp_sanitize_redirect()` made that substitution, so unlike the dropped characters this one is new. The host stays on the configured origin either way.

### 4. Lint

```bash
BASE_REF=trunk php .github/bin/phpcs-branch.php
```

Clean on the changed lines.

<details>
<summary>Local multisite, if you want to run step 3 end to end</summary>

wp-env has no multisite option, so it is a convert after start. Node 22, because wp-env exits 0 with no output on Node 24.

Save as `.wp-env.json` anywhere and point `wp-env` at it. Paths below are absolute because the file lived outside the repo; the configs under `environments/` use paths relative to their own directory instead.

```json
{
	"core": "WordPress/WordPress#master",
	"phpVersion": "8.4",
	"testsEnvironment": false,
	"port": 8891,
	"themes": [
		"/abs/path/to/wordpress.org/public_html/wp-content/themes/pub/wporg",
		"/abs/path/to/wordpress.org/public_html/wp-content/themes/pub/wporg-openverse"
	],
	"mappings": {
		"wp-content/mu-plugins": "/abs/path/to/environments/mocks/mu-plugins",
		"wp-content/mu-plugins/pub": "/abs/path/to/wordpress.org/public_html/wp-content/mu-plugins/pub",
		"wp-content/mu-plugins/wporg-mu-plugins": "WordPress/wporg-mu-plugins#build"
	},
	"config": { "WP_DEBUG": true, "WP_DEBUG_DISPLAY": false }
}
```

```bash
nvm use 22
npx wp-env --config /path/to/.wp-env.json start

# everything below runs through: npx wp-env --config /path/to/.wp-env.json run cli wp ...
wp core multisite-convert --title="WordPress.org"
wp site create --slug=openverse --title="Openverse"
wp theme enable wporg-openverse --network
wp theme activate wporg-openverse --url=http://localhost:8891/openverse/
wp theme mod set ov_redirect_url https://openverse.org --url=http://localhost:8891/openverse/
wp theme mod set ov_is_redirect_enabled 1 --url=http://localhost:8891/openverse/
```

`multisite-convert` says "Don't forget to set up rewrite rules" and means it. Replace the block in `~/.wp-env/<hash>/WordPress/.htaccess` with the standard subdirectory-multisite rules from the [Codex](https://wordpress.org/documentation/article/htaccess/#multisite), or deep paths under `/openverse/` never reach PHP.

One trap when comparing before and after: the container runs opcache with `revalidate_freq=2`, so swapping the theme files and immediately curling serves the old bytecode and both runs look identical. Run `docker restart <hash>-wordpress-1` after each swap.

</details>

## Notes for the reviewer

`js/iframe_nav.js` does its own subpath strip with `.replace( openverseSubpath, '' )`. JavaScript's string `replace()` already stops at the first occurrence, so the two sides now agree. Worth leaving alone rather than aligning it with the old PHP behaviour. (It sits under the `TODO: Delete this` block, so it may go away anyway.)

The suite extends the plain PHPUnit `TestCase` rather than `WP_UnitTestCase`, which does not work with the PHPUnit 11 runner these environments install. Nothing in it touches the database. `tests/locales-stub.php` stands in for the wporg locales mu-plugin, which the test environment does not install.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redirects from the Openverse WordPress site, including path, query value, and trailing-slash handling.
  * Prevented partial path matches from being treated as valid Openverse subpaths.
  * Added stricter validation for redirect destinations, requiring a non-empty host and explicit HTTP or HTTPS scheme.
  * Safely handled invalid or incomplete redirect URLs.
  * Preserved host allowlisting and URL sanitization, including conditional support for standalone redirect hosts.

* **Tests**
  * Added automated coverage for URL forwarding, locale handling, host validation, and redirect security.


wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/index.php
```diff
@@ -24,10 +24,10 @@
 $target_url = get_target_url();
 
 if ( $is_redirect_enabled ) {
-	wp_redirect( $target_url, 301 );
+	wp_safe_redirect( $target_url, 301 );
```
bor0: `wp_safe_redirect()` does not skip the redirect when validation fails, it falls back to `admin_url()`. Combined with the `301` here, any state where `allow_standalone_redirect_host()` produces a host that does not match the target's host means every `/openverse/*` request issues a permanent, browser-cached redirect into `/openverse/wp-admin/`. Recovering from that needs users to clear their cache, not just a revert.

I tried to construct a reachable trigger and mostly could not, since the filter derives its host from the same theme mod that builds the target. The one path I found is a non-http/https scheme in `ov_redirect_url`: the setting's `sanitize_callback` only strips a trailing slash, it does not validate the scheme, and `wp_validate_redirect()` rejects on scheme before the host check ever runs. Low probability, high blast radius.

A `wp_safe_redirect_fallback` filter scoped to this branch, returning the origin or `home_url()`, would cap the damage cheaply.

Worth noting for framing: the `'/' . ltrim( $path, '/' )` change in `functions.php` is what actually closes the host-escape surface, since the remainder can no longer start with `//`. `wp_safe_redirect()` is defense in depth on top of an already-closed hole. That is a defensible choice, just not a load-bearing one, so the fallback tradeoff is the thing to weigh.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -265,15 +270,50 @@ function get_target_url() {
...
+	}
+
+	$host = wp_parse_url( get_theme_mod( 'ov_redirect_url', OPENVERSE_STANDALONE_URL ), PHP_URL_HOST );
+	if ( $host ) {
+		$hosts[] = $host;
+	}
+
+	return $hosts;
+}
+add_filter( 'allowed_redirect_hosts', __NAMESPACE__ . '\allow_standalone_redirect_host' );
```
bor0: `allowed_redirect_hosts` is global, so this also whitelists the Openverse host for every other `wp_safe_redirect()` on the blog, including the login `redirect_to`. Minimal impact on a dedicated `/openverse` site, but it is a redirect-policy widening rather than a purely local change, so worth a line in the description.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -265,15 +270,50 @@ function get_target_url() {
...
 	}
 
-	$path = $_SERVER['REQUEST_URI'];
-	if ( $path ) {
-		$count = 1; // Only replace the leading Openverse subpath.
-		$target_url .= str_replace( OPENVERSE_SUBPATH, '', $path, $count );
+	// Sanitising is required by WordPress.Security.ValidatedSanitizedInput.
+	// Not `sanitize_text_field()`, which strips percent-encoded octets and
+	// would turn `?q=cat%20dog` into `?q=catdog`.
+	$path = isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) ) : '';
```
bor0, line 276: The description carefully rules out `sanitize_text_field()`, but `esc_url_raw` has behavior changes of its own that are not listed. Verified against core:

| Request | Forwarded |
| --- | --- |
| `/openverse/search/?q=a[]=1&a[]=2` | `?q=a%5B%5D=1&a%5B%5D=2` |
| `/openverse/search/?q=a"b` | `?q=ab` |

Brackets get percent-encoded, characters outside `esc_url`'s allowed class (`"`, `<`, `>`, `{`, `}`, `\`, backtick) are dropped, and `;//` is rewritten to `://`.

In practice these are all characters a browser would have percent-encoded anyway, and Openverse's params (`q`, `license`, `extension`, `source`) do not use bracket syntax, so I read this as acceptable. It belongs in the test table though, so the next person does not rediscover it as a bug.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -265,15 +270,50 @@ function get_target_url() {
...
+
+	// Only the leading subpath is removed. `str_replace()` cannot express that:
+	// it replaced every occurrence, including one in a later segment such as
+	// `/image/openverse-logo/` or one in the query string.
+	if ( str_starts_with( $path, OPENVERSE_SUBPATH ) ) {
```
bor0: Nit: the strip is prefix-based rather than segment-boundary-based. Verified: `/openverse-foo/bar/` becomes `{origin}/-foo/bar/`, and `/openverseXYZ` becomes `{origin}/XYZ`.

Pre-existing, and unreachable on the real mount since the blog lives at `/openverse/`. But given the PR is precisely about the strip being too eager, tightening it costs one line:

```suggestion
	if ( OPENVERSE_SUBPATH === $path || str_starts_with( $path, OPENVERSE_SUBPATH . '/' ) || str_starts_with( $path, OPENVERSE_SUBPATH . '?' ) ) {
```

bor0 (commented):
Reviewed the branch locally and verified the claims against real WordPress core (`esc_url` / `esc_url_raw` / `wp_validate_redirect` loaded standalone from a WP 7.0.2 checkout) rather than from memory.

**The core fix is correct and the diagnosis is accurate.** `str_replace()`'s fourth parameter is a by-reference count output, not a limit, so `$count = 1` never constrained anything. All eight rows in the description reproduce exactly, including the five that must stay unchanged.

Three supporting claims also check out:

* The trailing-slash change is not silent. `README.md` L124-133 already documents `https://ru.wordpress.org/openverse` -> `https://openverse.wordpress.net/ru/`, so this aligns the code with the docs.
* `wp_unslash()` is genuinely required, not ceremony: core's `wp_magic_quotes()` slashes `$_SERVER` alongside `$_GET` / `$_POST`.
* The `iframe_nav.js` note is right. JS `String.prototype.replace()` with a string pattern replaces the first occurrence only.

`esc_html()` in `index.php` is also a real fix rather than cosmetic: it turns `-->` into `--&gt;`, closing HTML-comment breakout.

Four comments inline. Only the first is one I would want resolved before commit; the rest are description and follow-up material.

**On tests:** the deferral is reasonable given the theme is not in `.github/unit-tests-suites.yml`, but I would push for the follow-up. This is the second time the strip logic has been wrong, and `get_target_url()` is now a clean function of `REQUEST_URI` plus two theme mods, so the table in the description is nearly a data provider already.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/index.php
mcliwanow: Reproduced: with `ov_redirect_url` set to `ftp://openverse.org`, every `/openverse/*` 301'd to
`/openverse/wp-admin/`. Scheme is rejected before the host check, so the allow-list never runs.

Fixed in 6a4c65b4a with a guard rather than `wp_safe_redirect_fallback`:

```php
if ( $is_redirect_enabled && wp_validate_redirect( $target_url, false ) ) {
```

A fallback still emits a 301. Falling through renders the page, same as when the redirect is off, so
nothing new gets cached. Happy to switch if you prefer the filter.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
mcliwanow: Fair, added to the description. It returns early when `ov_is_redirect_enabled` is off, so the
widening only applies while the redirect is on.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
mcliwanow: Added the bracket row.

Second row is not a change though: `?q=a"b` forwards as `?q=ab` before this PR too, since
`wp_sanitize_redirect()` already drops `"`. Brackets are the only new difference, it keeps `[` and
`]` where `esc_url()` encodes them.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
mcliwanow: Taken in 6a4c65b4a as written.

The `?` clause matters: with only `. '/'`, `/openverse?q=dog` stops matching and forwards as
`{origin}/openverse?q=dog`. Now `/openverse-search` goes through intact and `/openverse?q=dog` still
gives `{origin}/?q=dog`.

mcliwanow:
All four addressed, two in 6a4c65b4a and two in the description.

Tests: agreed, I'll open the follow-up. Needs the first theme entry in
`.github/unit-tests-suites.yml`, which is why it is not here.

On "second time": `git log -S OPENVERSE_SUBPATH` returns three commits. 0c2992b5d added the constant
for JS only, 3e9c629ac added the PHP strip with the bug, this is the first fix. Written once, wrong
once.

bor0 (commented):
Re-reviewed 6a4c65b4a against core.

The segment match is exactly right, and I re-ran the nine documented rows against the new implementation: all still match, so the tightening did not cost anything.

The `wp_validate_redirect()` guard is a better answer than the `wp_safe_redirect_fallback` filter I suggested. Falling through to the existing `else` branch means a bad setting renders the embed rather than a broken page, and I checked that this actually holds: `enqueue_assets()` is hooked unconditionally on `wp_enqueue_scripts` with no `ov_is_redirect_enabled` gate, so the fallback page is a working iframe, not an empty one.

The description additions read accurately, including the `wp_sanitize_redirect()` point. Every character `esc_url()` drops is also outside `wp_sanitize_redirect()`'s allowed class, so `esc_url_raw()` is not newly lossy there. (It is stricter in the other direction, dropping `$`, `|` and `'` that `esc_url()` keeps, but that is pre-existing and applies equally before this PR.)

One comment inline: a leftover of the same class as the `ftp://` case. Not a blocker.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/index.php
```diff
@@ -18,16 +18,21 @@
...
+	a 301 into wp-admin would sit in visitors' caches long after the setting
+	that caused it was corrected. Rendering the page is the safer failure.
  */
 
 $is_redirect_enabled = get_theme_mod( 'ov_is_redirect_enabled' );
 $target_url = get_target_url();
 
-if ( $is_redirect_enabled ) {
-	wp_redirect( $target_url, 301 );
+if ( $is_redirect_enabled && wp_validate_redirect( $target_url, false ) ) {
```
bor0: The `ftp://` door is closed, confirmed. One next to it is still open, same class and same reachability.

`wp_validate_redirect()` is not a verdict, it is a repair function: given a target with no host it rewrites the location rather than rejecting it. So a target it had to repair still returns truthy, and still passes this guard.

With `ov_redirect_url` set to `openverse.org`, no scheme, which the setting's `sanitize_callback` permits since it only strips a trailing slash:

```
target:   openverse.org/search/?q=dog
guard:    passes
Location: /openverse/search/openverse.org/search/?q=dog   (301)
```

Verified against core: `parse_url()` finds no host, so the `! isset( $lp['host'] ) && '/' !== $lp['path'][0]` branch prepends `dirname()` of the current `REQUEST_URI`, and the host allow-list check is then skipped entirely because `isset( $lp['host'] )` is false. That is the same permanent, cache-resident 301 the guard was added to prevent, reached through the relative-URL branch instead of the scheme branch. Reachable the same way you reproduced the last one, via `wp theme mod set`; the Customizer's `type => 'url'` control does not gate that path.

One expression covers it, since this target should always be absolute:

```suggestion
if ( $is_redirect_enabled && wp_parse_url( $target_url, PHP_URL_HOST ) && wp_validate_redirect( $target_url, false ) ) {
```

Tightening `ov_redirect_url`'s `sanitize_callback` to require an absolute `http`/`https` URL would complement it, but on its own it does not help: it only runs on Customizer save, so the `wp theme mod set` path stays open.

Tiny thing while you are in here: `wp_validate_redirect()` documents `$fallback_url` as `string`. `false` works, since the value is only ever returned, but `''` is the documented default and equally falsy.

mcliwanow: Good catch. Reproduced: `openverse.org` with no scheme gives `Location: /openverse/search/openverse.org/search/?q=dog`. Missed that it rewrites rather than bails.

Fixed in b20e54ecb, your expression as-is plus the `''`. Nine rows unchanged, checked on the multisite too.

Left `sanitize_callback` alone, it won't stop `wp theme mod set` anyway. Can do it later if you want.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/index.php
```diff
@@ -18,16 +18,26 @@
...
+	current directory and returns that, skipping the allowed-host check
+	entirely. This target is always absolute, so no host means no redirect.
  */
 
 $is_redirect_enabled = get_theme_mod( 'ov_is_redirect_enabled' );
 $target_url = get_target_url();
 
-if ( $is_redirect_enabled ) {
-	wp_redirect( $target_url, 301 );
+if ( $is_redirect_enabled && wp_parse_url( $target_url, PHP_URL_HOST ) && wp_validate_redirect( $target_url, '' ) ) {
```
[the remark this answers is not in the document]
mcliwanow: Right, and this one is on the branch rather than pre-existing. Before it, `//openverse.org` went out unchanged and the browser resolved it against the page scheme, so it stayed https. `wp_validate_redirect()` rewrites it to `http:`, so switching to `wp_safe_redirect()` turned that into a real downgrade.

Fixed in 61c67edc4. Scheme now has to be `http` or `https`, and the three checks moved into `is_valid_target_url()` since the `if` was getting silly. `//`, scheme-less, `ftp://` and `javascript:` all render the page now, the nine rows are unchanged. Checked on the multisite.

mcliwanow:
@bor0, I added three commits since your last look.

- 61c67edc4: your repair-not-reject point had a sibling. `//openverse.org` has a host, so it cleared b20e54ecb's check and wp_validate_redirect() rewrote it to `http:`. Scheme and host are checked together now, in `is_valid_target_url()`. The scheme comparison is redundant, core compares the same way, so it states the contract rather than doing work.
- cdc87b43e: tests, as promised. 67 cases, first theme entry in unit-tests-suites.yml. Reverting any of the fixes turns it red.
- 5f43f8cf2: an audit of the guard found no way to move the host, but three ways an operator can serve a 301 they did not want. Worst one: `wp theme mod set ov_is_redirect_enabled false` stores the string, which is true to PHP, so it switched the redirect on. Read through `wp_validate_boolean() `now. Also `untrailingslashit()` on the origin, which only the Customizer was stripping.

I'd appreciate another look

bor0 (commented):
Third pass, covering the three commits since my last look (`61c67edc4`, `cdc87b43e`, `5f43f8cf2`). Checked the load-bearing claims against a real WordPress checkout rather than from memory.

**All three hold up.**

* `61c67edc4`: the gap was real. `wp_validate_redirect()` prepends `http:` to a `//host` location before it parses (`pluggable.php:1660`), so `//openverse.org` arrived at the host check with a host and cleared it. Checking scheme and host together in `is_valid_target_url()` closes it, and you are right that this was introduced on the branch rather than pre-existing.
* `5f43f8cf2`: both docblock claims check out. `parse_url()` preserves scheme case (`parse_url( 'HTTPS://openverse.org/search/' )` returns `"HTTPS"`), and core compares `'http' === $lp['scheme']` case-sensitively at `pluggable.php:1686`, so the explicit `in_array()` genuinely states a contract rather than carrying weight. `wp_validate_boolean()` (`functions.php:7739`) returns false only for bool `false`, the string `false` in any case, and falsy scalars, so `'no'` and `'off'` stay truthy exactly as the commit message says.
* `cdc87b43e`: the `explicit port` row is right, which I had to check. Current core's allow-list comparison at `pluggable.php:1727` has no `host:port` clause, so `https://openverse.org:8443` passes with a bare `openverse.org` allow-listed.

`test_a_forwarded_request_is_sent_unchanged` is the best test in the file. Asserting `$target === wp_validate_redirect( $target, 'FELL-BACK' )` pins the guard to core's actual behaviour instead of to a restatement of it, which is the only way to catch the guard drifting looser than what `wp_safe_redirect()` will really send.

**One thing I went looking for and did not find**, worth recording since it was the way this change could have quietly broken working URLs: `esc_url_raw()` on `REQUEST_URI` cannot blank out a legitimate path. I expected a colon in a query (`?q=16:9`) to reach `wp_kses_bad_protocol()` and make `esc_url()` return `''`. It cannot, because `esc_url()` short-circuits at `formatting.php:4557` (`if ( '/' === $url[0] ) { $good_protocol_url = $url; }`) and never runs the protocol check for a path-relative URL. So the sanitiser is lossless for every shape `REQUEST_URI` can take here.

Two things left.

**1. The `allowed_redirect_hosts` widening still looks like it costs more than it buys.** `get_target_url()` now ends in `'/' . ltrim( $path, '/' )`, so the appended remainder always starts with exactly one slash and the authority is entirely determined by the admin-configured origin. There is no longer a request that can steer the host. `wp_safe_redirect()` is therefore protecting a path that cannot be steered, while `allow_standalone_redirect_host()` widens the safe-redirect allow-list for every other caller on the subsite, `redirect_to` on login included. You would get an equivalent and strictly narrower guard by comparing `$parts['host']` against `wp_parse_url( get_theme_mod( 'ov_redirect_url', OPENVERSE_STANDALONE_URL ), PHP_URL_HOST )` and going back to `wp_redirect()`, which runs `wp_sanitize_redirect()` too, so nothing is lost. Your call, and the description flags the trade-off honestly, but I would take the narrower version.

**2. The `esc_html()` in `index.php` deserves top billing in the description, not the last bullet.** Before this branch, `echo "<!-- " . $target_url . " -->"` put raw `REQUEST_URI` into an HTML comment, and `use_index_php_as_template()` calls `status_header( 200 )` so every path under the site renders. `/openverse/--><script>...` was a reflected XSS whenever `ov_is_redirect_enabled` was off. That is the highest-severity thing in the PR, and a committer reading the summary would not know it is in here.

Nit: every suite added to this repo recently uses `phpunit.xml.dist`; this adds `phpunit.xml`. Worth matching so a developer can still drop a local override alongside it.

mcliwanow:
Took the first one, ec519ee76. allowed_redirect_hosts and wp_safe_redirect() are gone. The guard compares the target's host against the configured origin instead, and wp_redirect() runs wp_sanitize_redirect() just the same. Narrower than what it replaced, since the allow-list also accepted the home host. get_standalone_origin() gives both callers one reading of the setting.

Nit taken, phpunit.xml.dist.

bor0 (commented):
Re-reviewed. Everything from the last round is closed, and the fix is a net security improvement. Two things I checked specifically and cleared:

- `esc_url()`'s bad-protocol branch, the one that can return `''`, is skipped entirely for URLs whose first character is `/`. So a colon anywhere in the path or query cannot blank the request and 301 a visitor to the origin root. I replayed core's real `esc_url()` against the mapping table to confirm.
- `'/' . ltrim( $path, '/' )` genuinely closes the authority-injection surface, including absolute-form request targets (`GET https://evil.com/ HTTP/1.1`) and `//evil.com` shapes, both of which the old `str_replace()` version mishandled.

`is_valid_target_url()` now correctly rejects the schemeless and scheme-relative origins from last round. Two low-severity notes below.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -255,25 +255,97 @@ function wporg_ov_customizer( $wp_customize ) {
...
+ * @return bool
+ */
+function is_valid_target_url( $target_url ) {
+	$parts = wp_parse_url( $target_url );
+
+	if ( empty( $parts['host'] ) || empty( $parts['scheme'] ) ) {
+		return false;
+	}
+
+	if ( ! in_array( $parts['scheme'], array( 'http', 'https' ), true ) ) {
```
bor0: The scheme check is case-sensitive, but `parse_url()` preserves scheme case:

```
php > var_dump( parse_url( 'HTTPS://Openverse.ORG/x' ) );
  ["scheme"] => string(5) "HTTPS"
  ["host"]   => string(13) "Openverse.ORG"
```

Schemes are case-insensitive per RFC 3986, and the old code redirected fine on such a value, so an origin saved as `HTTPS://openverse.org` silently disables the redirect site-wide on input that is legal. `Target_Url_Test.php:296` encodes that as the expected result.

`strtolower( $parts['scheme'] )` before the comparison keeps the guard and accepts the value. Host case is fine as-is, both sides of the comparison come from the same setting.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
```diff
@@ -255,25 +255,97 @@ function wporg_ov_customizer( $wp_customize ) {
...
 	}
 
-	$path = $_SERVER['REQUEST_URI'];
-	if ( $path ) {
-		$count = 1; // Only replace the leading Openverse subpath.
-		$target_url .= str_replace( OPENVERSE_SUBPATH, '', $path, $count );
+	// Sanitising is required by WordPress.Security.ValidatedSanitizedInput.
+	// Not `sanitize_text_field()`, which strips percent-encoded octets and
+	// would turn `?q=cat%20dog` into `?q=catdog`.
+	$path = isset( $_SERVER['REQUEST_URI'] ) ? esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) ) : '';
```
bor0, line 276: One `esc_url_raw()` rewrite the description's list does not cover. Beyond dropping characters, `esc_url()` also runs `str_replace( ';//', '://' )`:

| Request | Forwarded |
| --- | --- |
| `/openverse/search/?q=http;//example.com` | `?q=http://example.com` |

Neither the raw `REQUEST_URI` nor `wp_sanitize_redirect()` made that substitution, so it is the one change the "`wp_sanitize_redirect()` already dropped the same ones" rationale does not extend to. Harmless in effect, the host stays pinned to the origin, but it belongs in the list next to the bracket row.

bor0 (approved):
Approving. The subpath fix is correct, and I verified the two properties it rests on against core's real `esc_url()` rather than by inspection: the bad-protocol branch is skipped for URLs starting with `/`, and `'/' . ltrim( $path, '/' )` keeps the remainder in the path rather than the authority, including for absolute-form request targets. `is_valid_target_url()` closes the schemeless and scheme-relative origins from the earlier round.

The two notes in my review above are both one-liners and neither changes the shape of the patch. The `strtolower()` on the scheme is worth taking before this is committed (along with flipping the `upper case scheme` expectation in `Target_Url_Test.php`); the `;//` row is a description edit. Happy for either to land here or to be picked up by the committer.

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
mcliwanow: 😬 updated in https://github.com/WordPress/wordpress.org/pull/852/changes/715c71ede1a2268c59811f4c52518bb75aaf08ae

wordpress.org/public_html/wp-content/themes/pub/wporg-openverse/functions.php
mcliwanow: Added as its own row in the description

PR #860, opened 2026-09-02
Photo Directory: render the photo description as plain text

mcliwanow:
## Why

The description submitted with a photo is its alternative text. The submit form says "No HTML" and caps it at 350 characters, and since 9fc0f0ae8 `Uploads::sanitize_submitted_description()` stores it as plain text. On output it was still treated as post content, so text that looks like markup was interpreted instead of displayed.

## What changed

`Posts::render_content_as_plain_text()` runs first on `the_content` for the photo post type and escapes the description, so every later callback sees text. Paragraphs still apply. Texturize still runs, but quotes and apostrophes no longer curl since they are already escaped when it gets there. Other post types are untouched.

## Testing

Verified on wp-env against WordPress master with the plugin active: a description round-trips through intake unchanged, renders as text on the single photo page (through the template over HTTP, through `apply_filters( 'the_content' )`, and in the excerpt), and a page with `[gallery]` still expands it.

Of the existing photos, two published ones (24724, 28413) carry old `<a>`/`<p>` markup from before the intake sanitizer and will show it as text. Their descriptions get edited to plain text alongside the deploy.


bor0 (commented):
The diagnosis is right and the shape of the fix is right: the description is plain text on the way in, and `wp:post-content` was handing it to the whole `the_content` stack on the way out. Escaping at `PHP_INT_MIN` so that `do_blocks`, `do_shortcode` and `run_shortcode` only ever see text is the correct layer, and `&#91;` covers both `do_shortcode()` at 11 and `WP_Embed::run_shortcode()` at 8.

I checked the surrounding claims against a real WordPress checkout. Two of the three hold. One does not.

**`wpautop()` does not prevent auto-embedding. It creates the exact shape that triggers it.**

`WP_Embed::autoembed()` makes two passes (`class-wp-embed.php:446-451`):

```php
// Find URLs on their own line.
$content = preg_replace_callback( '|^(\s*)(https?://[^\s<>"]+)(\s*)$|im', ... );
// Find URLs in their own paragraph.
$content = preg_replace_callback( '|(<p(?: [^>]*)?>\s*)(https?://[^\s<>"]+)(\s*</p>)|i', ... );
```

Wrapping first defeats the first pass and walks straight into the second. I ran core's `wpautop()` standalone over `esc_html( 'https://youtu.be/abc123' )` and it returns `<p>https://youtu.be/abc123</p>\n`, which matches the second pattern with `$2 === 'https://youtu.be/abc123'`. So a description that is a single URL still reaches `WP_Embed::shortcode()` after this change, exactly as it does on trunk.

That the testing did not catch it is easy to explain: `autoembed_callback()` sets `linkifunknown = false`, so a URL with no oEmbed provider comes back byte-identical and the output looks untouched. A provider URL (YouTube, WordPress.tv, Twitter) does not.

It is worth fixing rather than deferring, because the consequences are not only cosmetic. On a front-end view `$this->usecache` is true but the cache starts empty, so `WP_Embed::shortcode()` falls through to `wp_oembed_get()`, which is an outbound HTTP request during page render, and then `update_post_meta( $post_id, '_oembed_' . md5( ... ), $html )` writes to the photo post (`class-wp-embed.php:307-316`). A write and a remote fetch on a public GET, driven by submitted text. This is pre-existing rather than introduced here, but the PR is presented as closing it and does not.

Options, roughly in order of how much I like them:

* Unhook the embed filters for photo requests, for example `remove_filter( 'the_content', array( $GLOBALS['wp_embed'], 'autoembed' ), 8 )` and the matching `run_shortcode` at 8, guarded the same way the escape is. Explicit about what is being turned off, and no longer relies on out-guessing a regex.
* Make the follow-up you describe (`wp:post-content` to a plain-text block in `wporg-photos-2024`) part of this change rather than optional. If the content never goes through `the_content` for the single view, the whole class of problem goes away, and the `the_content` filter is left as defence for REST, feeds and excerpts.
* At minimum, correct the comment. `// Wrap paragraphs now, so a URL on a line of its own is not auto-embedded.` will send the next reader in the wrong direction.

**The two claims that do hold:**

* `wpautop()` is idempotent over this output, so core's own `wpautop` at priority 10 running a second time does not produce nested `<p>`. I ran it twice over the escaped forms of your test strings and got identical results, including `<p>line1<br />\nline2</p>` for a soft break.
* The global-post check is sound for the paths named in the description. `WP_REST_Posts_Controller::prepare_item_for_response()` assigns `$GLOBALS['post']` and calls `setup_postdata()` before it renders `content.rendered` (`class-wp-rest-posts-controller.php:1889-1891`), and `wp_trim_excerpt()` and the feed both run inside the loop, so `get_post_type()` sees the photo in all three.

Smaller notes:

* Keying on `get_post_type()` with no argument keys on the global post, not on the content being filtered. Any `apply_filters( 'the_content', $something_else )` that runs while a photo is the global post gets escaped too. On this site that is unlikely to bite, and the existing `Uploads` filters are on a page rather than a photo, so they are unaffected. Mentioning it only so the constraint is written down somewhere.
* `plain_text_to_html()` has no caller but `render_content_as_plain_text()`. `private` unless you have a second use in mind.
* Excerpts get `strip_shortcodes()` applied to the raw content by `wp_trim_excerpt()` before any `the_content` callback runs, so a bracketed word still vanishes from the excerpt even after this change. Pre-existing, out of scope, but it means the excerpt and the body will not agree.

mcliwanow:
Good catch on `autoembed()`, I missed the second pass. I went a different way than your first option after trying it: unhooking and rehooking the embed callbacks needed state to survive nested calls, which is more machinery than this needs. The URL delimiter is now encoded the same way `[` is, so the embed callbacks stay in place and never match. Checked with a description that is only a YouTube URL: text, no iframe, no `_oembed_*` meta.

`plain_text_to_html()` is gone, the global-post note is in the docblock, and the excerpt point is pre-existing, leaving it.

wordpress.org/public_html/wp-content/plugins/photo-directory/inc/posts.php
```diff
@@ -256,6 +259,32 @@ public static function use_photo_url_instead_of_media_permalink_url( $url, $post
...
+	 * text.
+	 *
+	 * Keys on the global post, like core's own content callbacks, so it applies
+	 * to whatever 'the_content' is run for while a photo is the current post.
+	 *
+	 * @param string $content Post content.
+	 * @return string
+	 */
+	public static function render_content_as_plain_text( $content ) {
+		if ( Registrations::get_post_type() !== get_post_type() ) {
```
bor0, line 281: The guard keys on the global post, not on the post whose content is being filtered.

`get_post_type()` with no argument reads `$GLOBALS['post']`, but `the_content` is routinely applied to a *different* post's content. Core's `wp_trim_excerpt()` is the clearest case: it resolves `$post = get_post( $post )` and then calls `apply_filters( 'the_content', $text )` on that post's content while leaving the global alone.

So `get_the_excerpt( 24724 )` from outside the loop (a widget, a sidebar list, a page template) skips the escaping entirely and the legacy `<a>`/`<p>` markup goes through, which is the case this patch exists to prevent. `wp_trim_words` strips tags afterwards, so the practical damage is limited, but the guarantee has a hole in it rather than a documented tradeoff.

The docblock does own the other direction (a non-photo post filtered while a photo is the global post gets escaped and shows its markup as text). Worth noting `WP_REST_Posts_Controller` is safe here, it calls `setup_postdata()` first.

wordpress.org/public_html/wp-content/plugins/photo-directory/inc/posts.php
```diff
@@ -256,6 +259,32 @@ public static function use_photo_url_instead_of_media_permalink_url( $url, $post
...
+			return $content;
+		}
+
+		$content = esc_html( $content );
+
+		// Shortcode and URL syntax stay visible text: hide the characters shortcodes and embeds key on.
+		return str_replace( [ '[', '://' ], [ '&#91;', '&#58;//' ], $content );
```
bor0, line 288: `esc_html()` calls `_wp_specialchars( $text, ENT_QUOTES, 'UTF-8', false )`, and with `$double_encode = false` an already-valid entity passes through untouched. A description submitted as the literal nine characters `&amp;amp;` survives `sanitize_textarea_field` intact and then renders as a single `&`, and `&#91;gallery&#93;` renders as `[gallery]`.

Neither is exploitable, `do_shortcode` matches only a literal `[`, but it contradicts the stated contract that the submitted text is shown verbatim. `htmlspecialchars( $content, ENT_QUOTES, 'UTF-8', true )` would be faithful.

bor0 (commented):
Re-reviewed. The approach holds up: `PHP_INT_MIN` puts the escape ahead of `WP_Embed::run_shortcode`/`autoembed` (8), `do_blocks` (9), `wptexturize`/`wpautop` (10) and `do_shortcode` (11), and neutralising `[` and `://` closes the shortcode and autoembed paths. I could not construct markup through `esc_html`'s `double_encode = false` behaviour: existing character references are preserved and the parser renders them as text.

Three things below, none blocking.

wordpress.org/public_html/wp-content/plugins/photo-directory/inc/posts.php
```diff
@@ -35,6 +35,9 @@ public static function init() {
 		// Sync photo post content to photo media on update.
 		add_action( 'post_updated',       [ __CLASS__, 'sync_photo_post_to_photo_media_on_update' ], 5, 3 );
 
+		// Photo content is plain text (the alternative text), never post markup.
+		add_filter( 'the_content', [ __CLASS__, 'render_content_as_plain_text' ], PHP_INT_MIN );
```
bor0, line 39: Knock-on effect one file down, at `posts.php:407`: the RSS `alt` attribute is now double-escaped.

`add_photo_to_rss_feed()` is on `the_content_feed`, which core feeds the *output* of `the_content`. It does `strip_tags( $content )` and passes the result as `[ 'alt' => $content ]` to `get_the_post_thumbnail()`, and `wp_get_attachment_image()` runs every attribute through `esc_attr()`.

Before this change, a description `Fish & chips` reached that point as raw text and `esc_attr` produced the correct `alt="Fish &amp; chips"`. Now `esc_html` has already produced `Fish &amp; chips`, so the feed emits `alt="Fish &amp;amp; chips"` and the entity is announced literally. Same for `<`, `>` and `"`.

The `<figcaption>` on the following line is unaffected, it is inserted as HTML.

wordpress.org/public_html/wp-content/plugins/photo-directory/inc/posts.php
mcliwanow: That's intentional. Intake goes through `wp_filter_post_kses()`, which stores a typed `&` as `&amp;`, so the output has to leave existing entities alone or every ampersand ever submitted shows up as `&amp;`. From the sandbox, including a real published row:

image

wordpress.org/public_html/wp-content/plugins/photo-directory/inc/posts.php
mcliwanow: `esc_attr()` doesn't double encode. It uses the same `_wp_specialchars()` call as `esc_html()`, with `double_encode` off, so an `&amp;` that is already there stays `&amp;`. Checked the patched path on the sandbox:

Screenshot 2026-09-03 at 10 27 01

Encoded once, same as before the change.

wordpress.org/public_html/wp-content/plugins/photo-directory/inc/posts.php
mcliwanow: Agreed, documented in the docblock now. Nothing in the plugin or the theme builds a photo excerpt outside the loop, so it's a known edge rather than a path anything uses.

bor0:
Reviewed this against the branch, and against a WordPress checkout for the parts that depend on core's own ordering.

The approach holds up:

* Intake really does store plain text, so this is output-side defence in depth plus a fix for the legacy rows, which is how the description frames it. `Uploads::sanitize_submitted_description()` maps `post_content` through `sanitize_textarea_field`, which escapes `<` and strips tags.
* The `PHP_INT_MIN` placement does what it needs to. It lands ahead of `WP_Embed::run_shortcode` and `autoembed` (8), `do_blocks` (9), `wptexturize` and `wpautop` (10), and `do_shortcode` (11), so hiding `[` and `://` genuinely keeps shortcodes and embeds from firing.
* No cross-talk with the plugin's other `the_content` callbacks. `Uploads::insert_upload_disallowed_notice` (priority 1) and `insert_upload_form` (10) are both gated on `is_page( SUBMIT_PAGE_SLUG )`, so neither ever sees a photo, and neither has its own markup escaped by running later.
* The global-post keying also holds for the REST response: `WP_REST_Posts_Controller::prepare_item_for_response()` assigns `$GLOBALS['post']` and calls `setup_postdata()` before applying `the_content` (`class-wp-rest-posts-controller.php:1889-1891`).

### One correction to the description

"Paragraphs and texturize still apply" is only half true. `wptexturize` does still run, but it now runs on already-escaped text, where every `'` is `&#039;` and every `"` is `&quot;`. So quote and apostrophe curling silently stops for photo descriptions, while ellipses and dashes still texturize. That seems like an acceptable price for escaping first, and escaping last is not an option since it would eat `wpautop`'s own tags, but the text should say so rather than imply nothing changed there.

### Smaller notes

* `esc_html()` does not double-encode, so a description in which someone literally typed `&amp;` now renders as `&`. Negligible, and not a safety question: an entity such as `&#91;` or `&#58;` passed through from storage stays inert, because shortcode and embed parsing runs against the escaped string and never sees a real `[` or `://`.
* The docblock is candid about the global-post trade-off, which I appreciate. I went looking for something on the site that would trip it and did not find anything.
* This is the only one of the three related PRs with no automated test. The plugin has no PHPUnit suite to wire into, so that is understandable, but it does mean the behaviour rests entirely on the manual wp-env pass.

mcliwanow:
Thanks. Fixed the texturize line in the description, quotes stop curling because they are escaped first.

bor0 (approved):
Re-reviewed at `df1870b`. Approving.

The description now states the quote-curling change, so my earlier correction is addressed.

**New this round: the RSS path checks out.** `Posts::add_photo_to_rss_feed()` runs on `the_content_feed`, after `the_content`, and reuses the escaped string twice, as `'alt' => $content` on `get_the_post_thumbnail()` and raw inside `<figcaption>`. Neither leaks literal entities: `esc_attr()` does not double-encode, so `&#91;` and `&#58;//` render as `[` and `://` in the alt attribute and in the caption, which is what you want.

Everything from my previous pass still holds: intake really does store plain text via `sanitize_textarea_field`, so this is output-side defence in depth plus a fix for the legacy rows; the `PHP_INT_MIN` placement lands ahead of `WP_Embed::run_shortcode` and `autoembed` (8), `do_blocks` (9), `wptexturize` and `wpautop` (10) and `do_shortcode` (11); the plugin's other `the_content` callbacks are gated on `is_page( SUBMIT_PAGE_SLUG )` so they never cross paths with a photo; and the global-post keying holds for REST, since `WP_REST_Posts_Controller::prepare_item_for_response()` assigns `$GLOBALS['post']` and calls `setup_postdata()` before applying `the_content`.

### Caveat attached to this approval

The approval takes on the operational step in the description: photos 24724 and 28413 carry pre-sanitizer `<a>`/`<p>` markup and will render it as visible text the moment this deploys. Their descriptions need editing to plain text **alongside** the deploy, not after it. Worth carrying into the commit message so whoever lands it in SVN does not lose the dependency.

### One thing the escaping does not cover, out of scope here

`convert_smilies` sits on `the_content` at priority 20, so it runs after the escape, and it converts text rather than markup. There is nothing for the `[` and `://` masking to catch:

```
Great shot at f/8 :) See &#91;fig 8) here] ...
                     ^^                ^^
```

Both `:)` and `8)` are in `$wpsmiliestrans`, so a description reading "see fig 8)" renders an `` in a field whose whole purpose is to be the photo's alternative text. This is pre-existing rather than a regression, but after this change it is the last `the_content` callback that still turns a photo description into markup, which sits a little awkwardly against the premise. A separate ticket rather than a reason to hold this one.

Still the only one of the three related PRs with no automated test, for the reason you gave: the plugin has no suite to wire into.

PR #861, opened 2026-09-02
Badge Management: map the badge manager capability against the site

mcliwanow:
## Why

Tools → Profile Badges is gated on `wporg_profile_badge_required_cap`, a per-site setting with three choices: `publish_posts` ("Can publish posts (Author+)"), `manage_options` and `manage_network`. The setting describes a role on the site, and r14547 introduced it as "those who can publish on the P2". The check was `current_user_can()` on the raw value, which also counts capabilities that filters add to users without a role on the site, so the check and the label didn't describe the same set of users. The same value gated the settings tab that writes it.

## What changed

- One meta capability, `wporg_manage_profile_badges`, for the screen, the write handler and the settings handler. It maps to the configured capability for users with a role on the site and to `do_not_allow` for everyone else. Super admins pass as before. `add_submenu_page()` takes it directly.
- Changing the capability needs `manage_options` on top of being able to manage badges, so a site administrator can't lower a `manage_network` setting. The rest of the settings tab stays with the badge managers. A posted value the user can't set, or one outside the choices, ends in `wp_die()` before anything is written.
- `get_required_cap()` falls back to `manage_options` for a stored value outside the choices, and the choices live in one place.
- The notice and the description say "users with a role on this site".

Nothing changes at the default: administrators are members with `manage_options`. A site on `publish_posts` keeps its Author+ members.

## Testing

PHPUnit suite added under the plugin and wired into the make test env next to the o2 Posting Access one. Nine tests over the three settings, membership, super admins, logged out, who may change the capability, and the read fallback.

    cd environments && WP_ENV_PORT=8902 npm run make:test:badge-management

Also driven over HTTP on that env as a super admin, an administrator, an author, a subscriber and an account with no role on the site. That needed a local stand-in for `mu-plugins/pub/profile-helpers.php`, which isn't part of the env, so only the phpunit suite is expected to run there as committed. phpcs is clean on the changed lines.

## Summary by CodeRabbit

- **New Features**
  - Added configurable badge-management permissions, including support for site roles and super administrators.
  - Badge-management settings now explain eligibility and indicate when changes are restricted to administrators.
  - Invalid permission settings safely fall back to the default administrator capability.

- **Bug Fixes**
  - Improved authorization checks to prevent unauthorized badge-management and settings changes.
  - Unauthorized permission submissions now return an access-denied response.
  - Stale permission selections are ignored when users cannot modify the setting.
  - Clarified that network-level badge management is limited to super administrators.


bor0 (commented):
Reviewed the branch locally and checked the premise and the mapping against WordPress core rather than from memory.

**The hole is real.** `wporg-o2-posting-access.php:275-285` grants `publish_posts`, `edit_posts` and `edit_published_posts` through `user_has_cap` to any logged-in user who is not a member of the blog. So on a Make site with `wporg_profile_badge_required_cap` set to `publish_posts`, the old `current_user_can( $required_cap )` in `admin-post.php` admitted every logged-in WordPress.org account to Tools > Profile Badges, and to `admin_post_wporg_profile_manage_badges`. The setting's own label ("Author+") described a role on the site, so the check and the label really did describe different sets of users.

**The mapping is correct.** I traced it through `WP_User::has_cap()`:

* `map_meta_cap()` runs before the super admin short-circuit, and that short-circuit honours `do_not_allow`, so the explicit `is_super_admin()` branch is load-bearing and correctly placed. Without it a super admin who holds no role on the site would be refused.
* `user_has_cap` (o2's grant) runs after `map_meta_cap`, so it cannot resurrect a `do_not_allow`. The denial holds.
* `is_user_member_of_blog()` returns true unconditionally when `! is_multisite()`, which is exactly why the bootstrap defines `WP_TESTS_MULTISITE`. Without it every non-member assertion would pass for the wrong reason.
* No recursion risk: `is_super_admin()` reads `site_admins` on multisite, and its single-site `delete_users` path re-enters the filter, which early-returns on the cap name.

Nothing outside the plugin reads either option, so the meta cap is self-contained. Adding the plugin to `.wp-env.test.json` is also the right place for the CI wiring, since per the comment at `.github/unit-tests-suites.yml:14-16` the workflow derives the make suite's path filter from that file. `WP: Make` is green.

Four things.

**1. `get_required_cap()` does not validate on read.** The write path now checks membership in `get_required_cap_choices()`, but the getter returns whatever the option holds, and that value becomes the primitive capability the meta cap maps to. This is not a regression (the old code fed the raw option to `current_user_can()` and `add_submenu_page()` alike), but since you are introducing the getter, falling back to `manage_options` for anything not in the choices list makes the read path self-defending for one line.

**2. A rejected capability save is silent.** `if ( array_key_exists( ... ) && current_user_can( $required_cap ) )` drops the value while still saving the note and the badges, with no notice. The form's `disabled()` means this only fires on a crafted POST, so it is minor, but `wp_die( 'Unauthorized user' )` there would match how the rest of the handler treats input it will not accept.

**3. Gating the whole Settings tab on `manage_options` is a separate tightening, and worth a decision rather than a side effect.** The security argument for the capability dropdown is solid: a site administrator should not be able to move a `manage_network` setting down to one they hold. But the same gate also takes "Note to Badge Managers" away from the badge managers, and on a site set to `publish_posts` those are exactly the people the note is addressed to. Consider keeping the tab visible to anyone who can manage badges and gating only the `required_role` select and its save on `manage_options`. If the intent is that team reps should not edit the note either, that is fine, but it should be stated as the intent.

**4. `phpunit/includes/testcase.php` is byte-identical to `wporg-o2-posting-access/phpunit/includes/testcase.php`** apart from the class name and two docblock lines. I diffed them. handbook and trac-watcher carry near-identical variants, so this makes four. Not a blocker for this PR, but this is the point where extracting a shared base under `common/` starts paying for itself, and it would be a good follow-up.

**On the tests:** the two guards are the right ones. `create_non_member()` asserting `publish_posts` before anything else means the non-member cases cannot pass because the o2 grant silently stopped applying, and `assertTrue( grant_super_admin( ... ) )` catches the `$GLOBALS['super_admins']` override making the super admin fixture a no-op. Both are the failure modes that would otherwise leave a green suite testing nothing.

mcliwanow:
Thanks for checking it against core.

1 and 2 applied: `get_required_cap()` falls back to `manage_options`, and a rejected `required_role` ends in `wp_die()` before anything is written.

3: you're right, that was a side effect. The tab is back for everyone who can manage badges, only the select and its save are limited to admins. Helper renamed to `current_user_can_change_required_cap()`.

4: agreed, follow-up. Suite is 9 tests now.

bor0:
Re-reviewed the branch. **The core fix is correct.**

`wporg-o2-posting-access` grants `publish_posts` through the `user_has_cap` filter, but `WP_User::has_cap()` rejects `do_not_allow` from the `caps` list regardless of what that filter puts into `allcaps`, so returning `[ 'do_not_allow' ]` for non-members genuinely closes the hole. The super admin short-circuit, the logged-out path, and the ordering of the `||` chain in `save_settings()` (so `array_key_exists()` runs before `current_user_can( $required_cap )`, never passing an arbitrary string to a cap check) all check out. The whitelist fallback in `get_required_cap()` also incidentally prevents the option from ever being set to `wporg_manage_profile_badges` itself, which would have recursed. Dropping the old `else 'manage_options'` default is a real fix on its own: a submission without `required_role` used to silently reset the setting. The new `esc_html()` closes an unescaped option echo.

I found no correctness bug in the authorization logic. The remaining points are minor.

### Findings

**`environments/make/.wp-env.test.json:7` — settings page fatals in the env.** Adding the plugin to `plugins` activates it on the env's dev site, but `WordPressdotorg\Profiles\get_badges()` lives in `wordpress.org/public_html/wp-content/mu-plugins/pub/profile-helpers.php`, which is not mapped into this environment. Visiting Tools > Profile Badges > Settings dies with "Call to undefined function" at `admin.php:251`. The entry also is not needed for the suite: the PHPUnit bootstrap already loads the plugin manually at `muplugins_loaded`.

**`admin.php:42` — notice text is wrong for the `manage_network` setting.** It asserts a role on the site is required, but `map_badge_manager_cap()` exempts super admins from the membership check. A site set to `manage_network` renders "Users with a role on this site and the `manage_network` capability can manage badges", when the people who actually can are precisely the ones who typically hold no role on that site.

**`admin-post.php:98` — a rejected cap change discards the whole submission.** `wp_die( 'Unauthorized user' )` fires before `note_to_team` is saved, so any disallowed `required_role` throws away the rest of the form, including a long note the user just typed, with a bare unstyled error and no back link. Reachable by an administrator posting `required_role=manage_network`. Ignoring the disallowed capability change while still saving the rest, or a `wp_die` with a title and back link, would be friendlier.

**`admin.php:233` — the Super Admin option is silently unselectable.** `disabled( ! current_user_can( $cap ) )` on each option, combined with the `! current_user_can( $required_cap )` guard at `admin-post.php:95`, means a site administrator can never select "Can manage network (Super Admin)". A team admin wanting to lock badge management down to super admins gets a greyed out control with no explanation, since the "Only administrators can change this setting" hint at line 241 is not rendered for them (`current_user_can_change_required_cap()` is true). If the restriction is intended, the option description should say so.

**`index.php:88` — new membership lookup on the admin hot path.** Switching `add_submenu_page()` to the meta cap means every admin page load on every site with this plugin active now runs `is_super_admin()` plus `is_user_member_of_blog()` (which calls `get_blogs_of_user()` -> `get_user_meta()` + `get_sites()`), where it previously ran a plain `current_user_can( 'manage_options' )` against already loaded caps. Mostly absorbed by the persistent object cache, but worth noting.

**`index.php:73` — docblock sentence is missing words.** "Administrators who can manage badges themselves, so a setting that excludes site administrators can't be changed by one." does not parse. Presumably intended as "Requires `manage_options` on top of being able to manage badges, so ...".

### Checked, no action needed

- The disabled `<select>` is not relied on for authorization: `save_settings()` re-checks `current_user_can_change_required_cap()` server side.
- `wporg-o2-allow-crossposting.php` only grants `edit_posts`, and only during xpost/xcomment transitions, so it cannot widen the badge manager set.
- No recursion risk from `is_super_admin()` inside the `map_meta_cap` filter on single site: `has_cap('delete_users')` returns early from the filter.
- The new `phpunit/includes/testcase.php` and `phpunit/bootstrap.php` are byte identical to the existing o2 Posting Access ones apart from names, comments, and the extra o2 plugin require. `create_non_member()` correctly deletes the capabilities meta key (an emptied role would still count as membership) and self-asserts against becoming vacuous.
- No other consumer of `wporg_profile_badge_required_cap` exists in the repo, so the `get_required_cap()` whitelist fallback has no external callers to break.

mcliwanow:
All in one commit.

- Notice: separate text for `manage_network`, and the other one now mentions super admins.
- Rejected cap: `wp_die()` with a title, 403 and a back link. It runs before anything is saved, so the note isn't half-applied; the select is disabled for anyone this can happen to.
- Super Admin option: pre-existing, and intended (a site admin picking it would lock themselves out). One sentence under the select says so.
- Docblock fixed, and `get_required_cap()` checks `is_string()` first.

The `.wp-env.test.json` entry stays: as you noted in the first pass, the workflow derives the make suite's path filter from it, and it's also what mounts the plugin into the container. The membership lookup on the admin hot path is real but cached user meta, and o2 Posting Access already does the same per capability check on those sites.

environments/make/.wp-env.test.json
```diff
@@ -4,6 +4,7 @@
 	"testsEnvironment": false,
 	"plugins": [
 		"../wordpress.org/public_html/wp-content/plugins/wporg-o2-posting-access",
+		"../wordpress.org/public_html/wp-content/plugins/wporg-badge-management",
```
bor0: Adding the plugin here also activates it in the make env, which surfaces Tools → Profile Badges.

`render_settings()` calls `WordPressdotorg\Profiles\get_badges()`, which lives in `wordpress.org/public_html/wp-content/mu-plugins/pub/profile-helpers.php`. The only mapping in this config is `wp-content/env-bin`, so that file is never loaded, and the `bpmain_wporg_groups` table it queries is not stubbed in `make/bin/database-tables.sql` either.

Concretely: `npm run make:test:env -- start`, log in, open Tools → Profile Badges → Settings, and you get a fatal `Call to undefined function WordPressdotorg\Profiles\get_badges()`. The Add/Remove tab survives only because it returns early on the empty `wporg_profile_badges` option.

The phpunit suite is unaffected, nothing in it reaches `render_settings()`. But it does mean the manual HTTP verification in the description is not reproducible by anyone else as written. Either stub the table and map the mu-plugin, or note that only the phpunit suite is expected to work here.

wordpress.org/public_html/wp-content/plugins/wporg-badge-management/admin-post.php
```diff
@@ -79,21 +78,38 @@ function manage_badges() {
...
+	$required_cap = isset( $_POST['required_role'] ) ? sanitize_text_field( wp_unslash( $_POST['required_role'] ) ) : null;
+	if (
+		null !== $required_cap &&
+		(
+			! current_user_can_change_required_cap() ||
+			! array_key_exists( $required_cap, get_required_cap_choices() ) ||
+			! current_user_can( $required_cap )
+		)
+	) {
+		wp_die(
```
bor0, line 102: This discards the whole submission, including the `note_to_team` edit, when the posted `required_role` is one the user may not set.

The only thing normally keeping a non-administrator badge manager from posting that field is the client-side `disabled` attribute rendered in `admin.php`. Concrete path: an administrator on a `publish_posts` site renders the Settings tab, is demoted to Author in another tab, then saves a note edit. The stale form still posts `required_role=publish_posts`, `current_user_can_change_required_cap()` is now false, and they get a 403 with the note silently lost.

Ignoring the field when `! current_user_can_change_required_cap()`, and reserving `wp_die()` for out-of-range or unheld values from users who *can* change it, keeps the rest of the tab working. That also reads closer to the intent in the description: the select is for administrators, the rest of the tab is for badge managers.

wordpress.org/public_html/wp-content/plugins/wporg-badge-management/phpunit/includes/testcase.php
```diff
@@ -0,0 +1,197 @@
...
+ * test before a single assertion runs. This mirrors the approach the handbook
+ * plugin's suite takes.
+ *
+ * @package wporg-badge-management
+ */
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Provides WordPress fixtures and multisite membership helpers.
```
bor0, line 20: This file is a verbatim copy of `wporg-o2-posting-access/phpunit/includes/testcase.php`, class name and two docblock lines aside. With the handbook plugin's variant that is three copies of the same PHPUnit 10 harness in the repo, so any fix to the transaction or hook-restore logic now has to land in three places.

Not this PR's debt to pay on its own, but worth extracting somewhere shared, or at least naming which copy is the source of truth.

bor0 (commented):
Re-reviewed. The authorization change is correct, and I re-checked the pieces carrying it:

- `WP_User::has_cap()` honours `do_not_allow` ahead of its super admin bypass, so super admins pass in every branch and are never denied.
- No recursion: `is_super_admin()` resolves `manage_network`, which short-circuits out of the filter on the first line.
- The o2 Posting Access grant this neutralises (`add_post_capabilities()`) only fires for non-members, so the membership gate is exactly the right complement, and a subscriber member on a `publish_posts` site is still correctly excluded.
- There is no state where the `<select>` is enabled but its selected `<option>` is disabled, and `manage_network` is not a lockout, since a super admin still satisfies both halves of `current_user_can_change_required_cap()`.
- Nothing else in the repo reads `wporg_profile_badge_required_cap` or the `profile-badges` slug.

Three low-severity notes below.

bor0 (approved):
Approving. The capability mapping is correct and I traced every path that carries it: `do_not_allow` beats the super admin bypass in `WP_User::has_cap()`, `is_super_admin()` short-circuits out of the filter so there is no recursion, and the membership gate is exactly the complement of the o2 Posting Access grant, which only fires for non-members. The screen, the write handler and the settings handler now agree on one capability, and nothing else in the repo reads the old option.

None of the three notes in my review above affect who can manage badges. The `wp_die()` one is worth taking since it is cheap, and the make env caveat is worth a line either way, but neither needs to block this.

wordpress.org/public_html/wp-content/plugins/wporg-badge-management/admin-post.php
mcliwanow: Taken as you describe in https://github.com/WordPress/wordpress.org/pull/861/changes/8c7c4ab4905cd1d0e2614874d6b97dc5892c5d55 . The field is ignored from anyone who can't change the capability, so a stale form saves the note and leaves the setting alone. wp_die() is left for an unheld or out-of-range value from someone who can. Checked both paths over HTTP.

wordpress.org/public_html/wp-content/plugins/wporg-badge-management/phpunit/includes/testcase.php
mcliwanow: 👍 for now names the source of truth in https://github.com/WordPress/wordpress.org/pull/861/changes/eb9d29af53dd7d45de48240b7ace673728b553e1

environments/make/.wp-env.test.json
mcliwanow: You're right that it activated it. Moved the plugin from plugins to mappings in https://github.com/WordPress/wordpress.org/pull/861/changes/4be7f67a01163a353689146fa2886f6f0e3ae225, so it's mounted for the suite but never active: no Tools → Profile Badges in that env, nothing to fatal. The CI path filter reads the plugins list, so the plugin's path is listed under the make suite in unit-tests-suites.yml instead. Checked on a fresh env: plugin present and inactive, suite green, no menu entry.

PR #871, opened 2026-09-04
Theme Directory: refuse a shortcode in the name and author headers

bor0:
Follow-up to [15147], which handled `Description:` this way. The `Theme Name:` and `Author:` headers reach
`post_title` and the `_author` post meta through the same writers, so they are handled the same way: refused at
upload rather than edited into shape.

`strip_shortcodes()`, which this PR used to do, is a single pass, and a single pass can leave a shortcode in place:

- It unwraps an escaped `[[tag]]` into a live `[tag]` — `strip_shortcode_tag()` returns `substr( $match, 1, -1 )`
  for that syntax by design.
- Removing one shortcode can splice the surrounding text into another. `[gal[caption]lery ids="1"]` carries no
  shortcode until the inner `[caption]` is taken out, at which point the remainder joins into a live
  `[gallery ids="1"]`.

Matching a shortcode has neither failing, and the uploader is told rather than having their header quietly rewritten.

`Theme URI` and `Author URI` come back out of scope. `WP_Theme::get()` returns them through `esc_url_raw()`, which
percent-encodes the delimiters, so those two cannot carry the syntax in the first place and the strip on them was a
no-op. `test_uri_headers_are_inert_before_the_check()` pins that, so they come back if it ever stops being true.

The name has the second writer the description has — `wporg_themes_approve_version()` re-reads `style.css` from SVN
when a version goes live, and a commit reaches it without passing the upload's check — so its delimiters are encoded
there alongside the description's. That also covers names stored before this change.

Bracketed prose is unaffected: `[developers]` is not a registered shortcode, so nothing matches it.

## Testing steps

1. Upload a theme whose `style.css` has `Theme Name: Fixture [caption width="1" caption="x"]y[/caption] Theme`. The
   upload is refused with a message pointing at the `Theme Name:` line. Try `[[caption …]y[/caption]]` and
   `[gal[caption]lery ids="1"]` too, and the same three in `Author:`; all are refused.
2. Upload a theme named `A theme for [developers]`. It is accepted, and the name reads as written.
3. Commit a `style.css` to SVN renaming a listed theme to a name carrying a shortcode, then take the version live.
   The stored title shows the brackets as text rather than running them.
4. `npm run themes:test` from `environments/`
   → 159 tests, 250 assertions, all passing. The pre-existing `Search_Published_Filter_Test` risky-test notice and
   the `has_screenshot()` deprecation from a fixture without a screenshot are both unrelated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Theme uploads now reject shortcode markup in theme names, descriptions, and author information before storage.
  * Theme names and descriptions are safely stored without interpreting shortcode syntax, while ordinary bracketed text remains supported.
  * Upload validation now identifies files omitted from automated review and filenames that may not work across supported platforms, with concise file lists in error messages.

* **Tests**
  * Expanded coverage for shortcode handling, URL sanitization, and live-version metadata storage.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
+++ b/wordpress.org/public_html/wp-content/plugins/theme-directory/class-wporg-themes-upload.php
@@ -788,13 +788,34 @@ class WPORG_Themes_Upload {
 			$style_errors->add( 'no_description', $error );
 		}
 
-		if ( preg_match( '/' . get_shortcode_regex() . '/', $theme_description ) ) {
+		/*
+		 * The one-line headers that can carry the syntax, as
+		 * `error code => [ style.css line, value that gets stored ]`. The description
+		 * reaches `post_content`, the name reaches `post_title`, and the author is stored
+		 * as the `_author` post meta that the themes API reads back. Each is read the way
+		 * `create_or_update_theme_post()` reads it, so the check sees the value that would
+		 * be stored rather than the one in the file.
+		 *
+		 * `Theme URI` and `Author URI` need no entry: `WP_Theme::get()` returns them
+		 * through `esc_url_raw()`, which percent-encodes the delimiters.
+		 */
+		$stored_headers = array(
+			'shortcode_in_description' => array( 'Description', $theme_description ),
+			'shortcode_in_name'        => array( 'Theme Name', (string) $this->theme->get( 'Name' ) ),
+			'shortcode_in_author'      => array( 'Author', (string) $this->theme->get( 'Author' ) ),
+		);
+
+		foreach ( $stored_headers as $code => list( $header, $value ) ) {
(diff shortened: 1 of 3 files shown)


mcliwanow (approved):
Thanks for the changes, code looks good and tests well.

bor0:
Reworked to follow [15147]: the headers are refused at upload instead of run through `strip_shortcodes()`, and the name — which `wporg_themes_approve_version()` also writes from SVN — has its delimiters encoded there. Trunk is merged in, so the description handling this branch used to carry is gone.

This also drops `Theme URI` and `Author URI` from the scope. `WP_Theme::get()` returns them through `esc_url_raw()`, which percent-encodes the delimiters, so they cannot carry the syntax and the strip on them was a no-op; `test_uri_headers_are_inert_before_the_check()` pins that.

That covers both CodeRabbit points: the `strip_shortcodes()` call it flagged is gone, and the test setup it flagged has been rewritten.

PR #872, opened 2026-09-04
Plugin Directory: Print the plugin excerpt as plain text

mcliwanow:
## Why

The plugin card block and the plugin embed template print the excerpt with `the_excerpt()`, as HTML. Every other consumer of the same value treats it as text: the `og:description` and `meta description` tags (`esc_attr( strip_tags( get_the_excerpt() ) )`), the JSON-LD (`wp_json_encode`), and the API's `short_description`. The readme parser only ever stores plain text in the excerpt; the 43 published plugins that still carry tags in it all predate readme short descriptions.

## What changed

- The plugin card (src and build) and the embed print `esc_html( get_plugin_excerpt_text() )` inside a `<p>` instead of calling `the_excerpt()`. The new template tag in `inc/template-tags.php` texturizes, turns `<br />` into a space, strips the remaining tags and collapses whitespace. `get_the_excerpt()` keeps the translation filter and the empty-excerpt fallback; `wptexturize` runs first so text inside `<code>` keeps its straight quotes as it does today; the `<p>` keeps the existing CSS working.
- Dropped from `the_excerpt()`'s filter chain on purpose: `convert_smilies`, `convert_chars`, `wpautop`, `shortcode_unautop`, `wp_replace_insecure_home_url` and `wp_filter_content_tags`. The last two are no-ops on text without tags.
- Visible effects: the 43 excerpts that still hold `<a>`, `<strong>`, `<em>`, `<code>` or `<br />` render as plain sentences, with a space where a line break was and runs of spaces collapsed. A literal `<` followed by a word now drops the rest of the excerpt, which is what the readme parser does with the same input.

## Testing

wp-env `environments/plugin-directory` (WP trunk), the same four plugin posts on trunk and on this branch, container restarted between the two. The card (search results) and the embed print the same string in both cases.

| excerpt in the database | trunk | branch |
|---|---|---|
| `Tools for A &amp; B -- it's "done"...` | `Tools for A &amp; B &#8212; it&#8217;s &#8220;done&#8221;&#8230;` | identical |
| `Use <code>&lt;?php getRSS('…', '5');?&gt;</code> to fetch a <a href="…">feed</a>.` | `<code>` and `<a>` rendered as elements, straight quotes inside `<code>` | `Use &lt;?php getRSS(&#039;…&#039;, &#039;5&#039;);?&gt; to fetch a feed.` |
| `(<p>,<div>,<span) and 5 < 6` | `(` then the browser closes the paragraph | `(,,` |
| `…aleatoriamente.<br />Uma forma…<br />No admin…` (random-thumbs) | four lines | `…aleatoriamente. Uma forma… No admin…` |
| `…of the blog.         To these…` (nine spaces) | nine spaces | one space |
| `START <a id="x" data-x="y" href="#">PCLICK</a> END` | element | `START PCLICK END` |

Production: the info API lists 71,547 published plugins, 43 with a tag in `short_description`, all of them `a`, `strong`, `em`, `code` or `br` from old-style plugin headers. Changed lines lint clean with `phpcs`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved plugin excerpt rendering across plugin cards and embedded plugin views.
  * Excerpts now consistently display as clean, plain text without HTML markup.
  * Line breaks in excerpts are converted to spaces for more natural, readable formatting.
  * Shared excerpt handling provides consistent results wherever plugin summaries appear.

### DIFF
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/template-tags.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/template-tags.php
@@ -822,7 +822,7 @@ function the_author_notice( $post = null ) {
  */
 function get_plugin_excerpt_text() {
 	$excerpt = wptexturize( get_the_excerpt() );
-	$excerpt = preg_replace( '#<br\s*/?>#i', ' ', $excerpt );
+	$excerpt = preg_replace( '#<br\b[^>]*>#i', ' ', $excerpt );
 
-	return trim( wp_strip_all_tags( $excerpt, true ) );
+	return wp_strip_all_tags( $excerpt, true );
 }
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/template-tags.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/template-tags.php
@@ -807,3 +807,22 @@ function the_author_notice( $post = null ) {
 		);
 	}
 }
+
+/**
+ * Returns the current plugin's excerpt as plain text, for the card and the embed.
+ *
+ * Excerpts are plain text by construction (the readme parser strips them), but
+ * excerpts imported from a plugin file header can still carry markup. Line
+ * breaks become spaces so the words on either side stay apart, then the tags
+ * go. Texturize runs first so that text inside <code> keeps its literal
+ * quotes, matching what the_excerpt() printed. The result is unescaped text:
+ * callers wrap it in esc_html().
+ *
+ * @return string The excerpt as plain text.
+ */
+function get_plugin_excerpt_text() {
+	$excerpt = wptexturize( get_the_excerpt() );
+	$excerpt = preg_replace( '#<br\s*/?>#i', ' ', $excerpt );
+
+	return trim( wp_strip_all_tags( $excerpt, true ) );
+}
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/build/blocks/plugin-card/render.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/build/blocks/plugin-card/render.php
@@ -20,7 +20,7 @@ do_blocks( '<!-- wp:wporg/link-wrapper /-->' ); // Import the styles
 		<?php echo wp_kses_post( Template::get_star_rating( $post, false ) ); ?>
 
 		<div class="entry-excerpt">
-			<?php the_excerpt(); ?>
+			<p><?php echo esc_html( wp_strip_all_tags( wptexturize( get_the_excerpt() ) ) ); ?></p>
 		</div><!-- .entry-excerpt -->
 	</div>
 
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/build/blocks/plugin-card/render.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/build/blocks/plugin-card/render.php
@@ -2,6 +2,7 @@
 namespace WordPressdotorg\Theme\Plugins_2024\PluginCard;
 
 use WordPressdotorg\Plugin_Directory\Template;
+use function WordPressdotorg\Plugin_Directory\Theme\get_plugin_excerpt_text;
 
 $post = get_post();
 // Simulates wporg/link-wrapper block until this is migrated to a block, or it supports href=permalink.
@@ -20,7 +21,7 @@ do_blocks( '<!-- wp:wporg/link-wrapper /-->' ); // Import the styles
 		<?php echo wp_kses_post( Template::get_star_rating( $post, false ) ); ?>
 
 		<div class="entry-excerpt">
-			<p><?php echo esc_html( wp_strip_all_tags( wptexturize( get_the_excerpt() ) ) ); ?></p>
+			<p><?php echo esc_html( get_plugin_excerpt_text() ); ?></p>
 		</div><!-- .entry-excerpt -->
 	</div>
 
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/embed-plugin.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/embed-plugin.php
@@ -129,7 +129,7 @@ if ( ! headers_sent() ) {
 		</p>
 
 		<div class="wp-embed-excerpt">
-			<?php the_excerpt(); ?>
+			<p><?php echo esc_html( wp_strip_all_tags( wptexturize( get_the_excerpt() ) ) ); ?></p>
 		</div>
 
 		<?php
--- a/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/embed-plugin.php
+++ b/wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/embed-plugin.php
@@ -129,7 +129,7 @@ if ( ! headers_sent() ) {
 		</p>
 
 		<div class="wp-embed-excerpt">
-			<p><?php echo esc_html( wp_strip_all_tags( wptexturize( get_the_excerpt() ) ) ); ?></p>
+			<p><?php echo esc_html( get_plugin_excerpt_text() ); ?></p>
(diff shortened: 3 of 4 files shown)


bor0:
Reviewed this against the branch, and against a WordPress checkout where the behaviour was load-bearing rather than from memory.

The change looks right, and the reasoning in the description checks out:

* Those are the only three `the_excerpt()` call sites in `wporg-plugins-2024`. `functions.php:332-333` and `api/routes/class-plugin.php:196` already went through `get_the_excerpt()`.
* The translation filter survives, since `class-plugin-directory.php:454` hooks `get_the_excerpt` rather than `the_excerpt`.
* The "identical output" claim in the test table holds. `esc_html()` calls `_wp_specialchars( $text, ENT_QUOTES )`, and that signature defaults `$double_encode` to `false` (`wp-includes/formatting.php:945`), so existing entities like `&amp;` and `&#8212;` are not re-encoded.
* The `<p>` wrapper is needed rather than incidental: `client/components/_plugin-card.scss:35` styles `.entry-excerpt p { margin: 0 }`.
* `src/` and `build/` stay in sync.

### One regression worth fixing

`wp_strip_all_tags()` does not substitute anything for a line break, so a legacy excerpt carrying `<br />` comes out with its words glued together:

```php
strip_tags( 'Do this.<br>Then that.' ) === 'Do this.Then that.'
```

The description says the 43 affected excerpts use `a`, `strong`, `em`, `code` or `br`, so some of them will render as run-together text rather than as the plain sentences the change is aiming for. `wp_strip_all_tags( $text, true )` does not help, since the second argument only collapses whitespace that is already there. Replacing `<br>` / `<br />` with a space before stripping would cover it, or the affected rows could be fixed alongside the deploy the way #860 handles its two.

### Smaller notes

* The list of dropped filters is missing two. Core also hangs `wp_replace_insecure_home_url` and `wp_filter_content_tags` on `the_excerpt` (`wp-includes/default-filters.php:217-218`). Both are no-ops on tag-free text, so nothing needs to change; the only visible effect is that a bare `http://` URL in excerpt text is no longer upgraded to https. Worth adding to the commit message for completeness.
* The expression now appears three times, across two files that are already copies of each other. A small `Template::` helper would keep the `src/` and `build/` copies from drifting, and the card already imports `Template`.
* With `wpautop` gone, a multi-paragraph excerpt collapses into a single `<p>`. Irrelevant for readme short descriptions, relevant for the same legacy set as above.

The trade-off in the third test-table row (a literal `<` followed by a word dropping the rest of the excerpt) is a fair call to make given the alternative renders old markup visibly, and I like that it is documented rather than left to be discovered.

mcliwanow:
Thanks for the review, @bor0. I made the changes in https://github.com/WordPress/wordpress.org/pull/872/commits/d9ec45905732ae785c43bc74f418334d8560834a

- extracted code to a helper script
- made a safer `<br />` stripping
- added the two missing filters to the commit message and the PR text; both are no-ops on tag-free text

wordpress.org/public_html/wp-content/themes/pub/wporg-plugins-2024/inc/template-tags.php
```diff
@@ -807,3 +807,22 @@ function the_author_notice( $post = null ) {
...
+ * breaks become spaces so the words on either side stay apart, then the tags
+ * go. Texturize runs first so that text inside <code> keeps its literal
+ * quotes, matching what the_excerpt() printed. The result is unescaped text:
+ * callers wrap it in esc_html().
+ *
+ * @return string The excerpt as plain text.
+ */
+function get_plugin_excerpt_text() {
+	$excerpt = wptexturize( get_the_excerpt() );
+	$excerpt = preg_replace( '#<br\s*/?>#i', ' ', $excerpt );
```
bor0: `#<br\s*/?>#i` only matches bare tags, so an attribute-bearing break still glues the words on either side:

```
A<br class="clear" />B  =>  AB
```

which is the case this line exists to prevent. `\b[^>]*` closes it without affecting anything else:

```suggestion
	$excerpt = preg_replace( '#<br\b[^>]*>#i', ' ', $excerpt );
```

Whether any of the 43 affected excerpts actually carries an attribute on a `<br>` I cannot check from here, so this may well be theoretical. It costs one token to rule out.

bor0 (commented):
Re-reviewed at `d9ec459` (plus the trunk merge). The fix I asked for landed and does the job:

```
Do this.<br>Then that.  =>  Do this. Then that.
A<br />B                =>  A B
A<BR>B                  =>  A B
```

Extracting the helper was the right call too, it keeps the card and the embed provably identical rather than incidentally so.

Everything I verified previously still holds on this head: those are the only three `the_excerpt()` call sites in the theme, the translation filter survives because `class-plugin-directory.php:454` hooks `get_the_excerpt` rather than `the_excerpt`, `esc_html()` does not double-encode so the "identical output" rows in your table stand, the `<p>` wrapper is load-bearing for `.entry-excerpt p { margin: 0 }`, and `src/` and `build/` are in sync. I also checked that nothing in `wp-content` hooks `the_excerpt`, the only other reference is `wporg-main/front-page.php:244` applying it for a different theme, so dropping the filter chain drops core's callbacks and nothing else, exactly as the description says.

### One residual, same class as the original

Leaving this as a comment rather than an approval only because it is a one-liner, and cheaper to fix now than to have a committer land it and find a glued excerpt afterwards. See the inline note.

I would skip CodeRabbit's related point about adjacent block tags (`<p>a</p><p>b</p>`). Those tags do not occur in this data, and widening the substitution to all tags would insert a space mid-word for `foo<strong>bar</strong>`, which is a worse trade.

### Nit

`trim()` around `wp_strip_all_tags()` is redundant, that function already trims its return value.

mcliwanow:
Both in db7b5715f: `\b[^>]*` on the break pattern and the `trim()` gone. None of the 43 rows has an attribute on a `<br>`, so this one is for the future. Skipped the adjacent-block-tag suggestion for the reason you gave.

bor0 (approved):
Re-reviewed at `db7b5715f`, checked against the branch rather than from my notes on the previous round.

Both items from the last round landed, and the helper now reads:

```php
$excerpt = wptexturize( get_the_excerpt() );
$excerpt = preg_replace( '#<br\b[^>]*>#i', ' ', $excerpt );

return wp_strip_all_tags( $excerpt, true );
```

I ran that pattern plus core's `wp_strip_all_tags()` body over the cases that motivated it:

| input | output |
|---|---|
| `A<br class="clear" />B` | `A B` |
| `A<br/>B` | `A B` |
| `A<BR>B` | `A B` |
| `<br` with a newline before its attributes | `A B` |
| `A<brx>B` | `AB` (correctly not matched) |
| `one   two\n\nthree` | `one two three` |
| `5 < 6 and <p>x` | `5 < 6 and x` |

`\b` sits between `r` and `/` as well as between `r` and `>`, so the bare forms still match and `<brx>` stays untouched. Dropping the explicit `trim()` was right too: `wp_strip_all_tags()` trims unconditionally on its way out, so it was dead code.

### Verified on this head

* `git grep the_excerpt` across the theme now returns only `get_the_excerpt()`, the two meta tags in `functions.php:332-333`, and the doc comment. Every `the_excerpt()` call site in the theme is gone.
* Nothing in `wp-content` hooks `the_excerpt`, so dropping the filter chain drops core's callbacks and nothing else, exactly as the description says.
* The only `get_the_excerpt` hook in the estate is the translation filter at `class-plugin-directory.php:454`, priority 1. No `wptexturize` is attached there, so the manual call is not a double texturize.
* `embed-plugin.php` sits in the same `WordPressdotorg\Plugin_Directory\Theme` namespace as the helper, so it resolves without an import, and the card's `render.php` has the `use function` line it needs.
* `src/` and `build/` render.php are byte-identical.
* `.wp-embed-excerpt` (inline CSS at `embed-plugin.php:84`) has no `p` rule of its own, and `the_excerpt()` plus `wpautop` already emitted a `<p>` there, so the embed markup is unchanged. The card still needs its `<p>` for `.entry-excerpt p { margin: 0 }`.

### One follow-up, not for this PR

`functions.php:332-333` build `og:description` and `meta description` from a bare `strip_tags( get_the_excerpt() )`, so they still glue words together across a `<br>`, the same case this PR just fixed for the card and the embed. Now that a shared helper exists, those two are the obvious next callers. Not worth holding this up for.

LGTM.

PR #878, opened 2026-09-07
Plugin Directory: treat screenshot captions as text, not block markup

bor0:
A screenshot caption is display text. It is composed into the Image block markup that `Screenshots::display()` hands to `do_blocks()`, so it should stay a text leaf of that block.

- `Screenshots::build_image_block()` encodes block-comment delimiters in the caption before it goes into the `<figcaption>`.
- `Readme\Parser::filter_text()` drops HTML comments, so newly imported captions do not carry the syntax at all.

Both layers are covered by `tests/Screenshot_Caption_Block_Delimiters_Test.php`.

https://claude.ai/code/session_01T1kfkymZDgMdeVRpDUUQL7

## Summary by CodeRabbit

* **Bug Fixes**
  * Sanitized readme text and screenshot captions now prevent HTML and Gutenberg block-comment delimiters from disrupting rendered content.
  * Screenshot captions containing comment syntax remain correctly contained within their image blocks.

* **Tests**
  * Added coverage for comment removal and safe screenshot block rendering.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
+++ b/wordpress.org/public_html/wp-content/plugins/plugin-directory/readme/class-parser.php
@@ -697,6 +697,10 @@ class Parser {
 		// $text = make_clickable( $text );
 		$text = wp_kses( $text, $allowed );
 
+		// Readme text has no use for HTML comments, and dropping them keeps
+		// comment syntax out of the fields that are later composed into markup.
+		$text = preg_replace( '#<!--.*?(?:-->|$)#s', '', $text );
+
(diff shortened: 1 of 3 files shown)


mcliwanow (approved):
Tested locally on the wp-env site with the fixture caption stored in post meta: with the trunk files it comes out as a block, on this branch it's literal text in the figcaption.  I also ran a few more variants (unterminated comment, `<<!---->!--`, `--!>`, nested, triple dashes, newlines inside the comment), all end up as `core/image` only, and Markdown code samples containing block markup survive the parser strip since they're entity-encoded by the time it runs.

PR #880, opened 2026-09-07
Support HelpHub: remove the unused contributors module

KokkieH:
Removes the `helphub-contributors` module from the Support HelpHub plugin, along with its `require` in the plugin bootstrap.

The module added a Contributors field to the Publish metabox and rendered a contributor byline after the post content.

It never applied to the article post type. The module flags post type support from its constructor on `plugins_loaded`, and takes a snapshot of the post types registered at that point, so it picked up the core post types but not the article post type, which is registered later on `init`. The field is also hooked only into the classic editor, and the sites that load this plugin use the block editor, so it does not render there either.

Nothing outside the module referenced its meta key, hooks, assets or CSS classes.

## Testing

With the Support HelpHub plugin active:

- Articles render normally and no contributor byline appears.
- The Publish metabox on the article, post and page editors no longer shows a Contributors field.
- The Articles, Posts and Pages list tables no longer show a Contributors column.
- No PHP notices or errors.

The remaining modules each register something checkable, so `wp eval` confirms they all still load rather than leaving it to inspection:

```php
wp eval '
$checks = array(
  "codex languages (shortcode)"    => shortcode_exists( "codex_languages" ),
  "codex languages (class)"        => class_exists( "HelpHub_Codex_Languages" ),
  "read time (save_post hook)"     => (bool) has_action( "save_post", "hh_calculate_and_update_post_read_time" ),
  "manager (class)"                => class_exists( "HelpHub_Manager" ),
  "front page blocks (widget)"     => class_exists( "Support_HelpHub_Front_Page_Blocks_Widget" ),
  "post types (class)"             => class_exists( "HelpHub_Post_Types" ),
  "handbook TOC (class)"           => class_exists( "WPorg_Handbook_TOC" ),
  "contributors gone"              => ! class_exists( "HelpHub_Contributors" ),
);
foreach ( $checks as $label => $ok ) { printf( "%-32s %s%s", $label, $ok ? "PASS" : "FAIL", PHP_EOL ); }
'
```

All rows should report `PASS`.

Note that `helphub-front-page-blocks` registers a legacy `WP_Widget` on `widgets_init`, not editor blocks — there is nothing to look for in the block editor.

## Summary by CodeRabbit

- **Removals**
  - Removed the HelpHub Contributors module from the Support HelpHub plugin.
  - Contributor selection and management features are no longer available in the WordPress admin.
  - Contributor information is no longer displayed on frontend content.
  - Removed related contributor styling, metadata, documentation, and uninstall handling.


KokkieH:
Companion PR: https://github.com/WordPress/wporg-documentation-2022/pull/114

`support-helphub` has two upstreams — it lives in meta SVN and is directly committable, but `bin/sync/wporg-documentation.sh` syncs it from `wporg-documentation-2022`, overwriting SVN wholesale. See WordPress/wporg-documentation-2022#113 for background.

**Both need to land.** Committing this one alone would be undone by the next sync, since the documentation repository would still carry the module and the sync re-adds it. Merge order between the two does not matter, provided both land before the next sync runs.

mcliwanow (approved):
Thanks for the removal. I checked on the sandbox and can't find any traces of the plugin being used.

KokkieH:
Closing--superseded by https://github.com/WordPress/wporg-documentation-2022/pull/114

PR #887, opened 2026-09-09
Add Existing Users: resolve the invite role through the editable list

bor0:
The invite created by `admin_action_adduser` carries a role that core applies later, when `/newbloguser/{key}/` is opened. That role should therefore be one the inviting user is allowed to assign at the moment the invite is written.

- `wp_ensure_editable_role()` is called before the invite is stored, the same check core uses on its own `user-new.php` paths, and the stored role is the value that was checked.
- The invite key comes from `wp_generate_password( 20, false )`, matching core, instead of being derived from the invited user's ID.
- `$_REQUEST['email']` and `$_REQUEST['role']` are read as strings, so a missing or non-scalar value takes the normal path instead of a PHP warning or a `TypeError`.

https://claude.ai/code/session_016iz3Ho52xV2mFqe4HoVnNX

## Summary by CodeRabbit

- **Security Improvements**
  - Improved validation and sanitization of email addresses and requested roles.
  - Strengthened invitation key generation with more secure, unpredictable values.
  - Ensured only editable roles are accepted when inviting existing users.

### DIFF
--- a/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-add-existing-users-via-username.php
+++ b/wordpress.org/public_html/wp-content/mu-plugins/pub/wporg-add-existing-users-via-username.php
@@ -44,7 +44,7 @@ function add_existing_user_to_site() {
 	check_admin_referer( 'add-user', '_wpnonce_add-user' );
 
 	$user_details = null;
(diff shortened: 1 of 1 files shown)


mcliwanow (approved):
thanks for the changes, PR looks good. `wp_ensure_editable_role` is a good addition - I didn't know this one but makes perfect sense to use it

PR #891, opened 2026-09-09
Translation Events: Show a notice after trashing, restoring, or deleting an event

ekamran:
Trashing, restoring, or permanently deleting an event redirects the user away from the page the action was triggered from, and nothing was shown afterwards, so there was no confirmation that the action actually happened. This sets a GlotPress notice before each redirect, so it is carried across in the notice cookie and rendered on the destination page.

It covers all the places an event can be trashed, restored, or deleted: the Delete Event button on the edit form, which trashes over AJAX and then redirects to My Events, on both success and failure; the trash and restore links handled by the trash route; and the permanent delete link handled by the delete route.

Adds tests covering all five notice paths by capturing the notice cookie through the `gp_set_cookie` filter. The tests are ports of the ones that passed in the archived repository's CI, adapted to this suite's conventions; they were not executed locally because the wporg and GlotPress test environment is not available here.

This ports the archived standalone repository's pull request, closed when Translation Events moved into this repository: https://github.com/WordPress/wporg-gp-translation-events/pull/392 (issue https://github.com/WordPress/wporg-gp-translation-events/issues/372). Opening as a pull request per the closing guidance on that PR.

## Summary by CodeRabbit

- **New Features**
  - Added clear confirmation notices when events are moved to the trash, restored, or permanently deleted.
  - Added error notices when moving an event to the trash fails.

- **Bug Fixes**
  - Failed trash actions now retain the event’s current status instead of reporting an incorrect result.
  - Notices are displayed consistently before redirecting after event actions.

### DIFF
--- a/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-events/includes/event/event-form-handler.php
+++ b/wordpress.org/public_html/wp-content/plugins/wporg-gp-translation-events/includes/event/event-form-handler.php
@@ -88,14 +88,25 @@ class Event_Form_Handler {
 				wp_send_json_error( esc_html__( 'Event has stats so it cannot be deleted.', 'gp-translation-events' ), 422 );
 			}
 
-			if ( false === $this->event_repository->trash_event( $event ) ) {
-				$response_message = esc_html__( 'Failed to delete event.', 'gp-translation-events' );
-				$event_status     = $event->status();
-			} else {
-				$response_message = esc_html__( 'Event deleted successfully.', 'gp-translation-events' );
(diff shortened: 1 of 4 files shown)


ekamran:
I kept this port aligned with the archived PR, where the route-level notices were unconditional and the explicit failure notice was added for the form-handler path.

I agree the route methods could be made more defensive by checking the repository return values, but that would expand this port into route/repository error handling. I would prefer to keep this PR focused on restoring the missing redirect notices, and handle that separately if maintainers want the extra hardening.

PR #898, opened 2026-09-10
Support Forums: Keep forum content to the blocks the forums support

bor0:
The Support Forums offer a fixed set of blocks. This applies that same set to
forum, topic and reply content as it is stored, and again before it is
rendered, recursing through nested blocks so the check cannot be sidestepped
by nesting.

Content that already only uses supported blocks is passed through untouched.

https://claude.ai/code/session_01GASiJLUcxvZemhxGM1Vfu1

## Summary by CodeRabbit

- **Enhancements**
  - Forum topics, replies, and other forum content now retain only supported block types.
  - Unsupported blocks are automatically removed when content is saved or displayed.
  - Nested block content is handled consistently to preserve valid formatting and rendering.
  - Reply content in subscription email notifications is filtered to include only supported blocks.


bor0:
Good catch on the mechanism, addressed in 770f122.

`Hooks::bbp_subscription_mail_message()` runs at priority 5 on `bbp_subscription_mail_message` and calls `remove_all_filters( 'bbp_get_reply_content' )`, which does drop the callback added here.

The consequence is a bit different from the one described, though. `strip_tags()` on the raw stored content also strips the block delimiters, so that particular line is not where it matters. What matters is that the content is read *again* afterwards: Blocks Everywhere hooks the same filter at priority 10, and its handler re-fetches `bbp_get_reply_content( $reply_id )` and runs it through the block renderer, which by then is operating on unfiltered content.

So rather than moving anything into `class-hooks.php`'s own retrieval, the filter is re-registered right after the `remove_all_filters()` call, which covers both that read and the later one.

The forum-subscription path (`bbp_forum_subscription_mail_message`, topic content) was already covered, since nothing clears `bbp_get_topic_content`.

On the docstring coverage warning: the three functions without docblocks are pre-existing (`__construct`, `allowed_blocks`, `editor_settings`) and are only counted because the diff touches them. Leaving them out to keep the diff focused.

mcliwanow (approved):
Tested locally against core, bbPress 2.7-alpha and Blocks Everywhere. Stripping holds at every depth and allowed content round-trips byte-for-byte. Checked the save, display and email paths: a stored payload doesn't render, a submission that strips down to nothing hits the existing empty-content error, and the subscription mail comes out clean. phpcs clean on the changed lines. LGTM.

PR #902, opened 2026-09-11
Escape the remaining output, clearing EscapeOutput

obenland:
Clears the last of `WordPress.Security.EscapeOutput`. Repo-wide the sniff goes from 477 findings to **0**; the whole security family drops 2,711 → 2,234.

Based on current trunk, including the changes from #901.

Roughly two thirds of the 477 was not template output at all, and escaping it would have broken it: WP-CLI scripts behind `php_sapi_name()` guards, cron log echoes, API bodies in JSON/JSONP/serialized PHP/XML, SVN authz text, `trigger_error()` messages. Those carry a scoped ignore naming what the call site actually emits, following the convention already in `class-import.php` and `class-builder.php`. The rest is real output and gets a context-appropriate escaper.

### Bugs this turned up

- **Three ignores that had been suppressing nothing.** They named `WordPress.XSS.EscapeOutput`, the sniff's pre-WPCS-1.0 name, so those lines were unprotected *and* invisible. Two renamed, one dead file-level `disable` removed. Same for three `// WPCS: XSS OK.` annotations, which current WPCS no longer honours.
- **Two helpers used as if they returned a value.** The Showcase comment byline passed `comment_type()`, `comment_author_link()` and `comment_date()` to `printf()` — all three echo and return `null`, so the values printed before the sentence and the format rendered with three empty slots. The wporg theme's single-comment title used `esc_html_e()` as its format string, printing `One thought on &ldquo;%1$s&rdquo;` literally plus a PHP 8.1 deprecation. The sniff could not see the second one: `esc_html_e()` counts as escaping, so the line looked clean while rendering wrong.

### Decisions worth a look

- `do_blocks()` gets scoped ignores, not a ruleset entry. It does not escape its return value, so `customAutoEscapedFunctions` would be claiming something untrue.
- `esc_translation()` likewise stays an ignore. It is `htmlspecialchars( $text, ENT_NOQUOTES, false, true )` — no quote escaping, so it is only safe in element text, which is where all four call sites land. Existing entities are double-encoded so translations render verbatim.
- Two ruleset changes. The Google Cloud stream wrapper joins the other vendored excludes — its own header calls it "99.99% the code from `vendor/google/cloud-storage/src/StreamWrapper.php`", kept diffable against upstream. And `esc_html___()` is registered as auto-escaping, because it genuinely returns `esc_html( $text )`.
- Block `render.php` files are updated under `src/` and `build/` together, since wp-scripts copies them verbatim; a build reproduces the same result.

### Checked rather than assumed

The plugins-info JSONP callback is already restricted to `[a-zA-Z0-9_]`; the zip server's reflected filename is served as `text/plain`; the profiles activity `user_id` is `intval()`'d; the `?field= invalid.` message interpolates a literal loop variable, not request data; and the block style attribute is built from `sanitize_hex_color()` values. None needed escaping.

### Validation

Runtime comparisons cover the locale table, Rosetta project links, forum subscription notice, both WordPress.tv upload result notices, subtitle preview line breaks, GitHub invitation notices, and the standalone SSO redirect fallback. Restored markup also passes hostile-value fixtures. The subtitle preview already escapes file contents with `htmlspecialchars()` before adding line breaks; its existing behavior is preserved.

The Plugin Directory suite passes 453 tests and 1,540 assertions (27 PHPUnit deprecations). Changed-line PHPCS and PHP syntax checks pass for the follow-up fixes; full-repository PHPCS still reports existing violations.

`BASE_REF` set to trunk, the repository changed-line PHPCS check passes with zero violations across all 186 modified files. Every changed file passes `php -l`. I also diffed the msgid→translators-comment bindings for every touched file with the WP-CLI extractor to confirm no comment was orphaned by an inserted ignore, and confirmed the `src/`/`build/` pairs stay byte-identical.


obenland:
@coderabbitai help

obenland:
@coderabbitai full review

PR #910, opened 2026-09-11
Jobs: Document the post-a-job form render path

KokkieH:
The post-a-job form is added to `the_content`, but `get_template_part()` echoes its template and returns `null`, so the `$content .= get_template_part( ... )` calls append nothing — the form goes straight to the output stream. That is easy to misread as a bug and "tidy" into an output-buffered append, which would change what the rest of the filter chain sees.

- Documents the render path in the docblock, and corrects the `@return` description.
- Runs the filter at priority 12 rather than 10, so the form is emitted above core `do_shortcode()` at 11 and submitted values are never re-parsed as markup.

No behavioural change is expected: the form already renders outside the parser. Worth an eyeball on a sandbox though, since the priority bump shifts the form relative to other priority 10/11 `the_content` filters — both the empty-form render of `/post-a-job/` and the error/verify re-render.


bor0:
Fix is live, committed to svn

```
Sending        jobswp.php
Transmitting file data .done
Committing transaction...
Committed revision 15185.
[jobswor@morse jobswp]$ 
```

https://github.com/WordPress/wordpress.org/commit/8471c13d280a4652fefda0b86cbf63308c25e9f9

PR #915, opened 2026-09-12
Fix escaped committer links in release confirmation

AllTerrainDeveloper:
The release-confirmation listing builds an escaped profile link for every committer, then escapes the complete HTML link while interpolating it into the release sentence.

As a result, plugin committers see literal <a> markup instead of a clickable profile link.\n\nRender the already-escaped committer links directly and document that escaping boundary for PHPCS. Add a regression test that renders the release listing and verifies that the profile link remains HTML rather than becoming &lt;a&gt; text.\n\nValidation: PHP syntax checks passed for both changed files, and PHPCS passed for the added test.

image

## Bug Fixes

- Release listings now render committer profile links correctly, while filtering potentially unsafe markup from the displayed content.

## Tests

- Added coverage confirming that committer profile links appear as clickable links rather than escaped HTML in release listings.


obenland:
Sorry, I didn't see this until after I merged #923. I added props to https://meta.trac.wordpress.org/changeset/15214 in make/meta so this contribution is reflected in [your profile](https://profiles.wordpress.org/allterraindeveloper/). Thank you!

Closing in favor of #923.

PR #922, opened 2026-09-16
SSO: Tie remote-login hand-offs to the browser that started them

obenland:
A cross-domain sign-in hands the destination host a one-time `sso_token` in the URL it returns to. Until now that token stood on its own, so it was good in whatever browser presented it — including one that never started the login.

### What changes

A login that starts on a non-wordpress.org host now leaves that browser a `__Host-wporg_sso_bounce` cookie and passes its SHA-256 fingerprint along in `redirect_to`. The SSO host signs the fingerprint into the token and strips it from the target; the destination recomputes it from its own cookie when the token comes back, and a token whose fingerprint doesn't match isn't redeemed.

`__Host-` is the point of the prefix here: it can't carry a `Domain`, so no sibling host on the network can set the cookie for another.

### The restart

A token can legitimately arrive before the browser holds a ticket — a login begun on the SSO host itself, or a browser that dropped the cookie. Rather than dead-ending those, an unredeemable hand-off asks the SSO host once for a ticketed one and lands normally if that doesn't help either. It restarts whatever user the token names, so the response doesn't reveal which users exist, and only where a host-only ticket could actually reach: hand-offs bound for another host, off the network, already spent, or lapsed are landed as they were.

### Along the way

- Folded the repeated `wordpress.org` suffix test into `_is_wordpress_org_host()` and anchored it on a label boundary, so `notwordpress.org` no longer reads as part of the network. Moved `_normalize_token_host()` up beside it so the base class has both.
- Stopped handing a token to the SSO host itself — it set the cookie the user already holds — and compared the two hosts normalized, so a configured port still matches. This only shows up on local installs, where every host is the SSO host.
- Login screens now redirect to the SSO host with a 302 instead of a 301. The target names a ticket good for one browser and one login, so it isn't something a browser should cache permanently.
- Defaulted the temporary binding-reject log to a ticketed restart, the only case where a reject is unexpected, since the retry marker on its own comes from the URL.

### Testing

`npm run sso:test` in `environments/`. The new cases cover the binding in both directions (right browser, wrong browser, no browser), a hand-off link passed to someone else, each condition that declines a restart, and the two host-normalization cases above.

## Summary by CodeRabbit

* **Security**
  * Remote SSO login hand-offs are now bound to the initiating browser, reducing the risk of token reuse on another device or browser.
  * Bounce tickets are short-lived and securely stored to protect login transitions.

* **Bug Fixes**
  * Improved handling of SSO redirects across hosts, ports, and WordPress.org domains.
  * Failed remote login hand-offs can retry once with a fresh ticket when appropriate.
  * SSO-host redirects no longer receive unnecessary login tokens.


obenland:
Thanks — the mechanics here are right, but I'd rather not fold it into this PR.

It's pre-existing on both counts: `$this->host` has been raw `$_SERVER['HTTP_HOST']` since forever (`class-wporg-sso.php:89`), and neither `is_sso_host()` nor the `_get_targetted_host()` comparison in `_is_valid_targeted_domain()` is touched by this change. Both also fail closed — an odd `Host` means no redirect and no token, so the only person affected is whoever sent it.

It doesn't interact with what's here, either. Anything that clears `_is_valid_targeted_domain()` has already matched `VALID_HOSTS` exactly, so it's lowercase and portless by the time the new `_is_wordpress_org_host()` sees it and the normalization is a no-op — the normalized and unnormalized checks can't disagree on those values. The one place the port did matter is the local-install case where `sso_host` deliberately carries one, and that's handled at the comparison this PR adds in `maybe_add_remote_login_bounce_to_post_login_url()`, covered by `test_post_login_redirect_to_a_ported_sso_host_gets_no_token`.

Taking the suggested diff as-is would change behaviour for callers well outside this change — `add_allowed_redirect_host()`, `_get_safer_redirect_to()`, and the redirect entry guard all lean on `is_sso_host()`. On a local install with a port, `host=wordcamp.org` against `sso_host=wordcamp.org:8888` would flip from the non-SSO branch to the SSO branch. That's arguably the right outcome, but it wants its own tests rather than riding along here.

Happy to do it as a follow-up if a committer agrees it's worth normalizing those two sites.

PR #929, opened 2026-09-17
Support Forums: Restore the moderator profile edit link

obenland:
Moderators reported in #forums that the **Edit Forum Profile** link disappeared from support profiles, leaving `/edit` reachable only by typing the URL.

## What happened

`bbpress/user-details.php` gated the whole moderator nav block — Subscriptions, Edit Forum Profile, Edit WP.org Profile — on the site-wide `edit_users` capability:

```php
<?php if ( bbp_is_user_home() || current_user_can( 'edit_users' ) ) : ?>
```

Forum moderators hold no such capability. They only ever passed that check because `Moderators::map_meta_caps()` rewrote `edit_users` to `moderate` on profile pages.

bbPress [r7513](https://bbpress.trac.wordpress.org/changeset/7513) narrowed its own Super Moderator mapping to `edit_user` and `promote_user`, dropped `edit_users` entirely, and scoped what remains to `bbp_is_single_user_edit()` — grantable only once the profile editor is already open. r15202 handed the policy over to bbPress behind `function_exists( 'bbp_current_user_can_edit_user_field' )`, so the old shim stopped registering the moment the new bbPress reached the forums, two days after the commit. That is why the timing looked unrelated. `/edit` still works because bbPress grants the capability there.

## This change

**Ask about the profile being viewed**, as bbPress's own `user-details.php` does and as every other check in the forums already does:

```php
<?php if ( bbp_is_user_home() || current_user_can( 'edit_user', bbp_get_displayed_user_id() ) ) : ?>
```

**Answer that question on a profile page.** Under r7513's scoping the answer is false everywhere except the editor itself, so no template can ask it — bbPress's default profile template is dark for the same reason, and `bbp_current_user_can_edit_user_field()` inherits the limit. `Moderators::map_profile_view_caps()` extends the policy to the surrounding profile under the conditions bbPress applies there: front-end only, Super Moderators enabled, never the current user, and never keymasters, administrators or super admins. It hooks `bbp_map_primary_meta_caps`, the extension point r15202's docblock points at.

### What this widens, and what it doesn't

The page scope widens: `edit_user` and `promote_user` become answerable on any front-end profile view, not only the editor. That is the same scope as the upstream patch below, and it means `bbp_edit_user_handler()` will also accept a `bbp-update-user` POST sent to a non-editor profile URL — the handler itself gates on `is_admin()`, the nonce and the capability, with no page binding of its own. Same moderator, same target, same required `update-user_{id}` nonce, and nothing that was not already available from the edit screen.

Everything else r7513 tightened is untouched: no site-wide `edit_users`, nothing in wp-admin, the protected-target exclusions, and the field-level policy over passwords, WordPress roles and staff forum roles, which `bbp_filter_user_edit_post_data()` enforces per field on every request regardless of page.

Neither of those boundaries depends on the editor-only narrowing. `bbp_parse_query()` bails in wp-admin, so none of these flags are ever set there, and core's `edit_user()` will not write a site role without the site-wide `promote_users` that this does not grant.

## Upstream

The real fix is widening bbPress's scope from `bbp_is_single_user_edit()` to `bbp_is_single_user()`, a strict superset that `bbp_parse_query()` sets on every profile view including the editor. Filed with a patch and a regression test as [bbPress #3685](https://bbpress.trac.wordpress.org/ticket/3685); `map_profile_view_caps()` is written to be deleted when that lands, and is harmless until then.

Also restored, since they already use the canonical check:

- `bbpress/user-subscriptions.php` and `bbpress/user-reports-submitted.php`
- `Users::filter_query_reported_topics()` — the reported-topics profile view
- `Users::bulk_topic_unsubscribe()` and its POST handler — "Unsubscribe from all topics"

## Testing

As a forum moderator:

1. Visit another user's support profile — the sidebar lists Subscriptions and **Edit Forum Profile**, and the link opens the editor.
2. Set that user's forum role to Blocked and save.
3. Visit a keymaster's or an administrator's profile — no edit link.
4. Visit your own profile — links unchanged.
5. In wp-admin, confirm a moderator still has no user-editing rights.

Not yet exercised against a real moderator account; `support-forums` has no test suite and `environments/` has no forums setup, so this is reasoned from bbPress trunk rather than observed.


obenland:
Upstream ticket for the scoping fix: [bbPress #3685](https://bbpress.trac.wordpress.org/ticket/3685), with the patch and a regression test attached. `Moderators::map_profile_view_caps()` now references it and can be dropped once it lands and a bbPress with the fix is deployed to the forums.

